PR #13, opened 2017-06-11
demo futures API

vitiral:
demoing that japaric/nb#2 should be the way we go here. There really isn't a reason to create a separate API!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,6 +39,7 @@
 //!
 //! ## Traits
 //!
+//! -- opened issue: https://github.com/japaric/nb/issues/2
 //! The HAL is specified using traits to allow generic programming. These traits
 //! traits will make use of the [`nb`][] [crate] (*please go read that crate
 //! documentation before continuing*) to abstract over the asynchronous model
@@ -50,7 +51,8 @@
 //! Here's how a HAL trait may look like:
 //!
 //! ``` rust
-//! extern crate nb;
+//! extern crate futures;
+//! use futures::Async;
 //!
 //! /// A serial interface
 //! pub trait Serial {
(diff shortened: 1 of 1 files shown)


vitiral:
Note: there are a few outstanding questions in the comments -- but I think the basics here are sound -- we should not create a new data type when minimal `no_std` futures will do the job. No need to force users to do type conversion on something that is *already zero cost*.

japaric:
I'm afraid that this is *not* how you write future based APIs. You have to return a type that implements the Future trait. Let me show you why this API is wrong:

``` rust
timer.wait(); // returns Async<()>
```

This is wrong because is *not* a blocking operation but you wanted a blocking operation. However, this code doesn't return a warning because Async is not marked as `#[must_use]`. If the API had returned a future then you would have gotten a warning: "futures do nothing unless polled". If you had used the current `nb` based API you would have gotten a warning: "unused result which must be used".

This API is not appropriate for callback-style / event-driven code as explained in https://github.com/japaric/nb/issues/2#issuecomment-307675495. You want an error instead of Async::NotReady when this HAL is used in callbacks.

This API is not compatible with the await! operator planned for futures. That operator only works with functions that return a future.

> Also... isn't `await` supposed to just "work" with futures?

An await operator that works with futures will be added to the futures crate but you don't need the Future abstraction to implement the await generator; generators suffice as they are a more general version of futures. In fact the await operator that will land in the futures crate turns the future abstraction into ... [a generator](https://github.com/alexcrichton/futures-await/blob/master/futures-await-macro/src/lib.rs).

> we should not create a new data type when minimal no_std futures will do the job

"we should not create a new data type (futures) when generators will do the job (await)". Just that generators were no close to being implemented when a mechanism for async IO was sorely needed so the futures crate was created.

> No need to force users to do type conversion

What type conversion? Users can directly use the futures based API in the hal-futures crate. HAL implementers only need to implement the nb-based traits.

> something that is already zero cost.

(Except that futures are [not really zero cost] when you compose them which is how you have to use them to write cooperative code. I *really* hope that generators won't have this problem but we can't tell until we test them.)

[not really zero cost]: https://github.com/rust-lang/rust/issues/37930

vitiral:
@japaric you were right, there were several things I was not understanding. Hopefully japaric/nb#3 helps clarify. I will update this PR once we have reached a conclusion there (basically the update will change `futures::Async` -> `futures::Poll`

homunkulus:
:umbrella: The latest upstream changes (presumably dbafdc7) made this pull request unmergeable. Please resolve the merge conflicts.

vitiral:
I think this can be closed now

PR #16, opened 2017-06-21
Change traits to take &mut self instead of &self

f-bro:
See https://github.com/japaric/embedded-hal/issues/2#issuecomment-309663286

PR #19, opened 2017-09-15
Fix various errors in examples

hannobraun:
I've been working on picking up the stalled #16. Since this PR includes a change request regarding an example, I wanted to make sure to properly test all of them, but found out that they're kind of a mess. There were a lot of build errors that a user copying them into a new project would face.

I fixed a lot of those errors, and left others in there. The commits have more details about that. I do think that this PR, while clearly an improvement, is just papering over the deeper issue: That the examples are ignored (as in, not compiled during `cargo test`), and therefore prone to break and become out of date. I think we should find a way to change that, but that is beyond the scope of this humble pull request.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -403,14 +403,16 @@
 //!
 //! ``` ignore
 //! extern crate embedded_hal as hal;
+//! #[macro_use]
+//! extern crate nb;
 //!
 //! use hal::prelude::*;
 //!
 //! fn write_all<S>(serial: &S, buffer: &[u8]) -> Result<(), S::Error>
 //! where
-//!     S: hal::Serial
+//!     S: hal::serial::Write<u8>
 //! {
-//!     for byte in buffer {
+//!     for &byte in buffer {
 //!         block!(serial.write(byte))?;
 //!     }
 //!
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -508,9 +508,9 @@
 //!
 //! use hal::prelude::*;
 //!
-//! fn flush(serial: &S, cb: &mut CircularBuffer) -> Result<(), S::Error>
+//! fn flush<S>(serial: &S, cb: &mut CircularBuffer) -> Result<(), S::Error>
 //! where
-//!     S: hal::Serial,
+//!     S: hal::serial::Write<u8>,
 //! {
 //!     loop {
 //!         if let Some(byte) = cb.peek() {
@@ -532,12 +532,12 @@
 //!
 //! // NOTE private
 //! static BUFFER: Mutex<CircularBuffer> = ..;
-//! static SERIAL: Mutex<impl hal::Serial> = ..;
+//! static SERIAL: Mutex<impl hal::serial::Write<u8>> = ..;
 //!
 //! impl BufferedSerial {
-//!     pub fn write(&self, byte: u8) {
+//!     pub fn write(&self, bytes: &[u8]) {
 //!         let mut buffer = BUFFER.lock();
-//!         for byte in byte {
+//!         for byte in bytes {
 //!             buffer.push(*byte).unwrap();
 //!         }
 //!     }
@@ -545,10 +545,9 @@
 //!     pub fn write_all(&self, bytes: &[u8]) {
 //!         let mut buffer = BUFFER.lock();
 //!         for byte in bytes {
-//!             cb.push(*byte).unwrap();
+//!             buffer.push(*byte).unwrap();
 //!         }
 //!     }
-//!
 //! }
 //!
 //! fn interrupt_handler() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,16 +190,21 @@
 //! implementation:
 //!
 //! ``` ignore
+//! extern crate embedded_hal as hal;
+//! extern crate nb;
+//!
 //! /// A synchronized serial interface
 //! // NOTE This is a global singleton
-//! pub struct Serial1;
+//! pub struct Serial1(Mutex<..>);
 //!
 //! // NOTE private
 //! static USART1: Mutex<_> = Mutex::new(..);
 //!
-//! impl hal::Serial for Serial {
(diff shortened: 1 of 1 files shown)


hannobraun:
I've pushed some additional fixes for problems that I found while working on #20. I've amended the existing commits, no new commits have been added.

src/lib.rs
```diff
@@ -179,16 +190,21 @@
 //! implementation:
 //!
 //! ``` ignore
+//! extern crate embedded_hal as hal;
+//! extern crate nb;
+//!
 //! /// A synchronized serial interface
 //! // NOTE This is a global singleton
-//! pub struct Serial1;
+//! pub struct Serial1(Mutex<..>);
```
japaric: If this contains a field then it's no longer a global singleton. It's also not required to have this field. See below

src/lib.rs
```diff
@@ -179,16 +190,21 @@
...
 //! static USART1: Mutex<_> = Mutex::new(..);
 //!
-//! impl hal::Serial for Serial {
-//!     fn read(&self) -> Result<u8, nb::Error<Error>> {
-//!         hal::Serial::read(&Serial(USART1.lock()))
+//! impl hal::serial::Read<u8> for Serial1 {
+//!     type Error = !;
+//!
+//!     fn read(&self) -> Result<u8, nb::Error<Self::Error>> {
+//!         hal::serial::Read::read(&Serial1(USART1.lock()))
```
japaric: This can use the `Serial` struct from before like this `hal::serial::Read::read(&Serial(&*USART1.lock()))`

src/lib.rs
```diff
@@ -242,40 +258,46 @@
...
+//!     Future,
+//! };
+//! use futures::future::Loop;
 //! use stm32f103xx_hal_impl::{Serial, Timer};
 //!
 //! /// `futures` version of `Timer.wait`
 //! ///
 //! /// This returns a future that must be polled to completion
-//! fn wait<T>(timer: &T) -> impl Future<Item = (), Error = !>
+//! fn wait<T>(timer: T) -> impl Future<Item = (), Error = !>
```
japaric: With this signature the function will destroy the timer object. The pattern that the futures ecosystem uses to work around this issue is to return the input by value in `Item`. Thus the return type should be `impl Future<Item = T, Error = !>`.

src/lib.rs
```diff
@@ -242,40 +258,46 @@
...
+//!     future::poll_fn(move || {
 //!         Ok(Async::Ready(try_nb!(timer.wait())))
 //!     })
 //! }
 //!
 //! /// `futures` version of `Serial.read`
 //! ///
 //! /// This returns a future that must be polled to completion
-//! fn read<S>(serial: &S) -> impl Future<Item = u8, Error = Error>
+//! fn read<S>(serial: S) -> impl Future<Item = u8, Error = S::Error>
```
japaric: same comment as above

src/lib.rs
```diff
@@ -242,40 +258,46 @@
...
+//!     future::poll_fn(move || {
 //!         Ok(Async::Ready(try_nb!(serial.read())))
 //!     })
 //! }
 //!
 //! /// `futures` version of `Serial.write`
 //! ///
 //! /// This returns a future that must be polled to completion
-//! fn write<S>(byte: u8) -> impl Future<Item = (), Error = Error>
+//! fn write<S>(serial: S, byte: u8) -> impl Future<Item = (), Error = S::Error>
```
japaric: same comment as above

src/lib.rs
hannobraun: Ah, got it. I missed the point of that example before.

src/lib.rs
hannobraun: I don't think this pattern will work with `poll_fn`. Here's what I got now:
``` rust
fn wait<T>(timer: T) -> impl Future<Item = T, Error = !>
where
    T: hal::Timer,
{
    future::poll_fn(|| {
        try_nb!(timer.wait());
        Ok(Async::Ready(timer))
    })
}
```
And here's how it fails:
```
error[E0507]: cannot move out of captured outer variable in an `FnMut` closure
  --> src/lib.rs:29:25
   |
23 | fn wait<T>(timer: T) -> impl Future<Item = T, Error = !>
   |            ----- captured outer variable
...
29 |         Ok(Async::Ready(timer))
   |                         ^^^^^ cannot move out of captured outer variable in an `FnMut` closure
```
Which makes sense, as `poll_fn` takes an `FnMut` closure.
Of course I could re-create `poll_fn`/`PollFn`, but this seems a bit to involved for an example like this. What do you think?

hannobraun:
Thanks for the review! I've pushed a commit addressing your first comment. I didn't adress your comments on the futures example (see the comment I left in reply to yours).

src/lib.rs
japaric: Yeah, I *think* you'll have to use [loop_fn](https://docs.rs/futures/0.1.16/futures/future/fn.loop_fn.html) instead.

hannobraun: Thanks for the help! I think I got it now.

hannobraun:
I've amended my last commit and force-pushed. It addresses all the review comments now.

In case the changes are okay: Do you want me to leave it like that, or should I squash the changes from the last commit into the original commits?

japaric:
Thanks, @hannobraun.

@homunkulus r+

homunkulus:
:pushpin: Commit dbafdc7 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit dbafdc7 with merge dbafdc7...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/embedded-hal/builds/278732746?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing dbafdc79a0edfe38ea4797dcf62ef67c7f56b267 to master...

hannobraun:
Thank you!

PR #26, opened 2017-12-30
WIP: Attempt at DMA traits

nikhilkalige:
This does not work in any manner.

I am working on a Persistence of vision project of mine. I somehow need DMA to get the data rate high. This is my attempt at getting it to work. 

I have tried to replicate the DMA design from the bluepill crate. I liked the move semantics used there.  However I am not sure to get it working when you have `serial` stored in a structure. 

Hopefully someone who is better than me at rust, can give some ideas and suggestions.. :)

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -17,7 +17,7 @@ pub enum Error {
 /// DMA Transfer future
 pub trait Transfer {
     /// Return type
-    type Item: ?Sized;
+    type Item;
     /// Return type
     type Payload;
 
@@ -28,7 +28,5 @@ pub trait Transfer {
     fn is_done(&self) -> Result<bool, Error>;
 
     /// Block
-    fn wait(self) -> Result<(Static<Self::Item>, Self::Payload), Error>
-    where
-        Self::Item: Sized;
+    fn wait(self) -> Result<(Self::Item, Self::Payload), Error>;
 }
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -1,0 +1,34 @@
+//! DMA Interface
+
+/// Static alias for DMA
+pub type Static<T> = &'static mut T;
+
+/// DMA Errors
+#[derive(Debug)]
+pub enum Error {
+    /// Previous data got overwritten before it could be read because it was
+    /// not accessed in a timely fashion
+    Overrun,
+    /// Transfer error
+    Transfer,
+}
+
+
+/// DMA Transfer future
+pub trait Transfer {
+    /// Return type
+    type Item: ?Sized;
+    /// Return type
+    type Payload;
+
+    /// Get buffer
+    fn deref(&self) -> &Self::Item;
+
+    /// Check completion
+    fn is_done(&self) -> Result<bool, Error>;
+
+    /// Block
+    fn wait(self) -> Result<(Static<Self::Item>, Self::Payload), Error>
+    where
+        Self::Item: Sized;
+}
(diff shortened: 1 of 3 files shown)


nikhilkalige:
I have implemented DMA for Spi in [nikhilkalige/stm32f411-bsp](https://github.com/nikhilkalige/stm32f411-bsp/blob/singletons/src/spi.rs)

I have then tried to use the Spi in a driver for TLC5955 in [nikhilkalige/tlc5955](https://github.com/nikhilkalige/tlc5955/blob/master/src/lib.rs#L229)

nikhilkalige:
Hmm.. May be it would make sense to use something like a `Cell` or `RefCell` to store a reference to `spi` in the driver.

homunkulus:
:umbrella: The latest upstream changes (presumably #27) made this pull request unmergeable. Please resolve the merge conflicts.

src/dma.rs
```diff
@@ -0,0 +1,32 @@
+//! DMA Interface
+
+/// Static alias for DMA
+pub type Static<T> = &'static mut T;
```
ivegotasthma: I don't see this alias being used anywhere

src/spi.rs
```diff
@@ -33,6 +34,46 @@ pub trait FullDuplex<Word> {
...
+}
+
+
+/// DMA Write mode
+pub trait DmaRead<Word> {
+    /// Return type
+    type Transfer: dma::Transfer + ?Sized;
+
+    /// Recieve `words` from the slave.
+    fn recieve_dma<Buffer, Payload>(self, words: &'static mut Buffer) -> Self::Transfer
```
astro: Spelling: `receive_dma`

src/spi.rs
```diff
@@ -33,6 +34,46 @@ pub trait FullDuplex<Word> {
...
+    where
+        Buffer: Unsize<[Word]>;
+}
+
+/// DMA Write mode
```
astro: Documentation needs to be updated

src/dma.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// DMA Transfer future
+pub trait Transfer {
+    /// Return type
+    type Item;
+    /// Return type
+    type Payload;
+
+    /// Get buffer
+    fn deref(&self) -> &Self::Item;
```
astro: Does it make sense to allow access to the data while the transfer is still ongoing?

src/dma.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// Transfer error
+    Transfer,
+}
+
+
+/// DMA Transfer future
+pub trait Transfer {
+    /// Return type
+    type Item;
+    /// Return type
```
astro: Documentation could be more helpful. What's the difference between `Item` and `Payload`?

astro:
I'm trying to [implement DMA for the STM32F429](https://github.com/astro/stm32f429-hal/blob/master/src/dma.rs). As my use-case is (gapless) audio, I'm interested in using it in a continuous (double-buffered/circular), not one-shot mode.

japaric:
Thanks for the PR @nikhilkalige.

I think we should first design an API that meets all the requirements in https://github.com/japaric/embedded-hal/issues/37#issuecomment-377823801 and then test it out of tree for a bit before landing any new struct / trait in this crate so I'm going to close this PR until we have a better idea of what pieces we want to land in this crate.

PR #36, opened 2018-02-01
Add InputPin trait

thejpster:
I needed a basic InputPin trait for the buttons on my Stellaris Launchpad.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,8 @@
 [package]
-authors = ["Jorge Aparicio <jorge@japaric.io>"]
+authors = [
+	"Jorge Aparicio <jorge@japaric.io>",
+	"Jonathan 'theJPster' Pallant <github@thejpster.org.uk>"
+]
 categories = ["asynchronous", "embedded", "hardware-support", "no-std"]
 description = " A Hardware Abstraction Layer (HAL) for embedded systems "
 documentation = "https://docs.rs/embedded-hal"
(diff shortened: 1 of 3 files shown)


therealprof:
Hm, I thought I mentioned somewhere that this would be sorely missing but I can't find where. 😄 

Much appreciated!

dbrgn:
I'm also looking for a way to read the state of a GPIO pin :)

japaric:
Thanks, @thejpster 

@homunkulus r+

I read some comments on IRC I think there's some confusion about the role of the HAL traits. You don't **need** an InputPin trait to provide input pin functionality. It's *totally fine* to provide functionality using inherent methods like this:

``` rust
impl<MODE> PA0<Input<MODE>> {
    pub fn is_low(&self) -> bool { /* .. */ }
    pub fn is_high(&self) -> bool { /* .. */ }
}
```

In fact, I prefer this form because then you don't need to import a trait to be able to use the method (thankfully, we have preludes). The main reason you'll want to provide the functionality through a trait is to be able to use it for generic programming, i.e. as a bound in some generic driver. (The other reason could be that you want to create a `Vec<Box<InputTrait>>` but that's unlikely in bare metal context).

homunkulus:
:pushpin: Commit 278145a has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 278145a0d0517d910267a11a83348fe7ed5df7f0 with merge 54e0a1f94c2c2c1ae8965826e9eacd67346ef2f6...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/embedded-hal/builds/337792020?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 54e0a1f94c2c2c1ae8965826e9eacd67346ef2f6 to master...

therealprof:
@japaric Indeed. It is useful to bind it as an used "interrupt" pin to a driver to make sure it's correctly configured and cannot be used otherwise by accident. I'm curious how we could actually establish the connection with a real interrupt though since that setup is still somewhat messy and manual.

PR #44, opened 2018-02-25
set_state and toggle for digital::OutputPin

burrbull:
set_state and toggle with default implementation

### DIFF
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -25,7 +25,12 @@ pub trait OutputPin {
         }
     }
     
-    /// Toggles the pin state
+    /// Gets pin's state
+    fn get_state(&self) -> bool {
+        !self.is_low()
+    }
+    
+    /// Toggles pin's state
     fn toggle(&mut self) {
         if self.is_low() {
             self.set_high();
@@ -43,4 +48,9 @@ pub trait InputPin {
 
     /// Is the input pin low?
     fn is_low(&self) -> bool;
+    
+    /// Gets pin's state
+    fn get_state(&self) -> bool {
+        !self.is_low()
+    }
 }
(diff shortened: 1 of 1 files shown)


japaric:
I personally feel that the `get_state` methods don't add expressivity to the API. On the contrary,
using them makes the code less readable, IMO:

``` rust
if pa0.get_state() { // what do you mean by state? input state? output state? floating state?
    // ..
}

if pa0.is_high() { // OTOH, this is totally unambiguous
    // ..
}
```

Now if `get_state` returned a `State` enum then it would make more sense but I feel that a two
variant enum is a bit overkill in this case.

Both `get_state`s get a :-1: from me in their current state.

I'm of the (unpopular) opinion that `set_state(bool)` decreases readability but I understand that
people want to be able to easily write stuff like `pa0.set_state(pa1.is_high())`. I personally would
write:

``` rust
if pa1.is_high() {
    pa0.set_high()
} else {
    pa0.set_low()
}
```

Yes, it's longer but it's pretty clear what it does.

I give `set_state` a zero; I'm not opposed to merging it.

`toggle` gets a :+1: from me

---

And a remainder: adding methods to a trait is a breaking change *unless* the methods have a default implementation.

burrbull:
Will `set_value` be better?

therealprof:
@japaric Usually the state represents the current interpretation, i.e. the level while the "state" in your reading is called the mode, sometimes also separated into administrative mode and operational mode.

wez:
I came here to propose adding a `toggle_level` method and found this PR.

FWIW, over time I have come to the opinion that boolean input parameters are to be avoided because the intent is not clear.  These days I try to prefer using an enum.   In this case I would humbly advocate for `get_level` and `set_level` methods that return/accept an `enum { Low, High }` value.

Can we split out the `toggle` functionality as a separate PR since that one seems not to be controversial and could be accepted ahead of the rest of this discussion?

Issue #38, opened 2018-02-13
`Rng` API


therealprof:
I'd like to suggest to suggest a new `Rng` API for a common way of getting random numbers from devices with some source of randomness.

I've already implemented this as a proof-of-concept in my `nrf51-hal` crate (https://github.com/therealprof/nrf51-hal) and an example use in my `microbit` crate (https://github.com/therealprof/microbit).

The suggested (blocking) API is similar to the `I2c` API in that a mutable u8 slice is passed as parameter which will be filled with the random data and a Result is returned:
```
//! Blocking hardware random number generator

/// Blocking read
pub trait Read {
    /// Error type
    type Error;

    /// Reads enough bytes from hardware random number generator to fill `buffer`
    fn read(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error>;
}
```

Potentially this could also use an implementation for other (or even variable) array types, e.g. the `rand` crate requires a u32 slice as a seed.

I'd be happy to supply a PR for the the implementation.

nickray:
I'm implementing this for STM32L4(32): https://github.com/nickray/stm32l4-hal/blob/rng/src/rng.rs
It "works", but I'd like to make it rustic. Some questions:

- should `RngExt.enable` be called `.constrain` in line with other traits? For me personally, that naming is a little weird (as the API is constrained, not the RNG), but maybe it just needs getting used to and having conventions is a good thing.
- turning HSI48 on depends on other RCC settings, so it should probably not be done implicitly in `.enable`. On the other hand, I'm not happy with either a) asserting, or b) not checking settings at all
- is there room for additional helper methods? I'm talking about things like wrapping `self.rng.sr.read().cecs().bit()` as `.is_clock_error` for people like me who don't know data sheets and register names by heart. Or does this go against HAL minimalism?
- the `Error` enum: 
  - should this have only standardized errors, or reveal MCU-specific errors also?
  - is it efficient to check clock settings here on every call (admittedly, calling the RNG is not the fastest operation)?

eldruin:
This trait was added. There is discussion about replacing/complementing this with `rand_core` at #128 which provide further methods.

therealprof:
Hah! I knew there were previous discussions about this. Let's close this then.

PR #45, opened 2018-02-25
Added new (yet unproven) blocking Read trait for a rngs

therealprof:
Implements #38 . Identical custom trait already implemented in `nrf51-hal` crate and used in 
examples (directly and to seed a PRNG) in the `microbit` crate.

### DIFF
--- a/src/blocking/rng.rs
+++ b/src/blocking/rng.rs
@@ -1,6 +1,7 @@
 //! Blocking hardware random number generator
 
 /// Blocking read
+#[cfg(feature = "unproven")]
 pub trait Read {
     /// Error type
     type Error;
--- a/src/blocking/rng.rs
+++ b/src/blocking/rng.rs
@@ -7,5 +7,11 @@ pub trait Read {
     type Error;
 
     /// Reads enough bytes from hardware random number generator to fill `buffer`
+    ///
+    /// If any error is encountered then this function immediately returns. The contents of buf are
+    /// unspecified in this case.
+    ///
+    /// If this function returns an error, it is unspecified how many bytes it has read, but it
+    /// will never read more than would be necessary to completely fill the buffer.
     fn read(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error>;
 }
--- a/src/blocking/rng.rs
+++ b/src/blocking/rng.rs
@@ -1,0 +1,10 @@
+//! Blocking hardware random number generator
+
+/// Blocking read
+pub trait Read {
+    /// Error type
+    type Error;
+
+    /// Reads enough bytes from hardware random number generator to fill `buffer`
+    fn read(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error>;
+}
--- a/src/blocking/mod.rs
+++ b/src/blocking/mod.rs
@@ -8,3 +8,4 @@ pub mod delay;
 pub mod i2c;
 pub mod serial;
 pub mod spi;
+pub mod rng;
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -23,6 +23,8 @@ pub use ::blocking::spi::{
     Transfer as _embedded_hal_blocking_spi_Transfer,
     Write as _embedded_hal_blocking_spi_Write,
 };
+#[cfg(feature = "unproven")]
+pub use ::blocking::rng::Read as _embedded_hal_blocking_rng_Read;
 pub use ::digital::OutputPin as _embedded_hal_digital_OutputPin;
 #[cfg(feature = "unproven")]
 pub use ::digital::InputPin as _embedded_hal_digital_InputPin;


ryankurte:
I'm not immediately clear on how this is to be used, is the intention with this to block until entropy is available or return an error in the case it's not? 
And given entropy is often limited on the platforms we use, is there a reason this is better served by a blocking call than by polling on a non-blocking one (which is then easily wrapped to block)?

therealprof:
> I'm not immediately clear on how this is to be used, is the intention with this to block until entropy is available or return an error in the case it's not?

Almost. The Error is meant for situations where no entropy can be generated at all, e.g. if there's not random generator or it is not available for some reason (like being used by another hardware block).

> And given entropy is often limited on the platforms we use, is there a reason this is better served by a blocking call than by polling on a non-blocking one (which is then easily wrapped to block)?

Well, if you need entropy then you often need a specific amount of it and I find it a bit inconvenient to handle such a situation with async calls other than blocking on the spot. Different platform will also yield a different numbers of bytes per round, e.g. with NRF51 you'll get 1 byte per round while the STM32s will yield 4 byte per round.

If you need a larger amount of data I'd recommend setting up an interrupt handler and simply using the blocking call with the amount of data a platform will yield. Of course we still don't have any support whatsoever to properly include interrupt setup and handling in the HAL drivers so that requires some dancing around...

One could certainly craft an async interface, too, but at the moment I don't see too much benefit in that.

src/blocking/rng.rs
```diff
@@ -0,0 +1,10 @@
+//! Blocking hardware random number generator
+
+/// Blocking read
+pub trait Read {
```
austinglaser: This trait should be gated behind `#[cfg(feature = "unproven")]`, in addition to the gate in `prelude`.

At least, that's what I gather from the crate docs and the other unproven traits (like `digital::InputPin`)

austinglaser:
The benefit I can see to an async interface is interaction with nonblocking application models, such as those based on `await!` or futures.

The [`nb`](https://docs.rs/nb/0.1.1/nb/) crate is designed for exactly this kind of interoperability, and making the trait itself compatible should be as simple as replacing the `Result` with an `nb::Result`

src/blocking/rng.rs
therealprof: That can certainly be done. However the rules say that a trait is proven if at least two implementations and a driver or demo application exists and that is actually already the case. 😉

therealprof:
@austinglaser And where would you store the data in the meantime? The API is modelled like the other blocking APIs in that you pass in a buffer to be filled. Otherwise you can only return a single value in the result.

austinglaser:
@therealprof That's a good point -- I hadn't thought through that aspect enough.

What about a trait (eventually, and in addition to this one) which returns a fixed-size random value?

That fixed-size value could be (taking a page out of the `spi` traits' book) determined on a per-implementation basis, or it could simply be a `u8` -- meaning that on hardware like the STM32, you might just hold on to three bytes of data to give out on later calls.

I think I favor the simple `u8` approach, because it makes filling a buffer with values simple at the cost of a little more complexity for the implementer.

Something like this?

```Rust
// rng.rs
pub trait Read {
    type Error;

    fn read(&mut self) -> nb::Result(u8, Self::Error);
}
```

But this is getting beyond the scope of discussion for this particular pull request, I think.

therealprof:
@austinglaser A single `u8` might be somewhat limiting if that's not what you need in a go. I'm not sure how you would store the data internally in an implementation; once you leave the function any previous values stored on the stack would be gone, unless you have a (yuck!) static buffer.

austinglaser:
@therealprof I'd say that forcing a user to block is limiting in a different way -- at the trait level, at least, it makes sense to allow both usage patterns.

> I'm not sure how you would store the data internally in an implementation

You'd store the data in a context struct you pass back into the function each time. If that doesn't sound terribly ergonomic, you could use [futures](https://docs.rs/futures/0.1.18/futures/) or [generators](https://doc.rust-lang.org/std/ops/trait.Generator.html), which do much the same thing behind the scenes but (especially in the generator case) are a little cleaner to use.

austinglaser:
For a semi real-world example, take a look at the [docs](https://docs.rs/nb/0.1.1/nb/) for the nb crate. Note that the LED "thread" in each case has some mutable data (`state`) which is maintained across resumptions/polls of the asynchronous function.

therealprof:
> I'd say that forcing a user to block is limiting in a different way -- at the trait level, at least, it makes sense to allow both usage patterns.

Blocking on a MCU is relative anyway since you can always interrupt the wait. But sure, if you have a use case for partial random data being available whenever it shows up that an async interface can easily be added.

> You'd store the data in a context struct you pass back into the function each time. If that doesn't sound terribly ergonomic, you could use futures or generators, which do much the same thing behind the scenes but (especially in the generator case) are a little cleaner to use.

You're absolutely right that this is not ergonomic. ;) I'm familiar with the concepts of futures and generators in general. I'm still not sure about the use case in with a TRNG; maybe if you throw in DMA to automatically build up a pool from which you can deplete. But with a register based implementation there's very little point in having this; instead if you need large amount of data, use the TRNG to seed a PRNG and use that to generate random numbers whenever you need them.

But hey, feel free to propose the non-blocking version of this trait. I'd even provide a non-blocking implementation, just for the fun of it. 😉

austinglaser:
> But hey, feel free to propose the non-blocking version of this trait. I'd even provide a non-blocking implementation, just for the fun of it.

Fair enough!

japaric:
Oh and kind of off-topic but there are a few traits that look very similar to each other and that look similar to some of the API in `std::io::{Read,Write}` so I've been wondering if it would make sense to have some core `blocking::ReadExact` trait with method `fn read_exact(&mut self, buf: &mut [u8]) -> Result<(), Self::Error>` and then either (a) have traits like `blocking::Rng` be a super trait over that or (b) have `blocking::Rng` become a marker trait then the user would have to use `where RNG: blocking::ReadExact + Rng`.

therealprof:
@japaric Would you like me to change the name to `read_exact()` to be in line with the std::io::Read implementation?

PR #56, opened 2018-03-05
Add nonblocking RNG trait

austinglaser:
As discussed in #45, this is a nonblocking variant of an RNG interface.

### DIFF
--- a/src/rng.rs
+++ b/src/rng.rs
@@ -1,0 +1,17 @@
+//! Random Number Generator Interface
+
+#[cfg(feature = "unproven")]
+use nb;
+
+/// Nonblocking stream of random bytes.
+#[cfg(feature = "unproven")]
+// reason: No implementation or users yet
+pub trait Read {
+    /// An enumeration of RNG errors.
+    ///
+    /// For infallible implementations, will be `Infallible`
+    type Error;
+
+    /// Get a number of bytes from the RNG.
+    fn read(&mut self, buf: &mut [u8]) -> nb::Result<usize, Self::Error>;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
-
+- A nonblocking trait for interfacing with random number generation hardware.
 
 ### Changed
 - The current versions of `InputPin` have been proven. These are `digital::v1::InputPin` 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -689,6 +689,7 @@ pub mod blocking;
 pub mod digital;
 pub mod fmt;
 pub mod prelude;
+pub mod rng;
 pub mod serial;
 pub mod spi;
 pub mod timer;
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -25,6 +25,8 @@ pub use digital::OutputPin as _embedded_hal_digital_OutputPin;
 #[cfg(feature = "unproven")]
 #[allow(deprecated)]
 pub use digital::ToggleableOutputPin as _embedded_hal_digital_ToggleableOutputPin;
+#[cfg(feature = "unproven")]
+pub use rng::Read as _embedded_hal_rng_Read;
 pub use serial::Read as _embedded_hal_serial_Read;
 pub use serial::Write as _embedded_hal_serial_Write;
 pub use spi::FullDuplex as _embedded_hal_spi_FullDuplex;


Nemo157:
You could take a hint from `std::io::Read` and have the function return the number of bytes written to the buffer to support RNG devices that can generate >1 byte at a time.

```rust
/// Pull some bytes from this source into the specified buffer,
/// returning how many bytes were read.
///
/// If an object needs to block for a read but cannot it will return
/// an `Err(nb::Error::WouldBlock)` return value.
///
/// If the return value of this method is `Ok(n)`, then it must be
/// guaranteed that `0 <= n <= buf.len()`. A nonzero `n` value
/// indicates that the buffer buf has been filled in with `n` bytes
/// of data from this source.
fn read(&mut self, buf: &mut [u8]) -> nb::Result<usize, Self::Error>
```

In fact, it seems like `embedded-hal` should include some sort of core `Read` and `Write` traits similar to the `std::io` ones, then RNG devices could just implement those instead of having to have their own dedicated traits.

---

Looking a bit wider, it seems https://github.com/rust-lang-nursery/rand is `no_std` compatible. I haven't looked into whether it is actually suitable for implementing on microcontrollers, but it seems like integrating that might be a good way to have embedded RNG's work better with the wider ecosystem.

therealprof:
> fn read(&mut self, buf: &mut [u8]) -> nb::Result<usize, Self::Error>

Well, one could do this but would it be very user friendly? I have some doubts about that. If you need 3 bytes and only get 1 per call, what would you do? Slice'n'dice the buffer and call again until the 3 bytes are complete?

> Looking a bit wider, it seems https://github.com/rust-lang-nursery/rand is no_std compatible. I haven't looked into whether it is actually suitable for implementing on microcontrollers, but it seems like integrating that might be a good way to have embedded RNG's work better with the wider ecosystem.

I'm using that and my proposed blocking version to seed it in an microbit crate example: https://github.com/therealprof/microbit/blob/master/examples/rng_hal_printrandserial.rs

And I fully agree that Read/Write implementations would be nice for some things, like Serial (as you can also see from the mentioned example 😉).

Nemo157:
> Well, one could do this but would it be very user friendly? I have some doubts about that. If you need 3 bytes and only get 1 per call, what would you do? Slice'n'dice the buffer and call again until the 3 bytes are complete?

Yep, but there could be more ergonomic APIs provided built on top of this low-level API. For example in the `tokio` ecosystem there is [`tokio_io::read_exact`](https://docs.rs/tokio-io/0.1.5/tokio_io/io/fn.read_exact.html) that will handle the slicing and re-invoking until you finish reading the buffer from a non-blocking `std::io::Read` and represents that as a single future.

Nemo157:
Taking a look at the implementation of `rand 0.5` (as of yet unreleased, but nearly release candidate) it definitely cannot support non-blocking RNG sources. It allows sources to return `NotReady` while they prepare, but isn't designed so that sources can return something like `WouldBlock` in between generating bytes. Once they're ready they need to be able to return a continuous stream of data.

therealprof:
@Nemo157 Not sure I follow. I definitely have no interest using `rand` backed directly from a TRNG on an MCU. The application would do little more than wait for more random bits to be spit out all the time. Those TRNGs are really more meant to seed a PRNG, maybe even continuously...

Nemo157:
I was hoping it would be possible to use something like the [`ReseedingRng`](https://github.com/rust-lang-nursery/rand/blob/master/src/reseeding.rs#L65) trait by providing a non-blocking `RngCore` implementation using hardware RNG as the `reseeder`. As I said above though, the `rand` crate just doesn't really support having a non-blocking `RngCore` implementation.

I guess what it could do is have an internal buffer that is filled via interrupts, then when a new seed is requested it's only returned if the buffer has been refilled enough. That seems a bit annoying to do though.

therealprof:
Yeah, but then again most of the non-blocking stuff is rather wasteful if you don't have a trigger. 
Re reseeding the documentation even says: 

> Reseeding is never strictly necessary. Cryptographic PRNGs don't have a limited number of bytes they can output, or at least not a limit reachable in any practical way. There is no such thing as 'running out of entropy'.

So why not do it as I do and just seed the RNG once in a blocking fashion during initialisation with TRNG data and let that be?

Nemo157:
> So why not do it as I do and just seed the RNG once in a blocking fashion during initialisation with TRNG data and let that be?

For my usecase I want to be generating the seed in parallel to other setup tasks and some radio communication to minimize wake time. I think avoiding `RngCore` for the TRNG and having some non-blocking API like this is fine for that, then once the seed is ready I can just create a PRNG from that. So basically like you're doing, but with a non-blocking TRNG.

Nemo157:
For comparison I've thrown up a quick implementation of an `nb` compatible `Read` trait and `read_exact` method: https://github.com/Nemo157/embedded-hal/compare/master...read-trait

austinglaser:
@Nemo157 I'm happy to modify the trait interface if you think there'd be a cleaner way to go about it. I think I'm with @therealprof in that an unknown-till-return data size won't work very well. But what about taking a page out of the `spi` traits, and having a variable word-size, that matches the underlying RNG?

```Rust
pub trait Read<Word> {
    type Error;

    fn read(&mut self) -> nb::Result<Word, Self::Error>;
}

```

austinglaser:
@Nemo157 Regardless of what we discuss here, those seem like hugely valuable traits to have available -- especially if they're implemented for serial interfaces and the like

Nemo157:
> But what about taking a page out of the spi traits, and having a variable word-size, that matches the underlying RNG?
> 
> ```rust
> pub trait Read<Word> {
>     type Error;
> 
>     fn read(&mut self) -> nb::Result<Word, Self::Error>;
> }
> ```

First reaction, that could be really annoying if you use a library that requires an `rng::Read<u8>` but your system's HAL only provides `rng::Read<u16>`. There could be adaptor structs for transforming them, but that's a lot of boilerplate code.

---

Thinking about it some more, having the implementation on a non-`u8` providing device simply store some state and shift out a byte at a time wouldn't be too bad (which, re-reading the blocking RNG PR I see you actually mentioned, https://github.com/japaric/embedded-hal/pull/45#issuecomment-370096314).

So, afterall I'm pretty happy with this implementation, and it should be pretty trivial to write an adaptor from this trait to a generic `io::Read` trait like I'm proposing.

austinglaser:
Cool! I love the idea of adapters to io-esque `Read`/`Write`

japaric:
> `pub trait Read<Word> {`

You could have an `Unsize<[u8]>` bound on `Word` which basically means that `Word` is some array `[u8; N]`. `Unsize` is kind of a hack until we get proper const generics but with that in place it would be possible to provide a default implemantion of the `blocking::Rng` trait proposed in #45 on top of this trait:

*warning* untested; may contain errors

``` rust
#![feature(unsize)]
#![no_std]

#[macro_use(block)]
extern crate nb;

use core::marker::Unsize;

pub trait Read {
    type Bytes: Unsize<[u8]>; // poor man's [u8; N]
    type Error;

    fn read(&mut self) -> nb::Result<Self::Bytes, Self::Error>;
}

mod blocking {
    pub trait ReadExact {
        type Error;

        fn read_exact(&mut self, buf: &mut [u8]) -> Result<(), Self::Error>;
    }

    mod read_exact {
        use core::cmp;

        pub trait Default {}

        impl<RNG, E> super::ReadExact for RNG
        where
            RNG: ::Read<Error = E> + Default,
        {
            type Error = E;

            fn read_exact(&mut self, buf: &mut [u8]) -> Result<(), E> {
                let mut i = 0;
                while i < buf.len() {
                    let bytes = block!(self.read())?;
                    let slice: &[u8] = &bytes;

                    let n = cmp::min(slice.len(), buf.len() - i);
                    buf[i..i + n].copy_from_slice(&slice[..n]);
                    i += n;
                }

                Ok(())
            }
        }
    }
}
```

austinglaser:
So an `impl` of this trait might look like the following?

```Rust
impl rng::Read for SomeHardwareRng {
    type Bytes = [u8; 4];
    type Error = SomeError;

    // ...
}
```

It does seem like doing this otherwise would violate the "no stored state" aspect of `nb::Result`

japaric:
> So an impl of this trait might look like the following?

Yes

> It does seem like doing this otherwise would violate the "no stored state" aspect of nb::Result

How so? `Bytes` is to be picked to match the number of random bytes the interface can generate "in one go", or rather it matches the internal buffer of the hardware interface. So the implementer doesn't have to store anything in the program state; it will always read out the whole RNG buffer.

This is no different than the `Rng<Word>` where `Word = {u8,u16,u32}` that was mentioned before; the difference is that `[u8; N]` is easier to copy into a bigger buffer (pool) in a generic fashion.

austinglaser:
@japaric I think I was unclear in my comment before -- I was in fact saying that some of the previous suggestions (giving out a byte at a time, and holding onto the rest of the data) *would* violate the "no stored state" aspect, while your suggestion looks ergonomic while not doing so.

I like it; I'll update the pull request. Any objection to me amending/force pushing to this branch vs. adding another commit?

japaric:
@austinglaser Thanks for updating the PR. As of the upcoming v0.2.0 release we are not going to add dependencies to unstable features because we want this crate to compile on the beta and stable channels.

Could you try replacing `Unsized` with [`GenericArray`](https://docs.rs/generic-array/0.11.0/generic_array/). japaric/l3gd20#4 should give you an idea of the required changes.

austinbes:
Totally! I'll try to get to this over the weekend.

hannobraun:
@austinglaser Are you still interested in updating this PR?

austinglaser:
Yes, though I'm not sure when I'll be able to get to it. My available free time has been quickly swallowed by other things.

With that said, the changes should be fairly minimal. I'll try to get it done within two weeks -- does that sound reasonable?

hannobraun:
@austinglaser That sounds great, thank you. Take as much time as you need!

austinglaser:
@therealprof Still around, still busy. I do still want to update this -- I'll try to make time soon™. Thanks for the ping to get this back on my radar

austinglaser:
OK, I updated this to use `GenericArray` and be based off the current master. I'm going to take a stab at the default blocking implementation this evening or tomorrow!

src/rng.rs
```diff
@@ -0,0 +1,19 @@
+//! Random Number Generator Interface
+
+use generic_array::{ArrayLength, GenericArray};
+use nb;
+
+/// Nonblocking stream of random bytes.
+// reason: No implementation or users yet
```
therealprof: Shouldn't there be a `cfg` gate instead of a comment?

austinglaser: I've got `#[cfg(feature="unstable")]` on the entire `rng` mod -- I'm happy to put it here as well, or to move this comment to that point. What would you prefer?

src/rng.rs
```diff
@@ -0,0 +1,19 @@
...
+pub trait Read<N>
```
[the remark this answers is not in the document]
austinglaser: I actually tried with an associated type first, but did it in the wrong way (I tried to use `GenericArray` rather than `ArrayLength` as the bound, which obviously won't work).

I think I'm actually in favor of the associated type method -- I don't see enough value in multiple implementations for the same type to justify the ergonomic penalty. I'll defer to whatever more experienced Rust devs have to say, though.

```Rust
/// Nonblocking stream of random bytes.
// reason: No implementation or users yet
pub trait Read {
    /// Number of bytes in a result
    type Size: ArrayLength<u8>;

    /// An enumeration of RNG errors.
    ///
    /// For infallible implementations, will be `Void`
    type Error;

    /// Get a number of bytes from the RNG.
    fn read(&mut self) -> nb::Result<GenericArray<u8, Self::Size>, Self::Error>;
}
```

src/rng.rs
therealprof: Unstable? ;)

For consistency reasons I'd prefer to have it right in front of the trait definition instead of this comment.

austinbes: > unstable

Whoops. Got it right in the PR, at least =P

I'll put it there

austinglaser: > unstable

Whoops. Got it right in the PR, at least =P

I'll put it there

(used the wrong account for the first comment)

Cargo.toml
```diff
@@ -20,6 +20,9 @@ version = "1.0.2"
 [dependencies.nb]
 version = "0.1.1"
 
+[dependencies.generic-array]
+version = "0.11.0"
```
hannobraun: Why this version? The newest version is `0.12.0`.

src/rng.rs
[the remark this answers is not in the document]
hannobraun: So, I thought some more about this. Let's take it step by step.

First, I've come to the conclusion that the type argument makes no sense here:
1. A consumer of this API that values portability should always choose a size of one, because larger sizes might not be available on a given platform.
2. That means that any implementation that can provide more bytes in one go also needs to provide implementations for fewer bytes. This seems tedious, and sounds like something an implementer might forget.

So, if any implementation needs to provide a trait implementation for one byte, and no consumer should expect an implementation to deliver more than one byte (unless they _absolutely_ need more than one byte at a time, but when would that be the case?), why are we bothering with a type argument at all, and not just use `u8`?

But let's say we find `u8` too limiting in the general case and want to use the hardware to its full extent. So an associated type it is. But:
1. As a consumer of the API, we don't know what this associated type will be, unless we specify it in a where clause. That would lead to the same problems the type argument had, except now it's worse, because implementations capable of providing more than one byte can't provide the 1-byte implementation as a fallback.
2. So we accept that we don't know what the type is and get back a `GenericArray`. What is the only thing we can do with that? Get a slice. What happens if that slice contains fewer bytes than we need? We note down how many bytes we got, then call again, until we have enough bytes.

This is very inconvenient, which is exactly the criticism that the initial proposal received. Except no, this time it's worse. At least in the initial proposal, the RNG put the bytes into our buffer. Now we have to maintain a buffer, and copy bytes out `GenericArrays` into it ourselves.

So... am I missing some deeper thought that went into this iteration of the proposal? Because if not, the initial proposal looks way more attractive to me.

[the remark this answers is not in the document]
austinglaser: Just to confirm, was the the original proposal you're referring to this?

```Rust
// rng.rs
pub trait Read {
    type Error;

    fn read(&mut self) -> nb::Result(u8, Self::Error);
}
```

I don't think you're missing any deeper reasoning, and I see your point entirely. I guess it's more-or-less premature optimization to resist throwing away extra data that's provided by some TRNGs.

I'm happy to roll back my PR to the above version. If nothing else, it avoids leaking hardware details through the trait.

One method that I tried to propose (but that I did in an unclear way) would be for the trait implementor itself to hold on to the extra bytes, and feed them out one at a time as they're requested. The obvious question becomes: where does that state get stored? My proposal is to store it in `self`.

It already seems like the accepted pattern to wrap register-blocks in new structural types (a-la `stm32f30x_hal::i2c:I2c`, stm32f30x_hal::spi::Spi`, etc). Trait-users already expect these implementations to be stateful -- in every case, there's implicit state stored in hardware registers.

On the other hand, I've heard it said elsewhere (though I can't immediately find it) that `embedded-hal` trait implementations must _only_ depend on the state in hardware registers. I'll admit that I don't entirely understand the logic behind this restriction. It also seems like there are crates that break this -- [`shared-bus`](https://crates.io/crates/shared-bus) being the first that comes to mind, since it potentially depends on an in-memory mutex.

I don't see this as making or breaking the case for a non-generic trait, but if HAL implementors at least had the _opportunity_ for this kind of optimization, then there's absolutely no argument to be made for the generic, variably-sized variant.

I feel like I've let this response ramble a little much, so here is the core of my thoughts:

- The only thing we lose by simply returning a `u8` is, as you say, "full use of the hardware." I agree that that's not worth it
- There's a potential implementation strategy which _does_ make full use of the hardware, while relying on storing non-register state in `self`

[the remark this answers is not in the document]
hannobraun: > Just to confirm, was the the original proposal you're referring to this?

Actually, I was referring to what @Nemo157 proposed:
``` rust
fn read(&mut self, buf: &mut [u8]) -> nb::Result<usize, Self::Error>
```
Sorry for referring to it as the initial proposal. For some reason I thought it was. At this point, this slice-based version is the one I prefer. It's simple, takes full advantage of hardware capabilities, and it's straight-forward to write a (blocking or non-blocking) `read_all` wrapper, that provides as many bytes as the user wants.

Of course, I might be missing something, or might have overseen something from the discussion here, so I'm open to being convinced of something else.

Regarding your 1-byte version:
``` rust
fn read(&mut self) -> nb::Result<u8, Self::Error>;
```

I think this is fine, too. Not my preferred version, but a good choice.

> One method that I tried to propose (but that I did in an unclear way) would be for the trait implementor itself to hold on to the extra bytes, and feed them out one at a time as they're requested.
> [...]
> On the other hand, I've heard it said elsewhere (though I can't immediately find it) that `embedded-hal` trait implementations must only depend on the state in hardware registers. I'll admit that I don't entirely understand the logic behind this restriction.

I think this approach is fine. I don't see why `embedded-hal` trait implementations shouldn't have internal (non-hardware) state. Maybe there were good reasons for that statement in its original context, but it certainly doesn't hold in the general case.

> I don't see this as making or breaking the case for a non-generic trait, but if HAL implementors at least had the opportunity for this kind of optimization, then there's absolutely no argument to be made for the generic, variably-sized variant.

Agreed.

To summarize my own position:
- I currently prefer the slice-based variant, for the reasons stated above.
- I find the one-byte variant to be a good option, too. I think caching bytes is a fine implementation strategy.
- I think this whole generic argument discussion (either via type argument or associated type) went entirely in the wrong direction.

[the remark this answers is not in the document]
therealprof: > On the other hand, I've heard it said elsewhere (though I can't immediately find it) that embedded-hal trait implementations must only depend on the state in hardware registers.

I think that may have been my statement but in a different context: Trait implementations should not rely on internal state representing physical state, e.g. that GPIOs are and stay in the physical state they were set to by an implementation. Buffering data in internal state should be fine, though we shouldn't overdo it...

hannobraun:
Okay, so we need to decide how to go forward here. To recap: The current approached (based on feedback previously provided on this pull request) doesn't seem viable (see [my comment](https://github.com/rust-embedded/embedded-hal/pull/56#discussion_r221420307)). Now we need to decide which interface we want to go forward with.

The options:
1. Single-byte:
``` rust
fn read(&mut self) -> nb::Result<u8, Self::Error>
```
2. Slice-based:
``` rust
fn read(&mut self, buf: &mut [u8]) -> nb::Result<usize, Self::Error>
```

I prefer option 2, but I'm fine with option 1.

My reasons for preferring option 2:
> It's simple, takes full advantage of hardware capabilities, and it's straight-forward to write a (blocking or non-blocking) read_all wrapper, that provides as many bytes as the user wants.

Any other views?

jamwaffles:
> Any other views?

Slices seem to be the most versatile to me, although I'm not sure how an implementation would signal to the user how many bytes will be put into the buffer. Is that even necessary? Should the user have to check the length of the buffer after it's written to? I think I've seen some other Rust embedded APIs that return the number of bytes read/received/whatever like this:

```rust
let num_recvd = TheRandomThing::read(&mut buf);
```

which would make sense.

hannobraun:
Thanks for the comment, @jamwaffles.

> Slices seem to be the most versatile to me, although I'm not sure how an implementation would signal to the user how many bytes will be put into the buffer.

The method signature from my comment includes a `usize` return value, so it would work exactly as you proposed.

> Is that even necessary? Should the user have to check the length of the buffer after it's written to?

It is necessary, as this is a non-blocking API which might not be able to fill the buffer in one call.

Aside: It might be nicer, if the method returned a slice with all the bytes written instead of a number. I'm not sure if that plays well with all use cases, however, and right now it would lead straight into lifetime hell anyway, I think (NLL should help with this though, once it arrives on stable). In any case, it's a topic for later.

jamwaffles:
> The method signature from my comment includes a usize return value, so it would work exactly as you proposed.

Ah, I should've skimmed the PR more slowly. In that case, this looks fine to me :)

austinbes:
OK @hannobraun, I'm sold. The clincher is that a simple implementation of option 1 is easy to make in terms of option 2, and that option 2 makes it easier to take full advantage of the hardware.

Will try to update this PR again (and drop in the default blocking implementation while I'm at it) over the weekend.

austinglaser:
OK @hannobraun, I'm sold. The clincher is that a simple implementation of option 1 is easy to make in terms of option 2, and that option 2 makes it easier to take full advantage of the hardware.

Will try to update this PR again (and drop in the default blocking implementation while I'm at it) over the weekend.

Disasm:
@austinglaser Any progress on this?

austinglaser:
@hannobraun @Disasm Apologies for the year-late update, but this should be up-to-date based on all the discussions.

I removed the `unproven` gate on the `rng` mod, since that appears to be the pattern elsewhere. Please let me know if that was the wrong way to do it.

hannobraun:
@austinglaser Thanks, this is great! Unfortunately, I no longer have the power to merge anything, so let's hope someone else will take a look soon :-)

no111u3:
It's lgtm.

src/rng.rs
```diff
@@ -0,0 +1,17 @@
...
+#[cfg(feature = "unproven")]
+use nb;
+
+/// Nonblocking stream of random bytes.
+#[cfg(feature = "unproven")]
+// reason: No implementation or users yet
+pub trait Read {
+    /// An enumeration of RNG errors.
+    ///
+    /// For infallible implementations, will be `Void`
```
Disasm:
```suggestion
    /// For infallible implementations, will be `Infallible`
```

Disasm:
@austinglaser sorry for the late response and thanks for the contribution! Could you apply the change above and add a CHANGELOG entry?

src/rng.rs
austinglaser: Done

austinglaser:
Added the changelog entry, and rebased on `master`

Disasm:
Thanks!

@therealprof could you review this?

PR #60, opened 2018-03-11
Add MCP3425 ADC driver to README

dbrgn:
As suggested in the README, here's a PR to add another driver implementation to the README :slightly_smiling_face:


japaric:
Thanks @dbrgn

PR #69, opened 2018-03-29
Correct link address for embedded-hal-impl

MajorBreakfast:


therealprof:
Unfortunately this keyword doesn't exist on crates.io yet.

MajorBreakfast:
True

And https://github.com/japaric/embedded-hal/blob/master/src/lib.rs#L50 should probably be corrected as well: It says `embedded-hal` instead of `embedded-hal-driver` or `embedded-hal-impl`. I like the distinction. But it'd be probably best to use both keywords, e.g. `embedded-hal`+`embedded-hal-driver`, on each package.

Issue #29, opened 2018-01-20
[RFC] `digital::IoPin`, pins that can switch between input and output modes at runtime


japaric:
Some people have expressed interest in a trait like this for writing generic LCD drivers but there are probably other use cases.

We were discussing this [on IRC] yesterday and I proposed three different APIs:

[on IRC]: https://bot.tockos.org/tockbot-moz/embedded-bot/2018-01-19/?msg=53920&page=2

### Result based API

``` rust
/// A pin that can switch between input and output modes at runtime
pub trait IoPin {
    /// Signals that a method was used in the wrong mode
    type Error;

    /// Configures the pin to operate in input mode
    fn as_input(&mut self);
    /// Configures the pin to operate in output mode
    fn as_output(&mut self);

    /// Sets the pin low
    fn set_low(&mut self) -> Result<(), Self::Error>;
    /// Sets the pin high
    fn set_high(&mut self) -> Result<(), Self::Error>;

    /// Checks if the pin is being driven low
    fn is_low(&self) -> Result<bool, Self::Error>;
    /// Checks if the pin is being driven high
    fn is_high(&self) -> Result<bool, Self::Error>;
}

// this function won't panic; LLVM should be able to opt way the panicking branches
fn example(mut io: impl IoPin) {
    io.as_input();
    if io.is_low().unwrap() {
        /* .. */
    }
    if io.is_high().unwrap() {
        /* .. */
    }

    io.as_output();
    io.set_low().unwrap();
    io.set_high().unwrap();
}
```

### Closure based API

``` rust
/// A pin that can switch between input and output modes at runtime
pub trait IoPin {
    /// Pin configured in input mode
    type Input: InputPin;

    /// Pin configured in output mode
    type Output: OutputPin;

    /// Puts the pin in input mode and performs the operations in the closure `f`
    fn as_input<R, F>(&mut self, f: F) -> R
    where
        F: FnOnce(&Self::Input) -> R;

    /// Puts the pin in output mode and performs the operations in the closure `f`
    fn as_output<R, F>(&mut self, f: F) -> R
    where
        F: FnOnce(&mut Self::Output) -> R;
}

fn example(mut io: impl IoPin) {
    io.as_input(|i| {
        if i.is_low() { /* .. */ }
        if i.is_high() { /* .. */ }
    });

    io.as_output(|o| {
        o.set_low();
        o.set_high();
    });
}
```

### Implicit / Automatic API

``` rust
[...]

Issue #31, opened 2018-01-20
[RFC] `digital::TriStatePin`


japaric:
What the title says. The use case for this is some sort of generic charlieplexing interface / driver / library.

### Vanilla

``` rust
/// A tri-state (low, high, floating) pin
pub trait TriStatePin {
    /// Drives the pin low
    fn set_low(&mut self);

    /// Drives the pin high
    fn set_high(&mut self);

    /// Puts the pin in floating mode
    fn float(&mut self);

    /// Checks if the pin is currently in floating mode
    fn is_floating(&self) -> bool;

    // plus maybe other `is_*` methods
}
```

### Enum based

``` rust
/// The states of a tri-state pin
enum State { Low, High, Floating }

/// A tri-state (low, high, floating) pin
pub trait TriStatePin {
    /// Changes the state of the pin
    fn set(&mut self, state: State);

    /// Gets the state of the pin
    fn state(&self) -> State;
}
```

Once we pick an alternative that we are happy with we can land it behind the "unproven" feature gate. Once someone has demonstrated that it works by actually implementing the trait and building some generic library on top of it we can un-feature gate it.

cc @therealprof

jamesmunns:
I have a library that could use this (dw1000 driver).

I think I like the second alternative better (enum based), though we would need a "get state" (output/input), and "get value" (low or high, could work for input or output).

bachp:
+1 for the enum based variant, it seems easier to comprehend

EugeneGonzalez:
Enum approach is more pleasing. One concern is the "state" method could be confused for an InputPin read instead of the TriStatePin's set output.

therealprof:
Whoops, commented on so topics about this very thing that I forgot to raise my voice here. 😉

+1 for enum based, too.

austinglaser:
I think a way to read the pin's state is critical -- with it, for instance, you could build a portable bit-banged I2C implementation on top of a `TriStatePin`.

What if the API looked something like:

```Rust
/// The states of a tri-state pin
enum State { Low, High, Floating }

/// A tri-state (low, high, floating) pin
pub trait TriStatePin {
    /// Changes the state of the pin
    fn set(&mut self, state: State);

    /// Gets the output state of the pin
    ///
    /// Queries the state set in `set()`
    fn state(&self) -> State;

    /// Whether the pin is driven high
    ///
    /// Reads the pin's actual electrical state
    fn is_high(&self) -> bool;

[...]

PR #72, opened 2018-04-07
Move reading functions from OutputPin trait to StatefulOutputPin

astro:
As it has come up in #29, here is a proposal for removing the burden of caching the state on implementations that cannot read it from an output pin.

### DIFF
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -11,7 +11,7 @@ pub trait OutputPin {
 
 /// Output pin that can read its output state
 #[cfg(feature = "unproven")]
-trait StatefulOutputPin {
+pub trait StatefulOutputPin {
     /// Is the pin set to high?
     fn is_set_high(&self) -> bool;
 
@@ -20,23 +20,74 @@ trait StatefulOutputPin {
 }
 
 /// Output pin that can be toggled
+///
+/// See [toggleable](toggleable) to use a software implementation if
+/// both [OutputPin](trait.OutputPin.html) and
+/// [StatefulOutputPin](trait.StatefulOutputPin.html) are
+/// implemented. Otherwise, implement this using hardware mechanisms.
 #[cfg(feature = "unproven")]
-trait ToggleableOutputPin {
-    /// Toggle pin output
+pub trait ToggleableOutputPin {
+    /// Toggle pin output.
     fn toggle(&mut self);
 }
 
 /// If you can read **and** write the output state, a pin is
-/// toggleable by software. You may override the `toggle()` method
-/// with a hardware implementation.
+/// toggleable by software.
+///
+/// ```
+/// use embedded_hal::digital::{OutputPin, StatefulOutputPin, ToggleableOutputPin};
+/// use embedded_hal::digital::toggleable;
+///
+/// /// A virtual output pin that exists purely in software
+/// struct MyPin {
+///     state: bool
+/// }
+///
+/// impl OutputPin for MyPin {
+///    fn set_low(&mut self) {
+///        self.state = false;
+///    }
+///    fn set_high(&mut self) {
+///        self.state = true;
+///    }
+/// }
+///
+/// impl StatefulOutputPin for MyPin {
+///    fn is_set_low(&self) -> bool {
+///        !self.state
+///    }
+///    fn is_set_high(&self) -> bool {
+///        self.state
+///    }
+/// }
+///
+/// /// Opt-in to the software implementation.
+/// impl toggleable::Default for MyPin {}
+///
+/// let mut pin = MyPin { state: false };
+/// pin.toggle();
+/// assert!(pin.is_set_high());
+/// pin.toggle();
+/// assert!(pin.is_set_low());
+/// ```
 #[cfg(feature = "unproven")]
-impl<PIN: OutputPin + StatefulOutputPin> ToggleableOutputPin for PIN {
-    /// Toggle pin output
-    fn toggle(&mut self) {
-        if self.is_set_low() {
-            self.set_high();
-        } else {
-            self.set_low();
+pub mod toggleable {
+    use super::{OutputPin, StatefulOutputPin, ToggleableOutputPin};
+
+    /// Software-driven `toggle()` implementation.
+    pub trait Default: OutputPin + StatefulOutputPin {}
+
+    impl<P> ToggleableOutputPin for P
+    where
+        P: Default,
+    {
+        /// Toggle pin output
+        fn toggle(&mut self) {
+            if self.is_set_low() {
+                self.set_high();
+            } else {
+                self.set_low();
+            }
         }
     }
 }
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -2,12 +2,6 @@
 
 /// Single digital output pin
 pub trait OutputPin {
-    /// Is the output pin high?
-    fn is_high(&self) -> bool;
-
-    /// Is the output pin low?
-    fn is_low(&self) -> bool;
-
     /// Sets the pin low
     fn set_low(&mut self);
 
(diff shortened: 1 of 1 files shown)


astro:
> Also are you planning to follow up with the additional traits?

Do you mean a `ReadableOutputPin` trait?

therealprof:
@astro Hm, we had a discussion somewhere (can't remember where exactly) where we discussed a slightly different variant with composable traits. I think @austinglaser made a nice proposal based on a discussion we had.

The "problem" I see with your approach is that switching modes between input and output might not be all the much useful and be quite a bit of overhead since that actually requires changing the GPIO mode (which also has some side effects like electrically switching from a set low/high level into high impedance mode) where usually one might only want to check the logic state of the pin (i.e. what do the registers say) **or** the real hardware state (what is the actual voltage level at the pin), both of which may or may not actually be implementable on some MCUs.

astro:
Please refresh your browser, I edited my comment several times. :) Sorry!

I think this and #73 are separate problems. You can implement `StatefulOutputPin` (which I pushed later) without switching modes.

therealprof:
Err, yes, my bad, the comment was really meant for #73. Let me move it there.

hannobraun:
@therealprof 
> The only problem I see is that this might potentially break existing implementations?

As I stated before elsewhere, I don't think we should be losing sleep over backwards-compatibility at this point. Yes, breaking implementations sucks, but I'm more concerned about whether embedded-hal is going to look good a few years from now.

therealprof:
@hannobraun That's fine with me, too. Just wanted to raise the concern others might have to feel whether there's consensus on that topic.

austinbes:
FWIW, I'm totally of @hannobraun's opinion. Break things early, if they're to break at all

src/digital.rs
```diff
@@ -2,19 +2,32 @@
...
+/// Output pin that can read its output state
+#[cfg(feature = "unproven")]
+trait StatefulOutputPin: OutputPin {
+    /// Is the pin set to high?
+    fn is_set_high(&self) -> bool;
+
+    /// Is the pin set to low?
+    fn is_set_low(&self) -> bool;
+
+    /// Toggle pin output
```
ryankurte: Sorry if I am late to the party. Can you override a method in a trait? On [some cores](https://github.com/ryankurte/efm32-base/blob/1a8d48edae129075b35cea5ac0a36badebbc0c1d/emlib/inc/em_gpio.h#L925) toggle is exposed as a separate register / no read state is required.

I'd propose having the toggle as a separate trait that you can implement on StatefulOutputPin if toggle function is not available.

therealprof: As discussed in #31 (et al) I also think that different capabilities should be modelled in different composable (and mode dependent) traits instead of being munged together into a few traits that do a lot of things at once. Something like `is_set_low()` on a `StatefulOutputPin` is not clear what it actually means, i.e. is that supposed to be the electrical or the logical state of the pin? For `toggle()` it has to be the latter, not the former, in other cases you might want to have the electrical state, though and cannot have this. Also (as you said) there are implementations allowing an efficient toggle without reading the logical state from a pin. There might even be implementations which do not allow you to read back the logical state, although I don't recall any from the top of my head.

I only agreed to moving it out of `OutputPin` because it belongs there even less and this change turns into into an `unproven` functionality so if it doesn't provide the hoped for benefit (as I suspect it will not) then we can replace it by something better.

astro: Sounds reasonable. I'll remove it in favour of adding `ToggleableOutputPin` later.

astro: > Can you override a method in a trait?

@ryankurte [Yes, you can.](https://doc.rust-lang.org/book/second-edition/ch10-02-traits.html#default-implementations)

astro:
I have moved `toggle()` into a separate `ToggleableOutputPin` trait with a default impl for `StatefulOutputPin`.

Also, I removed the `OutputPin` constraint from `StatefulOutputPin` as it is now optional and possible for theoretical output pins from which you can read but not write the state (eg. set by hardware).

src/digital.rs
```diff
@@ -2,19 +2,45 @@
...
+trait ToggleableOutputPin {
+    /// Toggle pin output
+    fn toggle(&mut self);
+}
+
+/// If you can read **and** write the output state, a pin is
+/// toggleable by software. You may override the `toggle()` method
+/// with a hardware implementation.
+#[cfg(feature = "unproven")]
+impl<PIN: OutputPin + StatefulOutputPin> ToggleableOutputPin for PIN {
```
austinglaser: Would it be good to add an empty `DefaultToggle` trait here, and thus make this implementation opt-in rather than opt-out?

It would match the way the blocking SPI traits are handled. On the other hand, this is a simpler implementation.

I'm not particularly attached to one way or the other, but I do think consistency throughout the `embedded-hal` will be valuable.

astro: Thanks for the suggestion. I added that now by copying the style of the opt-in default traits in `spi` and `serial`. I hope that the doc example is helpful.

astro:
Is there anything I can do to help this move forward?

src/digital.rs
```diff
@@ -2,19 +2,96 @@
...
-    fn is_low(&self) -> bool;
-
     /// Sets the pin low
     fn set_low(&mut self);
 
     /// Sets the pin high
     fn set_high(&mut self);
 }
 
+/// Output pin that can read its output state
```
therealprof: As I've mentioned before there's a distinction between a logical state and an electrical state. It should be made clear that this is meant to represent the logical (or register) state of a pin and not the electrical state.

src/digital.rs
```diff
@@ -2,19 +2,96 @@
...
+#[cfg(feature = "unproven")]
+pub trait StatefulOutputPin {
+    /// Is the pin set to high?
```
therealprof: A pin could also be logically floating so either it needs to be made clear that a tri-stated pin may not implement this trait or that the implicit assumption `p.is_set_high() == !p.is_set_low()` does not necessarily hold and should not be assumed.

japaric:
This seems OK to me but it should be explicitly noted that these traits should be implemented only for push pull output pins and that low / high are referring to the electrical output of the pin. That was missing in the original trait as well.

---

@hannobraun just for the record: releasing a new minor version doesn't "just breaks implementations"; it fragments the ecosystem. Once a HAL implementation moves from e.g. v0.1.x to v0.2.x *it can no longer be used with driver crates that depend on embedded hal v0.1.x*. This happens even if the trait that the driver crate uses as a bound has not changed at all between the two minor releases. This situation increases the maintenance work of driver authors which now have to publish a new minor version of their crate that supports the new embedded-hal and probably will also have to maintain the v0.1.x version until all / most-of the existing / active / maintained HAL implementations move to v0.2.x. So bumping the minor version shouldn't be done often / lightly; as the embedded-hal ecosystem grows we'll have to allow more "settling time" between each new minor version release.

That being said we have to make a new minor version release beacuse of #80 so let's include this PR which also contains breaking changes in v0.2.0. @astro could you update the docs?

therealprof:
> that low / high are referring to the electrical output of the pin.

I disagree. We can not make any assumption of the electrical state of a pin, only the logical state. Even if you drive it low on a push/pull pin there might be a stronger source keeping it high and vice versa.

japaric:
@therealprof point taken

I'll merge this PR as it is and adjust the docs locally before releasing v0.2.0. Thanks for the PR, @astro!

bors r+

hannobraun:
@japaric 
> just for the record: releasing a new minor version doesn't "just breaks implementations"; it fragments the ecosystem.

Good point, thanks for the explanation. While that doesn't change my stance about breaking changes in general, it's good to keep in mind that we need to release them strategically.

Issue #75, opened 2018-04-17
Watchdog timer trait.


xd009642:
Setting up a watchdog timer to reset a microcontroller if it isn't serviced at regular intervals is fairly common. Also, from my experience watchdogs tend to be fairly similar and minimalist making them a potentially easy inclusion for a trait. I haven't analysed a large number of controllers but the typical operation seems to be as follows:

1. Once enabled watchdog cannot be disabled
2. A bit must be set or cleared periodically or processor is reset
3. Period length may be configurable
4. May be able to see current timer value.

I think 1, 2 and 3 would be essential for a watchdog trait, via functions `start`, `kick` and `set_period`. Due to it on some microcontrollers only offering a subset of timer functionality I'm also not sure how much crossover there should be with the timer traits? Open to discussion this is just prompted by some work I've been doing recently with STM32 processors.

ryankurte:
Love the idea of this, though ime it is possible and there are a number of reasons you might want to disable the watchdog (for example, rebooting from an app with the watchdog to a boot-loader without, during long sleep states depending on architecture).
I personally would like to see disable added so we're not making application level design decisions.

austinbes:
@ryankurte -- We'd have to think carefully about adding disable functionality -- not because I disagree with your point, but because some hardware watchdog timers cannot be disabled.

I think it might sense to have a separate trait (`DisableableWatchdog` is a bit of a mouthful, but it's what I can come up with off the top o fmy head) which owns the `disable()` function.

ryankurte:
Good call, I think that's inline with the discussion on composable gpio traits also.

GGreenwood:
I'm also familiar with platforms where the recommended method for performing a software-reset was to enable the watchdog timer, then to stall until the timer overflowed.

Should this action be captured in the trait, or should it be performed by the application?

austinbes:
@GGreenwood What platform(s) are you talking about, specifically?

I don't think that's necessary or desirable to capture that trait in the application. What I can see being valuable is eventually adding a `Reset` (or similar) trait, which could be implemented in terms of the watchdog timer on platforms where that makes sense.

GGreenwood:
One other application comes to mind: A computationally-heavy library, or one that blocks to handle communication, may hog the processor for long enough to cause a watchdog reset. In this case, it would make sense for the library developer to access the Watchdog trait in order to reset it.

[...]

PR #76, opened 2018-04-18
Added initial watchdog proposal.

xd009642:
Added initial trait proposal for watchdog timer as mentioned in #75

### DIFF
--- a/src/watchdog.rs
+++ b/src/watchdog.rs
@@ -1,0 +1,32 @@
+//! Traits for interactions with a processors watchdog timer.
+
+
+
+/// Feeds an existing watchdog to ensure the processor isn't reset. Sometimes
+/// commonly referred to as "kicking" or "refreshing".
+#[cfg(feature = "unproven")]
+pub trait Watchdog {
+    /// Triggers the watchdog. This must be done once the watchdog is started
+    /// to prevent the processor being reset.
+    fn feed(&mut self);
+}
+
+
+/// Enables A watchdog timer to reset the processor if software is frozen or 
+/// stalled.
+#[cfg(feature = "unproven")]
+pub trait WatchdogEnable {
+    /// Unit of time used by the watchdog
+    type Time;
+    /// Starts the watchdog with a given period, typically once this is done 
+    /// the watchdog needs to be kicked periodically or the processor is reset. 
+    fn start<T>(&mut self, period: T) where T: Into<Self::Time>;
+}
+
+
+/// Disables a running watchdog timer so the processor won't be reset.
+#[cfg(feature = "unproven")]
+pub trait WatchdogDisable {
+    /// Disables the watchdog
+    fn disable(&mut self);
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -690,6 +690,7 @@ pub mod prelude;
 pub mod serial;
 pub mod spi;
 pub mod timer;
+pub mod watchdog;
 
 /// Input capture
 ///


src/watchdog.rs
```diff
@@ -0,0 +1,18 @@
+
+
+
+/// A watchdog timer to reset the processor if software is frozen or stalled.
+pub trait Watchdog {
```
austinbes: You're going to want to add `#[cfg(feature = "unproven")]` to this, since it's a new trait

src/watchdog.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// Unit of time used by the watchdog
+    type Time;
+    /// Starts the watchdog, typically once this is done the watchdog needs to
+    /// be kicked periodically or the processor is reset. Once started normally
+    /// cannot be stopped
+    fn start(&mut self);
+    /// Set the period of the watchdog 
+    fn set_period<T>(&mut self, period: T) where T: Into<Self::Time>;
```
austinbes: Is it expected that `Self::Time` handles the prevention of invalid periods? Otherwise, `set_period` would have to be fallible, I think.

Additionally, it seems like this might all be more ergonomic if there was no `set_period()`, and instead `start()` took the period as a parameter. With a set-and-forget design like this, I don't see a lot of benefit to separating the act of setting a period into its own operation, but I might be missing something

xd009642: Sure thing, I'll make these changes when I get home from work! With set_period I was considering whether anyone may want to change period after its started. But that may not be supported on everything so I'll change it :+1:

xd009642:
@austinbes So I've responded to the feedback as well as outcomes of the discussion on the issue. My feeling regarding period correctness is that it should be handled by Self::Time as it could map to types like PLLs which could have a design shared with other timers on the processor. That's just a gut feeling though I haven't done any analysis on processors other than the ones I've experienced personally.

src/watchdog.rs
```diff
@@ -0,0 +1,31 @@
...
+    type Time;
+    /// Starts the watchdog with a given period, typically once this is done 
+    /// the watchdog needs to be kicked periodically or the processor is reset. 
+    fn start(&mut self, period: T) where T: Into<Self::Time>;
+}
+
+
+/// Disables a running watchdog timer so the processor won't be reset.
+#[cfg(feature = "unproven")]
+pub trait WatchdogEnable {
```
burrbull: maybe WatchdogDisable?

xd009642:
I haven't heard anymore on this, so is there anything that has to be done to progress it?

therealprof:
@xd009642 Do you have an example implementation somewhere?

xd009642:
No but I'll add one to my own embedded-hal implementation crate and link to it when I have

xd009642:
This is a pretty quick implementation and an example showing it used. Ideally I'd have timers etc implemented so I could more scientifically trigger resets etc but I feel this is alright as a proof of concept.

Hal implementation: 
https://github.com/xd009642/stm32f469xx-hal/blob/watchdog/src/watchdog.rs 

Example usage:
https://github.com/xd009642/stm32f469i-disco/blob/watchdog-test/examples/watchdog.rs

src/watchdog.rs
```diff
@@ -0,0 +1,31 @@
+//! Traits for interactions with a processors watchdog timer.
+
+
+
+/// Kicks an existing watchdog to ensure the processor isn't reset.
+#[cfg(feature = "unproven")]
+pub trait Watchdog {
+    /// Triggers the watchdog. This must be done once the watchdog is started
+    /// to prevent the processor being reset.
+    fn kick(&mut self);
```
hannobraun: This is just a nitpick, and maybe I'm squeamish, but I don't like the thought of kicking a dog :)

In the LPC82x documentation, this operation is called "feed"; nRF52 calls this "reload". I think "reload" is a better and more descriptive name.

hannobraun:
I think this looks reasonable, although I would prefer if `kick` was changed to `reload`.

@xd009642 Could you squash the commits? I'd prefer not to merge commits with build errors into master.

@rust-embedded/hal I'd like to get this merged soon. Does anyone have objections?

therealprof:
@hannobraun FWIW I only know the phrase "kicking the watchdog" (for over 20 years even). I know the term loading only with setting the watchdog timer so naturally I would associate "reloading" with changing the timer value (independent of whether watchdogs actually allow you to do that or not).

I'm not strongly attached to whether we kick the dog or not but deviating from commonly used terminology might create confusion.

hannobraun:
@therealprof Good to know. I've heared about kicking the watchdog for the first time in this PR, but I don't have nearly as much experience as you :)

I've also checked the STM32F0x1/STM32F0x2/STM32F0x8 reference manual, and there's no mention of kicking either (the concept doesn't seem to be called anything except "writing to the key register"). And for what it's worth, [feeding](https://www.google.com/search?q=%22feed%22+%22watchdog%22+%22timer%22) has more results than [kicking](https://www.google.com/search?q=%22kick%22+%22watchdog%22+%22timer%22) on Google (273.000 vs 152.000). Do you think `feed` would be an acceptable name?

I do find the kicking kind of distasteful, but I don't feel that strongly about it. If you think that `kick` is the right name, I won't object.

xd009642:
At my workplace we commonly use kick, even in documents sent out to customers and subcontractors. I'd never considered an alternate term it's only ever been kick for me.

That said I checked an STM register manual for the STMF469 and it used the term refresh (just to throw another word into the mix). 

I personally prefer kick as a term but I'm fine with whatever is understood by people. So I'll change it to feed and squash the commits and if there are any strong objections I'll just switch it to whatever the consensus is!

xd009642:
So I did the squash and push and it says I dismissed someones review and requested reviews. I assume this is github doing things automatically and that means it's ready to progress some more :smile:

hannobraun:
Thank you, @xd009642! Regarding the name, as I said, I don't feel that strongly, and among the three of us, I'm guessing I have the least experience in the embedded space.

I just noticed that there's something wrong with Travis, but on first glance, it looks like it's not related to this PR. I'll look into it, but not today.

therealprof:
@hannobraun https://en.wikipedia.org/wiki/Watchdog_timer also mentions "kick" but anyway, LGTM.

hannobraun:
I'll look into it when I get a chance. I'm a bit busy right now though, so it might take a while. If anyone wants to take a look in the meantime, feel free to do so.

hannobraun:
So, the problem here is that `after_success.sh` can push the documentation to the gh-pages branch, but needs an access token to do so. It tries to access a variable called `GH_TOKEN`, but that doesn't seem to exist. There's [configuration](https://github.com/rust-embedded/embedded-hal/settings/hooks/14277254) (link only works if you have admin access to this repository) that sets this up, but I suspect something might have been lost in the move to the rust-embedded organization.

I'm not sure what to do:
- I don't know what's wrong with the configuration I linked, and the page says that GitHub services are being deprecated. I suspect we should install Travis as a GitHub app, but I don't think I can do that for the rust-embedded organization.
- I could add the `GH_TOKEN` variable in the [Travis settings for embedded-hal](https://travis-ci.org/rust-embedded/embedded-hal/settings) (again, link only works if you have access), but I don't think I can generate an access token for just this repository. I don't want to add a personal access token there, that anyone with access to that configuration could use.

@japaric If you know what's going, I'd appreciate a hint. If you're unsure yourself, don't worry, I'll take another look when I get the chance.

astro:
You could use moving `self` to assert you feed a watchdog that must have been started:

```rust
/// Feeds an existing watchdog to ensure the processor isn't reset. Sometimes
/// commonly referred to as "kicking" or "refreshing".
#[cfg(feature = "unproven")]
pub trait Watchdog {
    /// Triggers the watchdog. This must be done once the watchdog is started
    /// to prevent the processor being reset.
    fn feed(&mut self);
}

/// Enables A watchdog timer to reset the processor if software is frozen or 
/// stalled.
pub trait WatchdogEnable {
    /// Unit of time used by the watchdog
    type Time;
    /// The started watchdog that should be `feed()`
    type Target: Watchdog;
    /// Starts the watchdog with a given period, typically once this is done 
    /// the watchdog needs to be kicked periodically or the processor is reset. 
    ///
    /// This consumes the value and returns the `Watchdog` trait that you must
    /// `feed()`.
    fn start<T>(self, period: T) -> Self::Target where T: Into<Self::Time>;
}

/// Disables a running watchdog timer so the processor won't be reset.
#[cfg(feature = "unproven")]
pub trait WatchdogDisable {
    type Target: WatchdogEnable;
    /// Disables the watchdog
    ///
    /// This stops the watchdog and returns the `WatchdogEnable` trait so that
    /// it can be started again.
    fn disable(self) -> Self::Target;
}
```

hannobraun:
bors retry

hannobraun:
@astro I kinda like your suggestions, but I think we shouldn't hold up this pull request right now. It has been open for a long time and has already been approved.

Would you mind opening a separate issue, so we can discuss your proposal separately?

astro:
@hannobraun not sure if worth the effort if you don't think it's important.

hannobraun:
@astro I do think it's important to further improve our traits, and I certainly think it's worth the effort to open an issue. I've gone ahead and done so: https://github.com/rust-embedded/embedded-hal/issues/98

But I also think we need to make forward progress. This proposal was in the pipeline for months, has gone through various reviews and iterations, and reached consensus. I think it was time to declare victory.

I'd love to see further improvements, but getting those merged will require another round of reviews, discussions, and more deep thinking. The right place to do this is a new pull request, in my opinion.

PR #77, opened 2018-04-19
[WIP] CAN Interface take 2

kjetilkjeka:
#53 has been quiet for a while, the natural step was to try again based on the feedback given in that discussion. I've not implemented the Traits yet, as i wanted to get some quick feedback before i bothered.

@japaric @therealprof is this something close to what you're comfortable merging?

@dunmatt @tibb888 t can this work for you?

The thing i dislike the most with the current state of this interface is that there is no function to return both CanFd and Can frames. This might be a source of a few different bugs.


therealprof:
@kjetilkjeka I'd need to check that in detail but on a first glance it looks good to me. I'm not sure I do understand the purpose (and quite different signature) of `transmit_without_priority()`, though.

kjetilkjeka:
> I'm not sure I do understand the purpose (and quite different signature) of `transmit_without_priority()`, though

Can is designed with RT systems in mind. This means that there is some priority of frames involved. This means that implementing a Can driver is not straight forward, the naive implementation will most likely include a potential unbounded priority inversion. 

The idea is that `transmit` and `receive` will avoid this issue, while for systems where this isn't an issue (and you don't want the complexity of swapping frames) there also exists alternative functions. I see now that `transmit_without_priority()` is a terrible name, so this will need to be changed either way (there should perhaps also be a receive equivalent?). Some alternative ways to handle this.
 - Rename `transmit_without_priority()` to something sensible.
 - Remove `transmit_without_priority()` and force everyone to use the priority inversion free alternative.
 - Include a flag in the transmit function instead. I don't like this as it should be harder to use the error prone alternative (longer function name).

therealprof:
I only see three minor issues with those methods:
- The name is weird
- The description doesn't properly explain when and how to use it
- By putting it in the trait you're forcing every implementor to provide implementations for those methods and I have no idea how often they would be used and whether they usually can be sensibly implemented

If they're "special" I'd prefer having them separately in an extra- or super-trait.

therealprof:
@kjetilkjeka Looks good to me.

I'd like to hear from a user (e.g.  @dunmatt @tib888) whether that could work for them.

kjetilkjeka:
I fixed my own concern by letting `FdInterface` return all types of frame.

Instead of having a second weird transmit function i included the `transmit_buffer_full` that can be checked before attempting to transmit.

I'll leave it for a day or two to let other users express concerns, after that I will make a test implementation.

src/can.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// A Can (11-bit) ID
+#[cfg(feature = "unproven")]
+pub trait BaseId: Sized {
+    /// A generic ID type that encapsulate this type
+    type Id: Id + From<Self> + TryInto<Self>;
+}
+
+/// A Can Extended (28-bit) ID
```
dunmatt: Minor nit:  Should say "29-bit"

src/can.rs
```diff
@@ -0,0 +1,111 @@
...
+pub trait ExtendedId: Sized {
+    /// A generic ID type that encapsulate this type
+    type Id: Id + From<Self> + TryInto<Self>;
+}
+
+/// A Can Frame
+#[cfg(feature = "unproven")]
+pub trait Frame {
+    /// The Id type of this Frame
+    type Id: Id;
```
dunmatt: I'm dubious of the value of having Frame type and the Id type be associated.  It's very common for both Id types to be present on the bus at the same time.

src/can.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Returns `None` if remote frame.
+    fn data(&self) -> Option<&[u8]>;
+}
+
+
+/// A CAN interface
+#[cfg(feature = "unproven")]
+pub trait Interface {
+    /// The Can Frame this Interface operates on
+    type Frame: Frame;
```
dunmatt: Doesn't this associated type mean that an interface only works with one frame type?  If so, and the frame type is associated with the Id type, you'd need two different interfaces for the same hardware in the event of a bus that contains both id types.  That doesn't strike me as a good design.

src/can.rs
```diff
@@ -0,0 +1,111 @@
...
+#[cfg(feature = "unproven")]
+pub trait Frame {
+    /// The Id type of this Frame
+    type Id: Id;
+
+    /// Returns the Can-ID
+    fn id(&self) -> Self::Id;
+
+    /// Returns `Some(Data)` if data frame.
+    /// Returns `None` if remote frame.
```
dunmatt: This seems like a non-obvious way to encode remote frames.  If I asked a random programmer "What would you guess is the difference between a CAN frame with None for the data, and a CAN frame with Some(empty_slice) as the data?"  I'm skeptical they'd guess "Whether or not its a remote frame!"

Or to voice my concern from a different angle, Option means "may not exist", where as RTR means "I would like you to do a thing", and those aren't well connected.

src/can.rs
kjetilkjeka: This is not a problem as Id is defined as "A type that can either be `BaseId` or `ExtendedId`"

src/can.rs
kjetilkjeka: I agree. How can we change this to the better?

Returning a len 0 slice for a remote frame is worse i believe. At the same time we will need to support frame encodings where it's not statically known if they are remote or data frames (from the discussion in the last PR)

Would it be sufficient to introduce `is_remote_frame(&self) -> bool` and `is_data_frame(&self) -> bool` as well?

dunmatt: Why is a len 0 slice worse?  On the wire the length field is indeed zero and zero bytes come before the checksum.  It seems like a great modeling of what is actually happening on the bus.

I'm not sure what you mean wrt not knowing if a frame is remote or data, it's a required bit, it will always be present and always be either 0 or 1...?  What's a situation where you might not know?

But yes, `is_remote_frame(&self) -> bool` seems like a fine solution.

src/can.rs
dunmatt: In Frame2, what would be the type of the id?

src/can.rs
kjetilkjeka: Just one Frame type and one ID type. Some examples of how the Frame type could look (only one of them, is actually required):
``` rust
pub enum Frame1 {
    BaseIdDataFrame(_),
    BaseIdRemoteFrame(_),
    ExtendedIdDataFrame(_),
    ExtendedIdRemoteFrame(_),
}

pub struct Frame2 {
    remote: bool,
    extended_id: bool,
    data: [u8; 8],
}
```

and the Id type could look like (only one of them, is actually required):

``` rust
extern crate ux;
use ux::{u11, u29};

pub enum Id1 {
    Base(u11),
    Extended(u29),
}

pub struct Id2 {
    extended: bool,
    id: u32,
}
```

When implementing the trait each representation of a `Frame` would have to choose a representation of a n`Id`.

dunmatt: So you would also have 

    impl BaseId for Id2 {...}
    impl ExtendedId for Id2 {...}

?

src/can.rs
kjetilkjeka: Probably something close to:
``` rust
pub struct Id2 {
    extended: bool,
    id: u32,
}
```

src/can.rs
dunmatt: I think you'll find it challenging to come up with an API that is natural for both static and dynamic typing.  Static typing would have it that the id types are different in some way, and dynamic would have it that everything from either id type is on both.

But I hear what you're saying about the conceptual difference.  And you made a good point in the other PR, namely, that the compiler should enforce that RTR frames have no data.  I would say, though, that this conceptual difference is a bit idealistic.  Obviously it's best if invalid configurations can't be represented, but what if invalid configurations are what other devices on the bus are sending?  Should the HAL prevent communication with such devices (that are, generally, not within the user's control, eg the ECUs in a car)?

kjetilkjeka: Due to the nature of this crate is essential that it works for all reasonable representations of the can frame. 

>  I would say, though, that this conceptual difference is a bit idealistic.

Sure, but I would also argue that it's an idiomatic use of Rusts typesystem.

>  Obviously it's best if invalid configurations can't be represented, but what if invalid configurations are what other devices on the bus are sending? Should the HAL prevent communication with such devices (that are, generally, not within the user's control, eg the ECUs in a car)?

The DLC field in a remote frame doesn't mean "how long is the data field for this message" it means "How much data do i want to receive". Therefore the CAN controller would never be able to assemble such frame and this would never be an issue. But even if it were, I would almost always go for the fail fast and loudly approach and not accept a malformed Frame.

dunmatt:
It's essential that it works for all reasonable uses of CAN, it's not clear
that it needs to be representation agnostic, another way to support all
reasonable uses of the hardware is to specify a representation that will
always work.  This is, for example, how the ChibiOS HAL works.

I get that that's how dlc works, but since this is a HAL we can't assume
the hardware will filter bad frames.  I agree with you that failing fast
and loud is usually the right call in these cases, but I'd argue that
failing in that way is the job of the driver rather than the HAL, that the
HAL should not contain any policy decisions.  There are legitimate uses for
drivers that don't fail in that case (eg, can pen testing, and interacting
with poorly implemented devices (both of which I do at work))

dunmatt:
>  If this is something you rely on you should probably use some functions from the driver instead from the hal.

That seems to suggest that you believe that the drivers are providers of the HAL rather than consumers of the HAL.  I assert it's the other way around, that the reason to have a HAL in the first place is to be consumed by drivers.  My whole objection is that drivers such as I need cannot be written to consume the HAL you're proposing (which is fine, you need not cater to me, I'm an unusual case and I'm perfectly happy scratching this itch for myself).

kjetilkjeka:
The device drivers are consumers of the HAL while HW drivers are "producers". The HW drivers will also implement things that are not included in the HAL, and is possible to call directly into from the device drivers.

Is it the assumption of no data in remote frames that doesn't work for you? Perhaps you can give a concrete example of when this happens so it's easier for me to understand the need?

dunmatt:
Ah, I see what you're saying, what you're calling HW drivers I had been calling HAL implementations, but you're right that they need not be the same.  That said, the functionality I'm talking about is send and receive, if I have to bypass the HAL in order to send and receive there is really no point depending on the HAL.

Sure, my exact use case is this:  the control and security of autonomous cars.  There are computers from a half dozen or so different vendors on the bus, of which some we have executive control, some we have a little control, and the remainder are black boxes (literally, in one case).  Moreover for business reasons we don't want to marry ourselves to our current vehicle platform, so there's a level of uncertainty even in terms of the devices on the network (ie I can't assume that all devices I'll ever care about are standards compliant, but I can assume that if one isn't I won't be able to fix it).

So given that I'm trying to control and secure black boxes, how would I do it?  Setting aside control, in security applications you always assume malicious input, so my software needs to be able to detect (receive) malicious input and respond to it.

The assumption that only valid data happens on the CAN bus is one problem I have.  Another is that my design intuitions skew more towards KISS than "make it impossible to fuck up" (but you're right that the latter is more idiomatic).  Another is that enforcing policies isn't the job of an abstraction layer.

kjetilkjeka:
> Sure, my exact use case is this: the control and security of autonomous cars. There are computers from a half dozen or so different vendors on the bus, of which some we have executive control, some we have a little control, and the remainder are black boxes (literally, in one case). Moreover for business reasons we don't want to marry ourselves to our current vehicle platform, so there's a level of uncertainty even in terms of the devices on the network (ie I can't assume that all devices I'll ever care about are standards compliant, but I can assume that if one isn't I won't be able to fix it).

In such network (with a lot of different can controllers) wouldn't at least some of them be standards compliant and inject Error frames? Wouldn't the node sending malformed messages go into bus-off state pretty quick? Have you encountered any concrete devices that behaves this way?

> So given that I'm trying to control and secure black boxes, how would I do it? Setting aside control, in security applications you always assume malicious input, so my software needs to be able to detect (receive) malicious input and respond to it.

Putting a black box node on a cars can network sounds to me like installing software you don't know what do on your servers.

With CAN my solution would be:
 1. Only allow standard compliant devices on the bus. CAN is quite robust in itself.
 2. Keep malicious nodes physically away from the bus. Require strict specifications of every node.
 3. Be vary of anything that can proxy stuff onto your can bus.
 4. Use a redundant interface for mission critical nodes, and only keep mission critical nodes on the second CAN-bus. This way if a malicious node messes up your primary can bus you can still provide some fail safes.

dunmatt:
> Wouldn't the node sending malformed messages go into bus-off state pretty
quick? Have you encountered any concrete devices that behaves this way?

Not if they're malicious they wouldn't, and I'm building one, respectively.

> Putting a black box node on a cars can network sounds to me like
installing software you don't know what do on your servers.

Quite right, it's a wonder anyone deploys Window or Linux servers without
first auditing the whole thing for themselves.  By which I mean, the black
boxes come with the cars when we buy them, and for the sake of getting
stuff done we do have to trust someone sometime... Trust but verify.
Believe me, if we could get the auto makers to give us their source we'd do
it.

therealprof:
@dunmatt Indeed the assumption for the HAL is to make it easy to write well-behaved applications across different hardware. If that means that special applications like pen testing are not possible then it is what it is.

My question to you would be why you want to use an HAL anyway? For such uses you'd typically want to use both the most capable hardware with the lowest level access possible. A HAL will never give you that kind of access... Even the existence of a HAL implementation doesn't necessarily mean that you have to use it, you can still directly access the hardware instead.

dunmatt:
Actually I might have missed your point, you were saying that a node I
might want to interact with should go to bus off.  I agree that it should,
but I don't have enough faith in the quality of automotive software in
general to assume that even the bus off part of the spec will be correctly
implemented.  Even if I had proof that our current cars are standards
compliant I'd be wary of committing to being incompatible with non
standards compliant devices

dunmatt:
@Daniel, I completely agree, the original question was if this would work
for my use case, and the answer is: no and that's fine.  I really don't
mind scratching my own itch.

As for why I'd want a hal, it's so that if later I find the can peripheral
too limiting I can swap in a bit banged implementation without rewriting
everything

kjetilkjeka:
I guess that means we're in agreement. Thanks for giving input even though it wont be all usable in your case @dunmatt . I'll give @tib888 a day or two to addresd concerns before I implement it

tib888:
Unfortunately nowadays I do not have much time to deal with this, but please find here a few things popped in my mind reading the proposed code:

I see that filtering is not addressed here (yet?), but from my experience the RTR should be part of the ID, since frame filtering involves the RTR bit also. I may be wrong, but please consider this.

I think this below will not work if you really want a non blocking implementation for more than one transmit mailbox, because it can not trace the state/progress of the transfer:
`fn transmit(&mut self, frame: &Self::Frame) -> nb::Result<Option<Self::Frame>, Self::Error>;`

You may have a function, which puts a frame in a transmit mailbox and returns a handle to the chosen mailbox, like this:
`fn start_transmit(&mut self, id: &Id, data: &[u8]) -> Result<TxMailBoxHandle, Self::Error>;`

Then you will have to pull that handle to competition with an other function.
`fn wait_transmitting_of(&mut self, handle: TxMailBoxHandle) -> nb::Result<!, Self::Error>`
Or each TxMailBox could have its own struct instance with a trait.

Furthermore this start_transmit(...) could have separate id and data parameters (as I written above) because you may reuse the same ID may times and change the data only. Wrapping these two always into a frame is just boilerplate... But again, maybe I'm wrong.

Also, the Interface trait should be splitted into RxInterface and TxInterface because those are almost independent. One part of the application may need only the ownership of the receiver and an other only the transmitter...

stm32f103 for example have two independent receive fifos (you can configure the filters where to put the received frames) and three independent mailboxes (but there is central a register which points to the empty or lowest priority tx mailbox).

I have a working, but still work in progress CAN implementation (not implements this HAL yet) here:
https://github.com/tib888/stm32f103xx-hal/blob/master/src/can.rs

kjetilkjeka:
> I see that filtering is not addressed here (yet?), but from my experience the RTR should be part of the ID, since frame filtering involves the RTR bit also. I may be wrong, but please consider this.

Yes filtering is core part of the can specification and will need to be added.

From the CAN specification (ISO-11898-1:2003, don't have access to the 2015 review, please mail me if you have):
> A  frame  transaction  initiated  at  the  LLC  sublayer  shall  be  a  single,  self-contained  operation  independent  of  previous  frame  transactions.  The  content  of  a  frame  shall  be  named  by  its  identifier.  The  identifier  does  not  indicate  the  destination  of  the  frame  but  describes  the  meaning  of  the  data.  Each  receiver  may  decide  by  frame acceptance filtering whether the frame is relevant or not. 

And the definition of identifier is:
> The  identifier  field  shall  be  composed  of  three  segments:  base  identifier,  extension  flag  and  identifier  extension. The length of the base identifier shall be eleven (11) bits (ID-28 to ID-18), the extension flag one bit, and  the  length  of  the  identifier  extension  shall  be  eighteen  (18)  bits  (ID-17  to  ID-0).  The  identifier  extension  shall be ignored if the extension flag is logic zero (0). 

I interpret this as a can controller is not required to filter on RTR to be compliant to specifications.

> I think this below will not work if you really want a non blocking implementation for more than one transmit mailbox, because it can not trace the state/progress of the transfer:
`fn transmit(&mut self, frame: &Self::Frame) -> nb::Result<Option<Self::Frame>, Self::Error>;`

I think you're correct, I will rethink this.

kjetilkjeka:
Rebased onto japaric/master.

I made a split between Receiver and Transmitter after @tib888 recommendation. This only makes the interface more flexible and i don't think there are many disadvantages to doing so.

The problem of following the progress of the transfer is a hard one. I don't think it's a good idea to hand out Mailbox handles for two reasons:
 1. The can controller might implement some sort of priority queue instead of mailboxes.
 2. It's not simple to invalidate/represent a handle to a Mailbox when the mailbox starts transferring a new Frame. Unless we force an inflexible API requiring to track mailboxes manually it will also be racy.

I instead provided a method to see if there are any pending transmissions matching an ID. I think this will be sufficient in most (all) cases. Perhaps anyone can think of a case where it is not? (@tib888 )

gbip:
Any news on this ? I am about to jump into creating CAN abstraction for the stm32f446. Is there any work that I can reuse ?

kjetilkjeka:
Not directly for stm32f446. I've implemented this interface for socketcan and my next steps when i have the chance to work on this will be to implement it for S32K144 or AtsamE70 as well (both can and can-fd). 

I've not planned to work with rust on any stm32 architectures. If you would like to collaborate shaping this interface it would be great. In that case you should find out if this is something that make sense to implement on stm32f446. If there are friction points we can discuss them and change the interface correspondingly.

johnthagen:
@kjetilkjeka FWIW, I've been looking into the STM32 C HAL libraries to better understand how they do CAN until the Rust side is ready. Thought I'd share in case it is useful for this effort.

[STM32CubeF3](https://www.st.com/en/embedded-software/stm32cubef3.html) has a CAN networking example written in C using their HAL layer for the [STM32303E-EVAL](https://www.st.com/en/evaluation-tools/stm32303e-eval.html) board, which uses the same microcontroller as the F3DISCOVERY @japaric has done a lot of work with.

`Projects/STM32303C_EVAL/Examples/CAN/CAN_Networking`

BearzRobotics:
@kjetilkjeka Is there any progress on this?

kjetilkjeka:
I've been busy with a lot of other things so I'm afraid not. I expect to be able to use some time on this within the next couple of months, as I'm going to need a CAN interface for one of this microcontrollers myself by this time. 

A couple of things that has made this not straight forward to implement is the following:
 - `svd2rust` is not able to generate code when there are registers without a reset value. This is allowed from the svd specification. This is the case for one of my example platforms.
 - Code reuse within a chip family using SVD files are not straight forward, making implementation of these drivers harder than it needs to be.

I've been investigating these two things recently, and will hopefully be able to suggest some imporvements soon.

jdemilledt:
I've recently taken up an interest in this. What issues need to be resolved to get this moving forward? I am willing to commit time to this.

kjetilkjeka:
> I've recently taken up an interest in this. What issues need to be resolved to get this moving forward? I am willing to commit time to this. 

Great!

First of all you should look at the traits I've proposed and see if they make sense to you or whether you think anything should be done in a different way. 

If you think my current proposal make sense, you should attempt 
 - Implement it for one or more hardware to locate friction points from the implementer side of things.
 - Use it for one or more applications to find any friction point from the application side of things.

src/can.rs
jdemilledt:
```rust
enum ((I don't know what this should be named)) {
    Remote,
    Data(&[u8])
}
```
This would be an idiomatic solution that clearly represents what this frame is.

jdemilledt:
One thing I'll point out here that, IMO, seems a bit inefficient: Since `ux::{u11,u29}` aren't primitives, the Rust compiler and runtime treat them differently. Not to mention, the CAN registers on at least the STM32 are all 32-bit. To decrease complexity and code size, IDs should just be u32 and either truncate or error when they run out of limits.

kjetilkjeka:
First I will mention some things I think the API should attempt to avoid:
 - When using the `u32` form the uppermost bit flags may be interpreted differently on different hardware. As a minimum, it should be specified in code what kind of bits are used for what kind of flags.
 - When handing invalid data (out of range IDs etc) different platform may choose to behave differently. Some may saturate, some mask out bits and some may do nothing. This should be avoided.

---

> One thing I'll point out here that, IMO, seems a bit inefficient: Since ux::{u11,u29} aren't primitives, the Rust compiler and runtime treat them differently. 

The current traits are not designed specifically to work with `ux::{u11,u29}`, but rather to be flexible enough to work with "anything".

> Not to mention, the CAN registers on at least the STM32 are all 32-bit. To decrease complexity and code size, IDs should just be u32 and either truncate or error when they run out of limits.

I agree that using a concrete type will be more ergonomic than the Id trait from this PR. Although it's idiomatic to this HAL to use minimal traits for a lot of things, also using this approach for ID is perhaps to overdo it and we should definitively discuss the advantages of making this a concrete type instead.

I'm not certain to what degree the bit format of the u32 is standardized over different hardware. If this is different, using a fixed layout u32 would only allow optimizing for one platform. This would, in turn, make the static dispatch trait approach more efficient.

If we want to go with a type instead of a trait, we should at least use a newtype of `u32` instead just an `u32`. This will not cause any bloat in code size. Something like the following can be an alternative

```rust
#[derive(Debug, PartialEq, Clone, Copy, Hash /* etc */)]
pub struct CanId(u32);

impl From<CanId> for u32 { /* ... */ }

impl CanId {
    const ID_MASK: u32 = 0x7ff;
    const EXTENDED_MASK: u32 = 0x1fffffff;
    const EFF_MASK: u32 = (1<<31);

    pub fn try_from_id(u16) -> Result<Self, _> { /* ... */ }
    pub fn try_from_extended(u32) -> Result<Self, _> { /* ... */ }
    pub fn try_from_bits(u32) -> Result<Self, _> { /* ... */ }

    pub fn is_extended(self) -> bool { /* ... */ }
    pub fn id(self) -> u32 { /* ... */ }
}
```

The problem with this is of course that `id.into()` cannot be used directly for hardware that doesn't use the bit map specified by this type. As previously mentioned, I'm not sure to what degree the bit format is standardized on, but this is potentially a probelm.

A different alternative would be to use a bit more friendly and simpler trait. This will allow for using optimized bit layouts for all hardware, while perhaps being a bit less cumbersome to work with than the trait from the PR. Let me suggest something like the following:

```rust
pub trait CanId : Into<u32> {
    /// The mask of the ID when a standard frame is converted into an `u32`
    const ID_MASK: u32;

    /// The mask of the ID when an extended frame is converted into an `u32`
    const EXTENDED_MASK: u32;

    /// The mask of the extended frame flag when converted into an `u32`
    const EFF_MASK: u32;

    pub fn try_from_bits(u32) -> Result<Self, _>;

    pub fn try_from_id(u16) -> Result<Self, _> {
        /* Default implementation based on `try_from_bits` and consts */
    }

    pub fn try_from_extended(u32) -> Result<Self, _> {
        /* Default implementation based on `try_from_bits` and consts */
    }

    pub fn is_extended(self) -> {
        /* Default implementation based on `Into<u32>` and consts */
    }

    pub fn id(self) -> u32 {
        /* Default implementation based on `Into<u32>` and consts */
    }
}
```

The approach for creating a new platform would be to create a `CanId` newtype over `u32` and trivially implement the trait in a way that matches with the hardware bit layout.

---

To gain more insight, we should look whether bit patterns of registers used in CAN hardware is standardized. I suggest we at least compare the following relevant platforms:
 - socketcan
 - Microchip MCP2515 and MCP2517
 - The STM implementation you were talking about
 - Bosch IP (MCAN) - Used by atmel/microchip atsam
 - NXP S32K implementation

jdemilledt:
A possible solution if the CAN ID types aren't compatible could be to have HAL implementors specify CanId <-> u32 conversion methods.

jdemilledt:
For the record, at least the ST implementation makes a signal that can be sent to an MCP2515 which translates it to the CAN differential signal.

jdemilledt:
Oops, I misunderstood what the MCP2515/17 are. Ignore my comments about ST and the MCP.

jdemilledt:
I've taken a look at the SAM and STM32 implementations, and so far, I'm already seeing differences in ID encoding, but they're along the lines of a mask and a bit shift from what I'm seeing. Nothing too fancy. Perhaps just add a bit shift const to the CanId trait?

jdemilledt:
@kjetilkjeka I think an idiomatic way to solve the `u32` ID issue would be to have `unsafe` methods that just shift out any extra bits, and safe methods that error out if an ID is too big. If there's enough interest, I could also implement a lint that warns users if they hardcode IDs that are too big.

kjetilkjeka:
@jdemilledt I don't think your changes related to ID will work. When converting from `u32` it's not valid to convert all IDs that will fit in the normal ID range as a BaseID. Using extended IDs with low numbers is a totally valid thing to do with CAN, and is done all the time in practice.

By making the ID an enum instead of a trait, we're also missing the opportunity of abstracting over the native bit layout of the CAN controller. This will be slightly less efficient in both memory and CPU, although this might be fine I think the trait I previously proposed might solve this in a good way.

I also don't like the removal of the small default implemented convenience functions. I think they are quite valuable in practice. And they are implemented by default, meaning no overhead for the implementer.

Returning the `FrameContents` is fine, but require a lot of convenience functions implementation to be as convenient as the `Option`.

jdemilledt:
Those are some good points. I don't necessarily agree about making a trait instead of a fully implemented struct since CAN IDs are the same across platforms except the underlying implementation's bit requirements, which could be implemented in the `Interface`. Can you please elaborate on what you're saying about convenience functions?

kjetilkjeka:
> Can you please elaborate on what you're saying about convenience functions?

I'm talking about the functions like `is_data_frame(&self)`, we should definitely have those kinds of functions to avoid always having to pattern match on enums. It's quite idiomatic to Rust. Take for instance Result, it implements `is_ok()`, `is_err()` and a ton of other "convenient" things.

jdemilledt:
I see. I removed them because I thought they were redundant, but if you want to keep those, that's fine.

no111u3:
@kjetilkjeka I think you may update your pr to current version, some changes is good for me, some need clarification.

arrowcircle:
Any updates on bringing can bus to embedded-hal?

src/can.rs
```diff
@@ -0,0 +1,190 @@
+//! Controller Area Network
+
+use nb;
+
+/// A type that can either be `BaseId` or `ExtendedId`
+#[cfg(feature = "unproven")]
+pub trait Id {
```
marcelbuesing, line 7: I don't really get the value of not specifying concrete types (u16 and u32) for base and extended id.

How about adding a 

```rust
fn id(&self) -> u32;
```

possibly even with a blanket implementation based on the `base_id` and `extended_id` fns, but I think right now there is no way to guarantee that not both are not `None` so this might need to return `Result<u32, Error>` which is kind of ugly.

PR #80, opened 2018-05-11
replace uses of ! with Void

japaric:
We are soon going to push for embedded *library* development on beta (1.27-beta) to minimize the
number of unstable features used in the ecosystem and ease the transition towards *application*
development on stable (1.28 or 1.29).

Unfortunately `!` didn't make it into the beta release but we still need to make this crate work on
beta because a lot of crates depend on it. So this PR replaces the single use of the `!` type (cf.
`CountDown.wait`) with the equivalent, but stable, `Void` type.

This is a [breaking-change]. Implementations of the `CountDown` trait will have to update the
signature of `wait` to return `nb::Result<(), Void>`.

cc @therealprof @hannobraun I'm going to publish v0.2.0 *soon* (ideally tonight). If there's
some **breaking**, but *uncontroversial*, change that should be included in it let me know ASAP so
we can land it before the release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,8 @@ before_cache:
 
 branches:
   only:
-    - auto
-    - master
-    - try
+    - staging
+    - trying
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,14 @@ language: rust
 matrix:
   include:
     - env: TARGET=x86_64-unknown-linux-gnu
+
+    - env: TARGET=thumbv6m-none-eabi
+      rust: beta
+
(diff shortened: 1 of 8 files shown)


therealprof:
@japaric Do they need to be breaking? I think there's a lot of stuff we should start getting in.

PR #56 looks like a go to me.
PR #72 looks good and is even breaking, however the two nits I raised were not addressed by @astro

bergus:
Meh, this sounds like a lot of unnecessary work for implementers of the hal. How long will it take until `!` becomes stable, and all of this has to be changed back? (With another breaking change, unless you consider https://github.com/reem/rust-void/issues/13)

PR #83, opened 2018-05-16
add ToggleableOutputPin to the prelude

japaric:


japaric:
Should `StatefulOutputPin` also be in the prelude? It seems to me that `is_set_{low,high}` are less common than `toggle`, but those two methods were in the v0.1.x prelude because they were part of `OutputPin` and `OutputPin` was in the v0.1.x prelude.

hannobraun:
> Should `StatefulOutputPin` also be in the prelude? It seems to me that `is_set_{low,high}` are less common than `toggle`, but those two methods were in the v0.1.x prelude because they were part of `OutputPin` and `OutputPin` was in the v0.1.x prelude.

I can't think of a good reason why it shouldn't be in there.

tib8:
StatefulOutputPin is definitely needed! (For example for 1-wire protocol implementations using an open drain output pin.)

PR #96, opened 2018-09-28
Fix Travis build

hannobraun:
Please ignore. I just need this pull request to figure out what's wrong with Travis.

### DIFF
--- a/ci/after_success.sh
+++ b/ci/after_success.sh
@@ -15,6 +15,6 @@ main() {
         echo OK
 }
 
-if [ $TRAVIS_BRANCH = master ]; then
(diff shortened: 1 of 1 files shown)


hannobraun:
This should be good to go.

@rust-embedded/hal Anyone, please review, or if not, just tell Bors to merge it. The least this will do un-break all pull request builds, which means we can finally merge pull requests again. If it breaks the automatic documentation updates, we can deal with that later.

jamesmunns:
(oops, I have no power here :) )

hannobraun:
Thanks everyone!

Issue #95, opened 2018-09-23
`digital` operations do not allow error propagation


eldruin:
Currently operations defined in the `digital` module traits do not allow error propagation.

For example, in an [I2C I/O expander driver](https://github.com/eldruin/pcf857x-rs) I am writing I want to provide access to the individual pins as a bunch of `Px` structs implementing `OutputPin` and `InputPin`. Setting a pin or reading it involves I2C communication, which could fail.
At the moment I just [panic if there was a problem](https://github.com/eldruin/pcf857x-rs/blob/7b74e1da6cef9ddfd72742b1ebc0e35191854bc7/src/pins.rs#L113) for either reading or writing.
The same would happen with the other traits: `StatefulOutputPin` and `ToggleableOutputPin`

Another example from @Caemor:
> The only problem I found with this api since using it is that errors can't get propagated any further and need to be converted to a default value.
> 
> E.g. the linux sys_fs can receive an file error while trying to read the value (https://github.com/rust-embedded/rust-sysfs-gpio/blob/master/src/lib.rs#L337) and this can't be handled by a simple bool.

Discussion started [here](https://github.com/rust-embedded/embedded-hal/issues/41#issuecomment-422394417)

ryankurte:
Seems like a good candidate for some `Try` versions of the digital traits? 

The downside is this would result in two possible digital IO interfaces.
The upside is error handling can be explicit where it needs to, and we can smooth that over with an automatic implementation from infallible to try, and a helper for the other direction.

eldruin:
@ryankurte  Sounds good to me :)

hannobraun:
@ryankurte @eldruin 
> Seems like a good candidate for some `Try` versions of the digital traits?

I may be missing something, but can't we just add an associated error type to the existing traits, and set those to `Void` for infallible implementations? This would be an inconvenience in the short term, but once the never type (`!`) is stabilized, the compiler should be smart enough to figure out that these `Results`s don't need to be handled. I think I'd prefer that to another set of traits.

hannobraun:
@eldruin 
``` rust
pub trait InputPin {
    type Error;
    fn is_high(&self) -> Result<bool, Self::Error>;
}
```

hannobraun:
I just noticed that my previous comment about the stabilization of `!` making a difference here is mostly wrong. It applies to code that knows which `embedded-hal` implementation it is using, but it won't make a difference for drivers, `embedded-hal`'s primary consumer.

[...]

PR #97, opened 2018-09-28
Make digital pin traits fallible

eldruin:
Following the discussion in #95, I have made these changes:
- Make `InputPin` trait methods fallible.
- Add `FallibleOutputPin` trait with fallible versions of the `OutputPin` methods.
- Make `ToggeableOutputPin` trait method fallible.

Now, should `StatefulOutputPin` methods also be fallible? I think so, but I was not able to get it to work:
In the implementation of `ToggeableOutputPin` for the software-driven toggle (`FallibleOutputPin` + `StatefulOutputPin`) there are now two `Error` types and I do not know how to solve this correctly.

You can see precisely what I mean by applying the patch below on top of this branch. If you do so,
the compiler will rightfully complain that the `Error` associated type in bounds of `P` is ambiguous.

What do you think?

```diff
diff --git a/src/digital.rs b/src/digital.rs
index 45a72bc..7d6f4f5 100644
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -39,15 +39,18 @@ pub trait FallibleOutputPin {
 /// *This trait is available if embedded-hal is built with the `"unproven"` feature.*
 #[cfg(feature = "unproven")]
 pub trait StatefulOutputPin {
+    /// Error type
+    type Error;
+
     /// Is the pin in drive high mode?
     ///
     /// *NOTE* this does *not* read the electrical state of the pin
-    fn is_set_high(&self) -> bool;
+    fn is_set_high(&self) -> Result<bool, Self::Error>;
 
     /// Is the pin in drive low mode?
     ///
     /// *NOTE* this does *not* read the electrical state of the pin
-    fn is_set_low(&self) -> bool;
+    fn is_set_low(&self) -> Result<bool, Self::Error>;
 }
 
 /// Output pin that can be toggled
@@ -93,11 +96,13 @@ pub trait ToggleableOutputPin {
 /// }
 ///
 /// impl StatefulOutputPin for MyPin {
-///    fn is_set_low(&self) -> bool {
-///        !self.state
+///    type Error = void::Void;
+///
+///    fn is_set_low(&self) -> Result<bool, Self::Error> {
+///        Ok(!self.state)
 ///    }
-///    fn is_set_high(&self) -> bool {
-///        self.state
+///    fn is_set_high(&self) -> Result<bool, Self::Error> {
+///        Ok(self.state)
 ///    }
 /// }
 ///
@@ -106,9 +111,9 @@ pub trait ToggleableOutputPin {
 ///
 /// let mut pin = MyPin { state: false };
 /// pin.toggle().unwrap();
-/// assert!(pin.is_set_high());
+/// assert!(pin.is_set_high().unwrap());
 /// pin.toggle().unwrap();
-/// assert!(pin.is_set_low());
+/// assert!(pin.is_set_low().unwrap());
 /// ```
 #[cfg(feature = "unproven")]
 pub mod toggleable {
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,27 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-### Added
-
-- Fallible versions of the digital `OutputPin`, `StatefulOutputPin`,
-  `ToggleableOutputPin` and `InputPin` traits have been added.
-  These are versions where the methods return a `Result` type as setting an output pin
-  and reading an input pin could potentially fail.
-  The fallible versions are available through the `use-fallible-digital-traits` feature.
-  After this release, the fallible versions of the traits will be made the default and
-  the infallible ones will only be available if activating the
-  `regress-infallible-digital-traits` feature.
-  These will be removed in the relase after that one.
-  Both versions of the traits are mutually-exclusive.
-  See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
-
 ### Changed
 
-- The current infallible versions of the digital `OutputPin`, `StatefulOutputPin`,
-  `ToggleableOutputPin` and `InputPin` traits are now marked as deprecated.
-  In the next release, the infallible versions will only be available if activating the `regress-infallible-digital-traits` feature.
-  These will be removed in the relase after that one.
-  Both versions of the traits are mutually-exclusive.
+- [breaking-change] The digital `OutputPin`, `StatefulOutputPin`, `ToggleableOutputPin`
+  and `InputPin` traits have been replaced with fallible versions.
+  The methods now return a `Result` type as setting an output pin
+  and reading an input pin could potentially fail.
   See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
 
 ## [v0.2.1] - 2018-05-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- A `FallibleOutputPin` trait has been added. It is a version of the `OutputPin` trait
+  where the methods return a `Result` type as setting an output pin could potentially fail.
+  The current `OutputPin` method signatures will become deprecated in a future version and
+  will be replaced by the ones defined in `FallibleOutputPin`.
+  After that, the `FallibleOutputPin` trait will be removed.
+  See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
+
 ### Changed
 
 - [breaking-change] The unproven InputPin trait methods are now fallible, as reading an input
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- A `FallibleOutputPin` trait has been added. It is a version of the `OutputPin` trait
+- A fallible version of the `OutputPin` trait has been added. It is a version
   where the methods return a `Result` type as setting an output pin could potentially fail.
-  The current `OutputPin` method signatures will become deprecated in a future version and
-  will be replaced by the ones defined in `FallibleOutputPin`.
-  After that, the `FallibleOutputPin` trait will be removed.
+  The current `OutputPin` trait version is now marked as deprecated. The fallible version
+  is available throught the `use-fallible-digital-traits` feature. After this release,
+  the fallible version of `OutputPin` will be made the default and the infallible one
+  will only be available if activating the `use-infallible-digital-traits` feature.
+  The infallible version of the trait will be removed in the relase after that one.
   See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- [breaking-change] The unproven InputPin trait methods are now fallible, as reading an input
-  could potentially fail.
+- [breaking-change] The unproven `InputPin` trait methods are now fallible, as reading an input
+  could potentially fail. Similarly, the methods in the unproven `ToggleableOutputPin` trait
+  are now also fallible.
   See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
 
 ## [v0.2.1] - 2018-05-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- [breaking-change] The unproven InputPin trait methods are now fallible, as reading an input
+  could potentially fail.
+  See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
+
 ## [v0.2.1] - 2018-05-14
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,22 +9,25 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Fallible versions of the `OutputPin` and `StatefulOutputPin` traits have been added.
+- Fallible versions of the digital `OutputPin`, `StatefulOutputPin`,
+  `ToggleableOutputPin` and `InputPin` traits have been added.
   These are versions where the methods return a `Result` type as setting an output pin
-  could potentially fail.
-  The current trait versions are now marked as deprecated. The fallible version
-  is available through the `use-fallible-digital-traits` feature. After this release,
-  the fallible version of the traits will be made the default and the infallible one
-  will only be available if activating the `use-infallible-digital-traits` feature.
-  The infallible version of the trait will be removed in the relase after that one.
+  and reading an input pin could potentially fail.
+  The fallible versions are available through the `use-fallible-digital-traits` feature.
+  After this release, the fallible versions of the traits will be made the default and
+  the infallible ones will only be available if activating the
+  `regress-infallible-digital-traits` feature.
+  These will be removed in the relase after that one.
   Both versions of the traits are mutually-exclusive.
   See [here](https://github.com/rust-embedded/embedded-hal/issues/95) for more info.
 
 ### Changed
 
-- [breaking-change] The unproven `InputPin` trait methods are now fallible, as reading an input
-  could potentially fail. Similarly, the methods in the unproven `ToggleableOutputPin` trait
-  are now also fallible.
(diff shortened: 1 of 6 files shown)


hannobraun:
@eldruin 
> You can see precisely what I mean by applying the patch below on top of this branch. If you do so,
the compiler will rightfully complain that the Error associated type in bounds of P is ambiguous.
> 
> What do you think?

 I have two ideas. First, one that I'm pretty sure will work, but might be too heavy-handed. Define the following enum:
``` rust
pub enum ToggleError<OP, SOP> {
    OutputPin(OP),
    StatefulOutputPin(SOP),
}
```

Use that as the error type for the default implementation, like this:
``` rust
type Error = ToggleError<<P as FallibleOutputPin>::Error, <P as StatefulOutputPin>::Error>;
```
Then update the implementation of `toggle` like this:
``` rust
fn toggle(&mut self) -> Result<(), Self::Error>{
    if self.is_set_low().map_err(|error| ToggleError::StatefulOutputPin(error))? {
        self.set_high().map_err(|error| ToggleError::FallibleOutputPin(error))?;
    } else {
        // ...
```
As I said, pretty sure that will work, but it might be too heavy-handed.

My second idea is more elegant, but might not work. Simply make `StatefulOutputPin` inherit from `FallibleOutputPin` and don't give it its own error:
``` rust
pub trait StatefulOutputPin : FallibleOutputPin {
    fn is_set_high(&self) -> Result<bool, Self::Error>;

    fn is_set_low(&self) -> Result<bool, Self::Error>;
}
```
Problem solved, because they both have the same error now. The big question is, are there valid use cases where both traits need different errors?

Bonus: A third idea that's not implementable right now. Give `StatefulOutputPin` its own error, but only provide the default implementation, if they both have the same error type. This can one day be achieved by adding the following where clause:
``` rust
where <P as FallibleOutputPin>::Error == <P as StatefulOutputPin>::Error
```
This is called an equality constraint, and it's [not implemented yet](https://github.com/rust-lang/rust/issues/20041).

eldruin:
I also think the `use-fallible-digital-traits` idea is great, so I implemented it already.
For the duplicated `Error` type, inheriting `StatefulOutputPin` from `OutputPin` actually works :)
I think this version is better than the `ToggleError` enum type because:
- As previously stated, `ToggleError` would be pretty heavy-handed.
- From a conceptual point of view, a `StatefulOutputPin` does not make much sense without being able to set the pin high or low, so inheriting from `OutputPin` is ok.
- If cases for separate `Error` types arise, I think we can avoid breakage by breaking the inheritance and adding a second `Error` type if the equality constraints have landed by then.

Thank you both @ryankurte and @hannobraun!

hannobraun:
This is looking good! From my perspective, here's what's left to to get this merged:
1. Fix the Travis failure (missing import, should be simple).
2. Wait for feedback from @ryankurte.
3. Squash this into fewer commits (I'd prefer one commit changing the unproven trait, another to add the fallible versions; but squashing everything into one commit is fine, too).

src/digital.rs
```diff
@@ -116,14 +135,30 @@ pub trait ToggleableOutputPin {
...
 /// pin.toggle().unwrap();
+/// #[cfg(not(feature = "use-fallible-digital-traits"))]
 /// assert!(pin.is_set_high());
+/// #[cfg(feature = "use-fallible-digital-traits")]
+/// assert!(pin.is_set_high().unwrap());
 /// pin.toggle().unwrap();
+/// #[cfg(not(feature = "use-fallible-digital-traits"))]
 /// assert!(pin.is_set_low());
+/// #[cfg(feature = "use-fallible-digital-traits")]
+/// assert!(pin.is_set_low().unwrap());
```
hannobraun: Hmm, I'm wondering, if we should simplify the example. I mean, we want people to use the new fallible versions, so why even show the infallible versions here? Thoughts?

CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- A `FallibleOutputPin` trait has been added. It is a version of the `OutputPin` trait
```
ryankurte: I think this needs to be updated to match the feature based approach, and we probably need to document that this release is opt-in and the next will be opt-out.

src/digital.rs
```diff
@@ -1,6 +1,7 @@
 //! Digital I/O
 
 /// Single digital push-pull output pin
+#[cfg(not(feature = "use-fallible-digital-traits"))]
```
ryankurte: Is it possible to mark all the infallible traits as depreciated with a suggestion to use the fallible ones?

src/digital.rs
ryankurte: I agree that the docs should prioritise the fallible version, though maybe we could keep the infallible documentation until the opt-out release?

eldruin: I moved the fallible versions up. However, the example still looks quite cluttered with `#cfg...` to me. Any ideas?

src/digital.rs
eldruin: done

CHANGELOG.md
eldruin: done

eldruin:
Ok here is a new version. Please review the points raised and let me know if you want further changes.
To summarize, I would say the definitely open points are:
- [ ] Squash into fewer commits (will do at the end)
- [ ] `ToggleableOutputPin` example code
- [ ] Demonstrate on other crate (I can do this with my pcf857x I/O expander driver)

Thoughts?

CHANGELOG.md
```diff
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Fallible versions of the `OutputPin` and `StatefulOutputPin` traits have been added.
+  These are versions where the methods return a `Result` type as setting an output pin
+  could potentially fail.
+  The current trait versions are now marked as deprecated. The fallible version
```
ryankurte: Should we be marking the new traits with `unproven` too? (I'm not sure if there is a precedent, seems ok either way)

src/digital.rs
ryankurte: It is a bit cluttered, maybe grouping each example would help? 
I'm ok with merging this either way.

ryankurte:
Thanks for the excellent work @eldruin! I appreciate your effort in working through these changes with us ^_^

The only niggling worry I have is whether we should be feature gating the unproven traits as well. While there is no guarantee that they don't change, the reality is everyone using it probably builds with the unproven feature on (I know I do) and is going to get hit by the API breakage, somewhat undermining our plan for a smooth transition.

I think my preference would be to take the same feature-flag migration approach with the unproven traits, so all the digital traits are feature-flagged one way or the other and marked unproven again. Any thoughts on this @hannobraun?

src/digital.rs
eldruin: I found no way to group the statements so that I can have fewer `#cfg`s. Do you have an idea?

CHANGELOG.md
eldruin: `StatefulOutputPin` is already marked as `unproven` and its interface will not change if not activating `use-fallible-digital-traits`. The only fallible trait one not behind the `unproven` feature is the fallible `OutputPin`.

eldruin:
Additional info: I took care of changing the code so that it is possible to activate the `unproven` and `use-fallible-digital-traits` features independently of each other.
If activating `unproven` but not `use-fallible-digital-traits`, only `InputPin` and `ToggleableOutputPin` have changed to be fallible. I could also re-add their infallible version for this case.

hannobraun:
@ryankurte 
> The only niggling worry I have is whether we should be feature gating the unproven traits as well. While there is no guarantee that they don't change, the reality is everyone using it probably builds with the unproven feature on (I know I do) and is going to get hit by the API breakage, somewhat undermining our plan for a smooth transition.
>
> I think my preference would be to take the same feature-flag migration approach with the unproven traits, so all the digital traits are feature-flagged one way or the other and marked unproven again. Any thoughts on this @hannobraun?

Well, if you think it should be done, and if @eldruin is willing to do it, then I'm fine with that. But honestly, I think it's unnecessary. I we can't change unproven traits, then what is the point of having the flag?

New thought: Should we (temporarily) test the new feature flags in the Travis build, or is it enough to test them locally before merging?

ryankurte:
>  I think it's unnecessary. I we can't change unproven traits, then what is the point of having the flag?

ryankurte:
Whoops, pushed the wrong button sorry.

>  I think it's unnecessary. I we can't change unproven traits, then what is the point of having the flag?

I agree in theory but am not really sure the unproven flag works the way it was intended, maybe when the bulk of the library has stabilised so it's not so compulsory it'll fit better, maybe it'd be better with more granularity (a flag per trait), but that is probably a discussion for another place.

If we don't do it, we are causing breakage for the people who do use those traits, but maybe they'll just opt in to the new traits and fix everything then anyway. I can see pros and cons of both and am happy merging either, so, @eldruin up to you.

(Regardless of the direction, given the widespread ramifications of changes to this project it'd be neat to have feedback on the release process from users? Maybe we could have a forum post or an issue for release feedback / experiences.)

> Should we (temporarily) test the new feature flags in the Travis build

Ahh, yeah that should probably go into the build matrix, to be changed when we change and eventually remove the flags.

hannobraun:
@ryankurte 
> I agree in theory but am not really sure the unproven flag works the way it was intended, maybe when the bulk of the library has stabilised so it's not so compulsory it'll fit better, maybe it'd be better with more granularity (a flag per trait), but that is probably a discussion for another place.

I agree. I've asked myself a few times what the point of the flag is, given the early state that this library still is in. More granular flags sound like a good idea, and would fit well with what we're doing in this pull request.

> (Regardless of the direction, given the widespread ramifications of changes to this project it'd be neat to have feedback on the release process from users? Maybe we could have a forum post or an issue for release feedback / experiences.)

I agree. We have an embedded category in [u.r-l.o](https://users.rust-lang.org/) now, which seems like a good place for this kind of thing.

> Ahh, yeah that should probably go into the build matrix, to be changed when we change and eventually remove the flags.

Okay, so I guess this updates the todo list for this pull request as follows:
- Do or don't use feature flags for the unproven traits too, according to @eldruin's decision.
- Add new features to Travis build matrix.
- Open issues to discuss the future use of the unproven feature and release procedures (forum post for feedback).

therealprof:
@hannobraun The main problem I see with `unproven` is that despite the availability of proven implementations we're not changing the flags. I think this flag would work a treat if we would actually remove the flags in a timely manner.

hannobraun:
@therealprof 
> The main problem I see with `unproven` is that despite the availability of proven implementations we're not changing the flags. I think this flag would work a treat if we would actually remove the flags in a timely manner.

Yes, that certainly doesn't help.

eldruin:
Ok, just to let you know, I am on vacation and will have a look at this next week :)

eldruin:
Ok I pushed my last local changes but interesting arguments were brought up in #100 so let's settle that first.

eldruin:
Ok, there seems to be some consensus in #100 that this should be implemented as a clean break. I will prepare a new version soon.

eldruin:
Ok I did not want to squash the commits in this branch in case github makes a mess of the discussion in this PR so I opened #105 with everything in one commit.
I will close this in favor of that one but feel free to reopen if appropriate.

Issue #92, opened 2018-07-31
New version?


little-arhat:
Hi,

Number of features have been added since 0.2.1 -- should there be another release? I can't use git dependency if I want to publish hal-impl to crates.io.

Are there any guildelines regarding versioning? Seems like any new feature merged to master should bump patch level, otherwise hal-impl can't really use them.

therealprof:
I agree both with the idea of bumping the version with each change and to have a new release soon.

little-arhat:
What about releasing new version with revised list of unproven features?

ryankurte:
I'd like to push a new version once we've landed #101, #102 and #105 (hopefully shortly), which will be an 0.X release (specifically 0.3.0) due to breaking changes, any comments @rust-embedded/hal ?

ryankurte:
The falible digital traits PR (#105) is breaking

therealprof:
Right, but I was hoping we would release a non-breaking 0.2.2 before pulling that in.

ryankurte:
Fine by me, we need more reviews on all the linked PRs so once 101 and 102 are reviewed and merged we can release an 0.2.3 then land 105 and 0.3.0.

therealprof:
I can see people wanting to stay on 0.2.x until all crates are ready for 0.3 which will take a while and the request to prove traits has been an elephant in the room for quite a while even longer than this ticket which is also is a few months old now.

hannobraun:
I understand that people want proven traits, but I question the point of declaring traits with known issues proven, then breaking them due to those known issues.

Yes, it sucks that those traits have known issues, and that we have to make breaking changes to fix those known issues. I don't see how declaring those traits proven right before making those breaking changes benefits anyone.

I understand that those users who want the traits proven might stay on 0.2.x for a while. I don't see how not having to enable the `unproven` feature during this time benefits them. They'll have to fix their implementations when upgrading to 0.3 in any case. Declaring those traits proven will remove a minimal obstacle to using them (enabling a feature) at the cost at confusing people about the future stability of those traits.

But as I said previously: I'm not making a stand here. It's not what I would do, but if you want to merge #102, go ahead.

therealprof:
> I understand that people want proven traits, but I question the point of declaring traits with known issues proven, then breaking them due to those known issues.

I'd like to see that we're sticking to the rules which were once decided (or change them).

[...]

PR #101, opened 2018-10-06
ADC-related traits

thenewwazoo:
Opening the discussion! One implementation currently exists [here](https://github.com/thenewwazoo/stm32l0x1-context-hal/blob/master/src/adc.rs), and I'll be adding another for the LPC177x/8x HAL soonish.

I feel like these traits encompass nearly all the useful "modes" that an ADC operates in; n.b. that IME "continuous" sampling of a single pin looks just like a "scan" with one pin enabled, so I didn't distinguish between the two. I'm not sure if I _should_, but it's a known t.b.d.


src/analog.rs
```diff
@@ -0,0 +1,50 @@
...
+    ///
+    /// e.g. u8 for ADC channels 1-n, or a (partid, pinid) tuple for something more complex.
+    type ID;
+
+    /// ID of the channel associated with this pin
+    const CHANNEL: Self::ID;
+}
+
+/// ADCs that sample on single channels per request, and do so at the time of the request.
+pub trait SingleMode<Word, Pin: AdcChannel> {
```
austinbes: Minor bikeshedding: `analog::SingleMode` doesn't seem like the a great name here -- and it doesn't really line up with `analog::Continuous`.

Some off-the-cuff ideas:

- `analog::OnDemand`
- `analog::Once`

src/analog.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Start a continuous conversion cycle
+    fn start(&mut self) -> nb::Result<(), Self::Error>;
+
+    /// Stop conversions.
+    ///
+    /// If the conversions cannot be stopped immediately, returns nb::Error::WouldBlock.
+    fn stop(&mut self) -> nb::Result<(), Self::Error>;
+
+    /// Read the specified channel
+    fn read(&mut self, chan: Pin::ID) -> nb::Result<Word, Self::Error>;
```
austinbes: I'm confused as to how this API ought to work -- maybe I'm missing something.

It seems as though the intended flow would be something like:

```Rust
fn get_samples() {
    let pin = Channel::new(); // actually get a channel though
    let adc = Adc::init(); // actually get an ADC handle

    adc.add_channel(pin);
    adc.start();
    for _ in 0..=10 {
        let _ = adc.read(pin); // fails borrow check
    }
    adc.stop();
    let rpin = adc.release_channel().unwrap();
}
```

I'm omitting error checking, etc, and might have made a mistake with this hasty example. The fact remains, however, that unless I'm misinterpreting when and how `add_channel()` and `read()` must interact, this interface cannot work.

austinbes: I'm also confused as to how this would work with multiple continuously-sampled channels.

src/lib.rs
```diff
@@ -689,6 +689,7 @@
 extern crate nb;
 extern crate void;
 
+pub mod analog;
```
hannobraun: (Please note that I haven't worked with a DAC so far, so the following concerns may be unfounded. It would be great if someone with more experience in this area could pitch in.)

I'm concerned about the naming, and how well it will suit our needs once we add DAC traits. While `analog` lines up nicely with `digital`, having a single `analog` module means both ADC and DAC will have to share the same namespace, and you're already reserving some attractive names for ADC.

Digital I/O is usually implemented in the same peripheral, and some drivers need to concern themselves with both input and output on the same pin. For that reason, I think it's a good idea to have a single `digital` module. The question is, is the same true for ADC and DAC?

If the answer is "no", then I suggest renaming the module to `adc`.

src/analog.rs
hannobraun: I like both suggestions. Another one: `OneShot`, which I've seen used in this context.

hannobraun: Not sure how I feel about the name `Pin`. Maybe it `Channel`, or just `C` would be simpler.

And have you considered moving `Pin` to `read_channel`? That would make it more convenient to use the trait (one less argument to specify in parameter lists), but I see two potential problems:
- I'm not sure if that would allow an implementation to specify what `Pin` is in a `where` clause.
- It would be inconsistent with `Continuous`.

So yeah, not sure if this is a good idea, but I'm interested in knowing if you tried/considered it.

src/analog.rs
```diff
@@ -0,0 +1,50 @@
...
+/// ADCs that sample continuously without requiring requests
+pub trait Continuous<Word, Pin: AdcChannel> {
+    /// Error type returned by ADC methods
+    type Error;
+
+    /// Take ownership of an ADC channel/pin
+    fn add_channel(&mut self, pin: Pin) -> nb::Result<(), Self::Error>;
+
+    /// Release the ADC channel/pin
+    fn release_channel(&mut self) -> nb::Result<Pin, Self::Error>;
```
hannobraun: Presumably you can call `add_channel` multiple times. At least the trait won't prevent that. What will `release_channel` return then?

src/analog.rs
hannobraun: In addition to @austinbes's concerns, I'm confused about the overall purpose. What's the point of starting a continuous conversion, if I then have to poll the `read` method anyway?

src/lib.rs
austinbes: My experience lines up with your intuition here -- a rename to `adc` would be preferable IMO.

It's even what I started typing when bikeshedding the trait name, before I double checked.

src/analog.rs
austinbes: On such point might be to get a sample rate that's known, and regulated by hardware.

With that said, it would probably be more ergonomic (and more in-line with standard embedded paradigms) to somehow provide a buffer of samples rather than each individual one. I'm not sure the right way to do that in a memory-safe manner, though...

thenewwazoo: Whoops, you're totally right. `read` should, I think, probably take an `AdcChannel::CHANNEL`.

src/lib.rs
thenewwazoo: `adc` and `Once` works for me.

src/analog.rs
thenewwazoo: I mostly kept it there for consistency with the other modes. Nothing's preventing moving `Pin` into the various functions. If it's cleaner/easier, I'm not opposed.

src/analog.rs
thenewwazoo: D'oh. This, too, should take an `AdcChannel::CHANNEL` argument to release the owned pin.

src/analog.rs
thenewwazoo: The `read` method here would _likely_ be non-blocking; For example, in the (iirc) LPC178x, each channel has a specific register and the results of the most recent conversion for that specific channel are stored there. (There's also a register for last conversion result for _any_ channel.)

thenewwazoo:
It occurs to me that some ADC operations would be blocking (`Once`-mode reads), while some would not (`Continuous`-mode reads). There's a place in the library for blocking traits; should this go there? Can some part of it go there?

src/analog.rs
thenewwazoo: Sorry for the earlier deleted comments - I misunderstood your question.

Note that `read` takes a `Pin::ID` argument, not the `Pin` itself. So the code would look something like:
```rust
let pin: gpio::PA0<Analog> = gpioa.PA0.into_analog(); // PA0<Analog> impls AdcChannel<ADC1>
let adc = adc::init(...);
adc.add_channel(pin);
adc.start();
for _ in 0..=10 {
    let _ = adc.read(PA0::CHANNEL);
}
adc.stop();
let rpin = adc.release_channel(PA0::CHANNEL).unwrap();
```

src/analog.rs
thenewwazoo: Sorry, mistake here. `Continuous::release_channel` should indeed take a `Pin::ID` argument.

src/analog.rs
thenewwazoo: Also, I contemplated a way to provide access to a buffer, but I couldn't see a way until const generics land because the number of channels would need to be constant in order to know the size of the buffer provided.

src/analog.rs
therealprof: Yeah, `OneShot` sampling is the typical term here. I think it's also important to recognise what is being described here: Is this the acquisition or the trigger mode? At the moment it sounds more like a mix of the two.

- `Continuous`: Continuous sampling with single value capture (Immediate trigger)
- `Once`: Single sampling with single value capture (Immediate trigger)

`Continuous` is only mildly useful as it only saves the reinitialisation cost for the ADC but running an ADC continuously to just capture a value every now and then is rather energy intensive.

`Continuous` mode is far more interesting if you can do automatic sampling of many values at a specific rate into a fixed buffer (e.g. audio signal), rolling values into a circular buffer or automatically trigger a recording of a a single or burst of samples at some event (edge/level triggering on the signal).

thenewwazoo: I think it's fair to say that I'm trying to describe the trigger mode here since it's closest to how the hardware configuration is realized. I do agree that Continuous gets more interesting when you use it in the ways you describe. I don't see a clear path there, and I feel like there's kind of an implied DMA activity too, traits for which are not yet very well-defined.

src/analog.rs
hannobraun: I don't know if it's appropriate here, but I think the canonical way to work around the lack of const generics is [`generic-array`](https://crates.io/crates/generic-array).

src/analog.rs
hannobraun: I'm not opposed to keeping it where it is. Just wondering out loud.

hannobraun:
@thenewwazoo 
> It occurs to me that some ADC operations would be blocking (`Once`-mode reads), while some would not (`Continuous`-mode reads). There's a place in the library for blocking traits; should this go there? Can some part of it go there?

Why would `Once` reads be blocking? As currently proposed, `Once::read_channel` returns `nb::Result`.

thenewwazoo:
@hannobraun 
> Why would Once reads be blocking? As currently proposed, Once::read_channel returns nb::Result.

I don't suppose they'd have to be, but I imagined them as such because that's how I've used single-ended conversions in the past: request a conversion, wait until done, return the result. I don't think there's anything about these traits that would require one behavior or another - that could be up to the implementor.

I also think I should gate this behind unproven, which I haven't done yet, so another commit will be incoming soon.

hannobraun:
@thenewwazoo 
> I don't suppose they'd have to be, but I imagined them as such because that's how I've used single-ended conversions in the past: request a conversion, wait until done, return the result. I don't think there's anything about these traits that would require one behavior or another - that could be up to the implementor.

I think they absolutely should be non-blocking then. The precedent in this crate so far has been to provide blocking traits only in two cases:
1. As a convenience, implemented on top of a non-blocking trait.
2. When a non-blocking implementation would not be possible.

I've had some time to reflect on this pull request now, and I have the following concerns and suggestions:
- Everything needs to go behind `unproven`, as already noted.
- I'd prefer `OneShot` to `Once`, as one-shot seems to me the common term for this (as confirmed by @therealprof). I don't feel strongly about this point.
- The code talks about "pins" and "channels", which to me seem like two words for the same thing. I'd just call it "channel" in every case.
- I'm unsure if `Continuous` is the right approach. I'm not specifically saying it's wrong, but I'm concerned that it's too specific to be broadly implemented and of limited use. I'd feel more comfortable if we had a few implementations for different platforms, or someone with solid experience with the matter telling me my concern is unfounded. (My own experience with ADC HAL API implementations is limited and not recent.)
- To my question what the purpose of `Continuous` is, @austinbes answered that one purpose might be to have a hardware-regulated sample rate. Should the sample rate be an argument of the `start` method?
- Could `Once` and `Continuous` be traits that are implemented for Channels instead? That could cut down on a lot of complexity (no dealing with pin/channel ids).

@thejpster You've already approved this PR without otherwise taking part in the discussion. What's your take on my list of concerns? I'm very interested in more input.

thenewwazoo:
@hannobraun all good points, thanks for the feedback.

I split the concepts of pins and channels because the pin (the physical, mux-able component of the MCU) is distinct from the channel (the voltage source for the ADC conversion). In some cases (e.g. LPC1788) the mapping is 1:1, but in others, you can select between multiple pins. Much as `unsafe impl RxPin<$USART> for PX<AFn>`, I'm imagining `unsafe impl AdcChannel<$ADC> for PX<Analog>`. The ID type needs to be generic because in some cases you can have banks of channels, which is a possibility I want to permit. If the conceptual split is unclear, then there's definitely room to improve the docs (which is true _anyway_), or rename them to make it more obvious. I don't think the two concepts should be combined or conflated, though.

It seems `Continuous` mode isn't entirely clear.  The two chips I'm working with on-hand both have a mode whereby you specify one-or-many channels to sample and then start conversions. The results of the conversions are stored in registers that can be accessed immediately, and interrupts can be generated based on single conversion completion or completion of _all_ selected channels. (`Burst` mode is similar, except it will _not_ continue conversion after all channels have been sampled).

I'll call out separately that I feel _very_ strongly that sample rates should not be part of this API. In every ADC I've ever seen, sample rate is a property of the ADC's core and is determined by some clock input and a prescaler. IOW, I think `sample_rate` should be an argument to whatever `init` function (or perhaps `new` method) the ADC struct impls. If you need to adjust the sample rate (if it's even possible, which it may not be!) that's outside of the scope of these traits.

thenewwazoo:
I'll rename `Once` to `OneShot`.

Another possibility - move forward with the `OneShot` trait since that seems to be well-understood and straightforward, and open a separate PR for more complex ADC modes?

thenewwazoo:
Sorry for the comment spam - one last note. I think implementing these as non-blocking (that is, returning `nb::Result`) would be easy. In every case I've seen there's a way to check if the result of a conversion is complete. I'm starting to think that makes the most sense anyway.

hannobraun:
Thank you the explanation, @thenewwazoo. Your comments have cleared things up a lot for me. My objections have been reduced to a vague feeling that better traits are possible, but until I dive into some ADCs and try to implement, and improve on, your traits (which, for the foreseeable future, I won't have time for) that's a moot point.

I'm still interested in hearing more feedback from others, but as far as I'm concerned, I'm not opposed to merging this (at least for now; can't guarantee I won't come up with more feedback, given enough time :) ).

> Another possibility - move forward with the `OneShot` trait since that seems to be well-understood and straightforward, and open a separate PR for more complex ADC modes?

We can do that, but unless someone comes up with new objections, I don't think it's necessary.

---

There's one more thing I thought about. I think it's part of a larger discussion and shouldn't hold back this pull request, but at least I'd like to note it, while we're all here.

`Continuous` has the methods `start`, `stop` and `read`. Presumably calling `read` is only valid between `start` and `stop`. Calling it before `start` or after `stop` would be an error or a no-op, I assume. This constraint, `read` only being valid between `start` and `stop`, could be encoded in the traits, to enforce it at compile time.

There's precedent for keeping things simple and not enforcing this stuff at compile time (see [`timer::CountDown`](https://github.com/rust-embedded/embedded-hal/blob/f91fc4bf9588fb7887fe4d326468598e613aa1cd/src/timer.rs#L60), for example. But I think at some point we should have the discussion whether we want to do this.

src/adc.rs
```diff
@@ -0,0 +1,79 @@
+//! Analog-digital conversion traits
+
+use nb;
+
+/// A marker trait to identify MCU pins that can be used as inputs to an ADC channel.
+///
+/// This marker trait denotes an object, i.e. a GPIO pin, that is ready for use as an input to the
+/// ADC. As ADCs channels can be supplied by multiple pins, this trait defines the relationship
+/// between the physical interface and the ADC sampling buffer.
+pub trait AdcChannel {
```
hannobraun: I think we should rename this to `Channel`. The module is already called `adc`, and I think `adc::Channel` is much nicer than `adc::AdcChannel`.

src/adc.rs
```diff
@@ -0,0 +1,79 @@
...
+/// of the request (in contrast to continuous asynchronous sampling).
+pub trait OneShot<Word, Pin: AdcChannel> {
+    /// Error type returned by ADC methods
+    type Error;
+
+    /// Request that the ADC begin a conversion on the specified pin
+    ///
+    /// This method takes a `Pin` reference, as it is expected that the ADC will be able to sample
+    /// whatever channel underlies the pin.
+    fn read_channel(&mut self, pin: &mut Pin) -> nb::Result<Word, Self::Error>;
```
hannobraun: This is called `read_channel`, the method on `Continuous` is called `read`. I think both are fine names, but it may be better to keep it consistent and call both of them the same.

src/adc.rs
```diff
@@ -0,0 +1,79 @@
...
+    ///
+    /// This method moves the Pin, and configures the ADC to include this pin's channel in those
+    /// that it will sample asynchronously.
+    fn add_pin(&mut self, pin: Pin) -> nb::Result<(), Self::Error>;
+
+    /// Remove the specified channel from those sampled, and release its pin.
+    ///
+    /// This method removes the channel identified by `id` from the channels that the ADC
+    /// asynchronously samples. It then releases the `Pin` and returns it to the user.
+    fn release_pin(&mut self, id: Pin::ID) -> nb::Result<Pin, Self::Error>;
```
thenewwazoo: I just realized that this can't work the way I'm intending it to because one cannot return a trait object by value.

thenewwazoo:
Given @hannobraun's excellent points regarding the design of `Continuous`, plus my realization that the API can't work in the way I'd proposed, I'm going to cut it out of this PR, and revisit the design of more ADC operation modes later on.

src/adc.rs
burrbull: In my opinion trait better name `AdcChannel` and add it to prelude.
`Channel` name better reserve for `enum` of existent channels.

It would be something like this:
For 103xx:
```

pub trait AdcChannel {
    /// Channel ID type
    ///
    /// A type used to identify this ADC channel. For example, if the ADC has eight channels, this
    /// might be a `u8`. If the ADC has multiple banks of channels, it could be a tuple, like
    /// `(u8: bank_id, u8: channel_id)`.
    type ID;
    /// The specific ID that identifies this channel, for example `0` for the first ADC channel.
    const CHANNEL: Self::ID;

    fn channel(&self) -> Self::ID {
        Self::CHANNEL
    }
}

/// ADC_channels
#[derive(Clone, Copy, PartialEq)]
pub enum Channel {
    /// ADC Channel 0 selected
    C0 = 0x00,
    /// ADC Channel 1 selected
    C1 = 0x01,
    ...
    /// ADC Channel 15 selected
    C15 = 0x0F,
    /// ADC Channel 16 selected
    Temperature = 0x10,
    /// ADC Channel 17 selected
    Voltage = 0x11,
}

macro_rules! analog_input {
    ($PXi:ident, $CHx:ident, $SMPRx:ident) => {
        impl AdcChannel for $PXi<Analog> {
            type ID = Channel;
            const CHANNEL: Channel = Channel::$CHx;
        }
    };
}
analog_input!(PA0, C0);
```

thenewwazoo:
@ryankurte I'm happy to drop a link in the doc pointing to my own crate which impls this if that's acceptable; I don't think I'll have time for a few days to work on a minimal-esque example.

src/adc.rs
hannobraun: @burrbull Traits in the prelude are re-exported under a different name to prevent name collisions, so that shouldn't be an issue. And I image non-prelude use would not be an issue either, as an implementation could do it like this:
``` rust
use embedded_hal::adc;

impl adc::Channel for MyChannel {
    // ...
}

pub enum Channel {
    // ...
}
```

ryankurte:
@thenewwazoo even just here for us to look at would be neat ^_^

thenewwazoo:
Sure! You can see my impl [here](https://github.com/thenewwazoo/stm32l0x1-hal/blob/master/src/adc.rs) for the STM32L0x1. I can't publish yet (it's pending this merge) so you'll have to generate docs yourself if you want to see them rendered.

jonwingfield:
Waiting anxiously for this! 

Also worth noting:
https://gitlab.henriktjader.com/pln/nucleo-64/commit/23d83bac3e26b5301bc54227a4bc2249150e43a8
I don't know if @japaric still prefers the DMA/ring-buffer method, but I like the idea!

thenewwazoo:
Do I need to do anything more on this? I'm eagerly anticipating its merge. :)

hannobraun:
Sorry, @thenewwazoo, I didn't notice that you pushed a new commit.

bors r+

hannobraun:
Oh, and thank you for your patience and working with us on this!

thenewwazoo:
Huzzah!

thenewwazoo:
Now the next question is: when can this be released (i.e. `cargo publish`ed)? :)

hannobraun:
@thenewwazoo 
> Now the next question is: when can this be released (i.e. `cargo publish`ed)? :)

Pretty much right now. Consensus in #92 was to release a new patch version right after landing this. Just requires someone from the team finding the time to do it.

Issue #41, opened 2018-02-15
`digital::InputPin` API


japaric:
This trait became available in release v0.1.1 behind the "unproven" Cargo feature.

therealprof:
Implemented in nrf51-hal = "0.4.1" and used in microbit = "0.4.1".

Nemo157:
Would this be the correct trait to include some sort of simple edge detection API (e.g. for buttons hooked up to GPIO)?

Caemor:
Does this api still needs more drivers?

It seems to be quite stable since it's introduction half a year ago.

Caemor:
The only problem I found with this api since using it is that errors can't get propagated any further and need to be converted to a default value. 

E.g. the linux sys_fs can receive an file error while trying to read the value (https://github.com/rust-embedded/rust-sysfs-gpio/blob/master/src/lib.rs#L337) and this can't be handled by a simple bool.

eldruin:
I also stumbled at the same problem as @Caemor with the `OutputPin` trait as well. In an I2C I/O expander driver I am writing I want to provide access to the individual pins as a bunch of `Px` structs implementing `OutputPin` and `InputPin`. At the moment I saw no other choice but to [panic if there was any communication problem](https://github.com/eldruin/pcf857x-rs/blob/a1ea28f9b333d066fe44f1453e00f3fca17c3121/src/pins.rs#L81). I was about to implement the `InputPin` traits and saw that I would have the same problem again.

Maybe we should discuss this in a separate issue?

ryankurte:
Interesting problems! I think as this spans all operations (not just `Digital::InputPin`) another issue would be great ^_^

nils-van-zuijlen:
If this API does not need driver nor implementation, when will it be proven ?

therealprof:
The problem is: There's no process to mark something as proven. There's a new RFC open #163 in attempt to completely revamp the way we get things added/changed.

Since there're enough implementations already we could simply accept a PR to mark it proven.

rubberduck203:
I think this issue can be closed.

therealprof:
Fair enough.

Issue #100, opened 2018-10-04
[RFC] Managing breaking changes to existing traits


ryankurte:
As highlighted in #95, #97 and #99, there are likely to be situations in which we need to fix or upgrade traits in breaking ways, and these are likely to have significant ramifications on the embedded ecosystem. This RFC proposes an approach to manage these breaking changes.

### Goals:
- Mitigate the downstream breakage caused by embedded-hal breaking changes
- Communicate breaking changes to give library uses an opportunity to update

### Approach:
We propose a staged approach for managing breaking changes, with some time between each stage to allow API consumers to update.

In the first release, replacement traits are introduced and feature-gated allowing users to opt in to their use. Feature names start with with `use-` and a description of the change, for example `use-fallible-digital-traits` for the new fallible digital trait update in #97. Existing traits are marked `#[deprecated]` to alert users to the need to update, and this opt-in flag should be added to the CI build matrix. Documentation should also be updated to recommend the use of the opt-in traits and the related github issue for the change. This stage alerts HAL users of deprecation and allows immediate migration should they desire, without requiring any explicit action.

In the second release, the opt-in feature flag is replaced with an opt-out flag to allow regression to the previous trait if required, this should start with `regress-`. Continuing the fallible digital trait example this would be `regress-infallible-digital-traits`. This opt-out flag should be added to the build matrix, and the previous flag removed. Documentation should be updated to mention opt-out trait availability if required. This stage will break HAL consumers using the previous traits, requiring explicit action to update or defer the change.

In the third release, the old traits are removed, feature flags are removed from the package and the CI build matrix, and the documentation should be updated to remove mention of the opt-out trait. This stage will break any hal consumers that have not updated to the new traits.

In each release, these changes will be documented in the changelog (as is standard) and should be published through other useful means. The linking of the related issue in the documentation should allow for feedback from HAL consumers at each stage.
Each stage will correspond to a semver version increase (though it is worth noting that more than one breaking change may be included in a single release). At `0.x` this will be `patch, minor, minor`, and `>=1.x`: will be `minor, major, major`.

I also wrote a small tool to analyse the dependency requirements, resolutions, and features of published crates (https://github.com/ryankurte/rust-dep-check), which would let us track migration through versions and feature flags. An example output showing the current use of `embedded-hal`:

```
Found 100 crates using 'embedded-hal'
Version requirements:
	^0.1	(0.1.3):	   1 / 100 (1.00 %)
[...]

PR #102, opened 2018-10-11
Remove InputPin unproven flag per #41

ryankurte:
`InputPin` trait has been demonstrated (see: https://github.com/rust-embedded/embedded-hal/issues/41#issuecomment-368315610 amongst other impls), this PR removes the `unproven` feature flag. This is either blocked by or dependent on #95 and #97 (what do you think, @eldruin?), and blocks https://github.com/rust-embedded/linux-embedded-hal/pull/11

### DIFF
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -109,9 +109,6 @@ pub mod toggleable {
 }
 
 /// Single digital input pin
-///
-/// *This trait is available if embedded-hal is built with the `"unproven"` feature.*
-#[cfg(feature = "unproven")]
 pub trait InputPin {
     /// Is the input pin high?
     fn is_high(&self) -> bool;


eldruin:
Thanks @ryankurte. It would be easiest for me if this was done after landing #97. I will push some changes to #97 this afternoon and I think it should be quite ready for merge then.

However, landing this would mean declaring an infallible trait proven and at the same time it is clear that we want to replace it with a fallible version.
What do you think?
An alternative could be to leave the traits behind `unproven` for this release and in the next release make only the fallible version proven.

ryankurte:
Ahh, you make a good point. Probably better to leave it unproven and close this and wait I suspect. Thanks for the response!

therealprof:
So here's the thing: Marking something as proven is a compatible change so we can easily do that anytime (and we should have a long time ago). The fallible version OTOH is a breaking change so it will require a bit more coordination (and pooling with other breaking changes into a new minor release) and it is going to be unproven (unless there's enough implementations already, to make it proven).

So for me those two topics are really separate issues and I would go ahead and make the compatible change now and the other change whenever everything is ready.

@rust-embedded/hal and all, really: Opinions?

hannobraun:
Question: We're about to change those traits, so would that not reset their proven-ness? I mean, the new fallible version won't initially have the number of required implementations and users.

And how will we handle this in general? If we decide to handle breaking releases as clean cuts as opposed to having a transition strategy (see #100), will it ever be okay to re-add a trait to the unproven set? Seems user-unfriendly to remove a proven trait and require the user not only to upgrade, but opt into `unproven`. But if we discover problems in a trait after it had been in use for a while, then the fixed version of the trait _is_ essentially unproven.

I'm not sure if that's going to be my position going forward, but this certainly is one of those moments where I'm thinking that the whole `unproven` concept is a useless waste of effort, as long as we're pre-1.0.

therealprof:
@ryankurte Care to check out the conflicts?

eldruin:
As I understood the discussion, implementing #97 as a clean break would mean the replacement fallible version of the `InputPin` would have to be `unproven`, right?

Disasm:
Ping from triage: is there any progress on this?

mathk:
Ping from triage: Can we rebase and merge ?

therealprof:
Hm, I guess the moment we land this someone is going to request a fallible version of it. 😅

mathk:
Rich Hickey has a good point saying that union type is better than class type as it break compatibility.

therealprof:
> Rich Hickey has a good point saying that union type is better than class type as it avoid breaking compatibility.

Must have missend that comment. Have a link to it?

mathk:
https://www.youtube.com/watch?v=YR5WdGrpoug

eldruin:
> Hm, I guess the moment we land this someone is going to request a fallible version of it. sweat_smile

I already did :) I don't do it for fun, just that I want to get rid of [panicking on error](https://github.com/eldruin/pcf857x-rs/blob/3f5361545906c49716816667810ecfcb64c1ce5c/src/pins.rs#L103). At the moment I have no other choice.

mathk:
@eldruin I guess what @therealprof  says is that someone would like the trait to have `fn is_high(&self) -> Result<bool>;` instead of bool.

caemor:
@mathk That (`Result<bool>`) is what Eldruin already needs to get rid of panicking on these errors (see eldruins link above) ;-)

ryankurte:
this PR applies to a previous version of the library, so i'm going to close it as no-longer-required.

feel free to open an equivalent if you would like to mark the `digital::v2` traits as proven, check out and contribute to existing discussion about the v2 traits at #135, or open a new issue for anything else ^_^

PR #107, opened 2018-10-25
Update changelog for v0.2.2 release and tick version

thenewwazoo:
Added updates to changelog (worded them as best I could based on reading PRs and code).

AFAICT there are no breaking changes here (e.g. those being discussed in #100)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.2.2] - 2018-10-25
+
+### Added
+
+- Added the Rust Code of Conduct to this repository
+- The first ADC-related trait. This is a simple trait for one-shot conversions.
+- Iterator-based blocking write and write+read traits have been added to I2C and SPI.
+- New helper constants for SPI modes.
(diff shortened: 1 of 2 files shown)


thenewwazoo:
Just a friendly ping to see if this can land soon. I've got a handful of HAL crates waiting on this before publishing.

ryankurte:
Hey, thanks for the ping. My perspective is that I'd like to see adaptors between the new/old traits before we publish a new release so we don't end up with fragmentation/incompatibility between implementations of the new/old versions, and I think this is something that we need agreement on first (maybe in #100).

ryankurte:
Ahh, sorry based on previous discussion in #92 that'll be `0.2.3`. This looks good to me.

bors r+

CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.2.2] - 2018-11-03
```
therealprof: @thenewwazoo The link definition for the changes between v0.2.1 and v0.2.2 is missing.

therealprof:
This has been published to crates.io.

PR #108, opened 2018-10-28
Add fallible version of the digital traits and deprecate the current ones

eldruin:
There seems to be an agreement in [#100](https://github.com/rust-embedded/embedded-hal/issues/100#issuecomment-433116884) on how to proceed with the fallible traits.
This adds the fallible traits under `digital::v2` and marks the current ones as deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- A new version of the digital `OutputPin`, `StatefulOutputPin`, `ToggleableOutputPin`
+  and `InputPin` traits has been added under `digital::v2`. These traits are now
+  fallible and their methods now return a `Result` type as setting an output pin
+  and reading an input pin could potentially fail.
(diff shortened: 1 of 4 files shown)


therealprof:
@hannobraun @ryankurte Are you okay with this?

therealprof:
@rust-embedded/hal Any other opinions?

ryankurte:
whoops

bors r+

therealprof:
Funny complaints from the nightly compiler. 👎 

@rust-embedded/resources Should we make the nightly compiler builds informal only, i.e. not fail CI? Our main target is beta/stable anyways.

caemor:
`cargo test` is only run on nightly atm.

This might be because two of the doc tests rely on nightly features (https://github.com/rust-embedded/embedded-hal/blob/master/src/lib.rs#L385 ++ and https://github.com/rust-embedded/embedded-hal/blob/master/src/lib.rs#L557 ++). Maybe the doc test for these two should be tagged with the `ignore` setting until these features are stable/beta.

but either way the failed doc test doesn't have anything to do with this PR.

ryankurte:
bors try

PR #111, opened 2018-11-10
impl Write for serial::Write

thenewwazoo:
Shamelessly pilfered from [@therealprof's code](https://github.com/therealprof/stm32f042-hal/blob/master/src/serial.rs). Still needs Changelog entry, but I thought this was a good idea after a discussion on IRC.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -1,0 +1,17 @@
+//! Implementation of `core::fmt::Write` for the HAL's `serial::Write`.
+//!
+//! TODO write example of usage
+use core::fmt::{Result, Write};
+
+impl<Word, Error> Write for ::serial::Write<Word, Error=Error>
+where
+    Word: From<u8>,
+{
+    fn write_str(&mut self, s: &str) -> Result {
+        let _ = s.as_bytes()
+            .into_iter()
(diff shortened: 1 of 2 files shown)


thenewwazoo:
Per an IRC comment, if this is fully generic, do we need the featurengate at all?

therealprof:
I think that was my comment. ;) Just to clarify, I think no.

@thenewwazoo Doesn't this require to pull in the `Write` trait to work?

thenewwazoo:
That's... a good question. I was under the (mistaken?) impression that `core::fmt::Write` was in `core::prelude`.

thenewwazoo:
Looks like this build was a spurious failure - any chance it could get merged? I'm not sure how to fix the build failure (or if it even should be fixed?)

therealprof:
bors retry

Disasm:
Blocked on https://github.com/rust-embedded/embedded-hal/pull/125

mathk:
@thenewwazoo does it worth doing a rebase and see if that solve the bors issue ?

Disasm:
bors r=ryankurte

Disasm:
bors r=ryankurte

PR #120, opened 2019-01-02
Document implicit assumption in non-blocking spi trait

david-sawatzke:
It should only be able to read the data returned by the last write call. Otherwise the overflow handling isn't really possible when doing only write calls (at least on stm32f0). @therealprof and me came to this conclusion after chatting on irc that this is the intended usage of this api.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -8,12 +8,12 @@ use nb;
 ///
 /// - It's the task of the user of this interface to manage the slave select lines
 ///
-/// - Due to how full duplex SPI works each `read` call must be preceded by a `send` call.
+/// - Due to how full duplex SPI works each `try_read` call must be preceded by a `try_send` call.
 ///
-/// - `read` calls only return the data received with the last `send` call.
+/// - `try_read` calls only return the data received with the last `send` call.
 /// Previously received data is discarded
 ///
-/// - Data is only guaranteed to be clocked out when the `read` call succeeds.
+/// - Data is only guaranteed to be clocked out when the `try_read` call succeeds.
 /// The slave select line shouldn't be released before that.
 ///
 /// - Some SPIs can work with 8-bit *and* 16-bit words. You can overload this trait with different
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -13,6 +13,9 @@ use nb;
 /// - `read` calls only return the data received with the last `send` call.
 /// Previously received data is discarded
 ///
+/// - Data is only guaranteed to be clocked out when the `read` call succeeds.
+/// The slave select line shouldn't be released before that.
+///
 /// - Some SPIs can work with 8-bit *and* 16-bit words. You can overload this trait with different
 /// `Word` types to allow operation in both modes.
 pub trait FullDuplex<Word> {
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -10,6 +10,9 @@ use nb;
 ///
 /// - Due to how full duplex SPI works each `read` call must be preceded by a `send` call.
 ///
+/// - `read` calls only return the data received with the last `send` call.
+/// Previously received data is discarded
+///
 /// - Some SPIs can work with 8-bit *and* 16-bit words. You can overload this trait with different
 /// `Word` types to allow operation in both modes.
 pub trait FullDuplex<Word> {
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -10,7 +10,7 @@ use nb;
 ///
 /// - Due to how full duplex SPI works each `try_read` call must be preceded by a `try_send` call.
 ///
-/// - `try_read` calls only return the data received with the last `send` call.
+/// - `try_read` calls only return the data received with the last `try_send` call.
 /// Previously received data is discarded
 ///
 /// - Data is only guaranteed to be clocked out when the `try_read` call succeeds.


david-sawatzke:
Removed the change for the default implementation, otherwise releasing cs after doing a blocking write will lead to unintended consequences

eldruin:
Any progress here?

david-sawatzke:
I've thought about this a bit more: If any hal (with hardware fifos) wants an efficient implementation with these restraints,
they'd have to count the current spi transactions. This would also make utilizing the receive fifo pretty much impossible.
This slows bidirectional communication (which, in my experience, is pretty uncommon) down, especially when considering techniques like dmas or platforms like linux.

In contrast leaving it like it's currently often implemented (just giving an error if the receive fifo overflows and otherwise returning the oldest byte) makes it uncertain when the fifo overflows,
forcing chip independent implementations to read after every write (and worse, if the peripheral is somehow shared, making off-by-one errors possible).
But it enables chip-specific implementations to utilize the receive fifo to it's full potential.

I'm not happy with both solutions, but the proposed one sounds a lot better to me. Open to alternatives

As-is, this PR is done, not needing more progress, but some discussion could improve it. 

(I think) This is also a breaking change, since existing implementations didn't have to consider any restraints.

src/spi.rs
```diff
@@ -10,6 +10,12 @@ use nb;
 ///
 /// - Due to how full duplex SPI works each `read` call must be preceded by a `send` call.
 ///
+/// - `read` calls only return the data received with the last `send` call.
+/// Previously received data is discarded
```
Disasm: I hope it isn't true for SPI controllers with FIFO. In case you have no overflows, you can extract data from FIFO in semantically correct order.

david-sawatzke: You can either force device-independent implementations to read directly after every write (the rest is undefined behaviour, since it's unclear which byte gets returned [or an error]) or force the read function to only return the latest byte. Both of these are non-ideal and would have to be documented.

The first option gets worse with shared spi devices, since some device-specific implementation could rely on the rx fifo, which would break drivers that rely on the calls matching (the latest `read` call returning the data transmitted with the `write` call). It makes utilizing the tx fifo impossible for device-independent code. 

Utilizing the rx-fifo requires in-depth knowledge of your implementation, since it possibly only makes use of the spi fifos or does it's own, higher level interrupt/dma-driven stuff. If you need the performance, it may be a better idea to just bypass the hal, since it needs to be device-specific anyways.

(Dropping this constraint also makes the second one impossible)

Disasm: > You can either force device-independent implementations to read directly after every write (the rest is undefined behaviour, since it's unclear which byte gets returned [or an error])

Yes, this is how it's done [in this helper function](https://docs.rs/embedded-hal/0.2.2/src/embedded_hal/blocking/spi.rs.html#34-55). Also with this function you can utilize RX FIFO without even without knowledge of its existence.

> The first option gets worse with shared spi devices, since some device-specific implementation could rely on the rx fifo, which would break drivers that rely on the calls matching

What do you mean under "shared spi devices"? In Rust we try to have nothing that could be shared in unsafe way.

david-sawatzke: >  In Rust we try to have nothing that could be shared in unsafe way.

*Memory-unsafe way. If you share a device with a mutex (for example with an interrupt), you have to take care, that the `read` & `write` calls match up.

> you can utilize RX FIFO even without knowledge of its existence.

You could. But for that use case it wouldn't make much difference.

Disasm: Yes, you have to use the device in a way that prevents simultaneous access during a single transaction.

Disasm:
It seems we need more discussion here. Will you be able to participate in our [weekly meeting](https://github.com/rust-embedded/wg/issues/318)? Next will be on Feb 16.

src/spi.rs
david-sawatzke: As the trait doesn't have a notion of transactions, this is still an issue.
For example:
```
fn main() {
  let cs1 = ...;
  let cs2 = ...;
  let spi = ...;
  f1(&mut spi, &mut cs1);
  f2(&mut spi, &mut cs2);
}

// Driver 1
fn f1(spi: &mut ..., cs: &mut ...) {
    cs.try_set_low();
    block!(spi.try_send(0b00));
    cs.try_set_high();
}

// Driver 2
fn f1(spi: &mut ..., cs: &mut ...) {
    cs.try_set_low();
    block!(spi.try_send(0b00));
   // `ret` is now dependent on the implementation
   let ret = block!(spi.try_read());
   cs.try_set_high();
}
```

The only way to prevent this is to *always* pair send & read calls, nullifying any benefit of the rx fifo.

david-sawatzke:
As this hasn't been resolved yet, I'd like to try & get this (or some alternative) into v1.0.

eldruin:
Any progress here?

david-sawatzke:
This is basically ready as-is. If we want to do the simpler, but way less useful alternative of unspecified behaviour for unpaired read & send calls, that would also work, I don't care much either way, I just want there to be some thought given to it & documented instead of "the most convenient interpretation".

src/spi.rs
```diff
@@ -8,7 +8,13 @@ use nb;
 ///
 /// - It's the task of the user of this interface to manage the slave select lines
 ///
-/// - Due to how full duplex SPI works each `read` call must be preceded by a `send` call.
+/// - Due to how full duplex SPI works each `try_read` call must be preceded by a `try_send` call.
+///
+/// - `try_read` calls only return the data received with the last `send` call.
```
Disasm, line 13:
```suggestion
/// - `try_read` calls only return the data received with the last `try_send` call.
```

Disasm: Other than that looks good to me.

therealprof (approved):
LGTM

bors r+

david-sawatzke:
Maybe the unresolved review? I've resolved it, might work now.

therealprof:
Nope, bors is picky about that. @Disasm will need to explicitly approve to get rid of the concern.

Disasm (approved):

Disasm:
bors r=therealprof

Issue #123, opened 2019-01-31
ADC OneShot: Clarify asynchronous behavior


chrysn:
The OneShot trait defines returning a nb::Result, but it's not fully clear to me what a a WouldBlock means; could be either of

* Can't do, there's another ADC conversion in progress (should really not happen, b/c it takes a `&mut self`, but maybe it's on a shared bus, or maybe there was a channel switch and some calibration time needs to be accounted for or whatever)
* Yes I started it, but the result is not ready yet.

If it can be the latter, that begs the question of whether calling .read() incurs the risk of not actually getting a fresh value but the result an earlier read attempt that was aborted – and a caller can't rely on an instant return to mean "this has been requested earlier" either, for some ADCs might give results instantly.

My gut feeling is that nb::Result functions should be "Retry means I ignored you (and when I do something, I return something)" and not "Retry means I heard you (but I'll ignore any future requests until I've once returned something)" – but I might be wrong there and it doesn't need to be so strict to still give reliable oneshot results. What are other peripherals doing there? SPI doesn't have that issue as it has a send and a read method, maybe OneShot should too?

HarkonenBade:
My general opinion is that an ADC implementation should reject attempts to 'read' different channels if there is an in progress conversion. I believe the use of nb is because ADC conversions can take a very long time and it may be advantageous to perform them asynchronously. It might be worth adding an explicit 'cancel' call that allows you to end the current conversion. So that in the case you need to force a new read through at the expense of an older one, you can call the cancel method to force the end of the previous read.

therealprof:
I agree. Real errors should always result in an `Err()` != `Err(nb:WouldBlock)`.

chrysn:
I don't think that a proper state machine *can* avoid the risk of having stale values, for that state machine doesn't get told when an attempt to read is aborted, and when a new read is started.

Having a `.cancel()` would help a lot; whoever cares about having a fresh read would cancel and then read. (And the analogous .start()/.read() could be implemented in terms of it.)

chrysn:
(Thinking on this a bit longer, other errors than "still waiting" are really not the point here, and the first bullet on my original post which @therealprof probably commented on is rather moot).

HarkonenBade:
> I don't think that a proper state machine _can_ avoid the risk of having stale values, for that state machine doesn't get told when an attempt to read is aborted, and when a new read is started.
> 
[...]

PR #124, opened 2019-01-31
[WIP] Initial proposal for CancellableOneShot

chrysn:
Add a trait that extends OneShot and allows cancellation, along with an incomplete example.

This is a first shot at #123; not ready for review as in "ensure we can merge this", but pushed into a PR to have actual discussion material in #123.

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -96,3 +96,37 @@ pub trait OneShot<ADC, Word, Pin: Channel<ADC>> {
     /// whatever channel underlies the pin.
     fn read(&mut self, pin: &mut Pin) -> nb::Result<Word, Self::Error>;
 }
+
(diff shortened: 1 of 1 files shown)


therealprof:
... but does not compile:
```
122 | pub trait CancellableOneShot<ADC, Word, Pin: Channel<ADC>>: OneShot<ADC, Word, Pin, Channel<ADC>> {
    |                                                                                     ^^^^^^^^^^^^ unexpected type argument
```

chrysn:
Oups, tested without the unproven feature.

Fixed, and `cargo +nightly test --features unproven` now runs through the test as well after having ironed out the details.

therealprof:
Unfortunately this has staled out so after some consideration we're going to close this.

Issue #30, opened 2018-01-20
[RFC] `digital::OutputPort`, atomically drive several pins


japaric:
In the v0.1.0 release of this crate we have a `digital::OutputPin` that represents a single digital output pin. That trait is useful for, e.g., implementing the NSS pin of a SPI interface but not enough to implement 8 or 16 bit parallel port interfaces where all the pins need to change at the same time (atomically) to meet timing requirements (e.g. LCD interfaces) -- using 16 `impl OutputPin` would result in them changing state at different time intervals.

I think the obvious trait to use would be the following:

``` rust
/// A digital output "port"
///
/// `Width` is the size of the port; it could be `u8` for an 8-bit parallel
/// port, `u16` for a 16-bit one, etc.
///
/// **NOTE** The "port" doesn't necessarily has to match a hardware GPIO port;
/// it could for instance be a 4-bit ports made up of non contiguous pins, say
/// `PA0`, `PA3`, `PA10` and `PA13`.
pub trait OutputPort<Width> {
    /// Outputs `word` on the port pins
    ///
    /// # Contract
    ///
    /// The state of all the port pins will change atomically ("at the same time"). This usually
    /// means that state of all the pins will be changed in a single register operation.
    fn output(&mut self, word: Width);
}
```

cc @kunerd

kunerd:
Never thought about this, but you are right, without an atomic set the communication timing could become broken by interrupts. But, wouldn't we than need an `InputPort` and an `InputOutputPort`, too?

japaric:
> But, wouldn't we than need an InputPort and an InputOutputPort, too?

Possibly. You let me know what *you* need for your driver. :-)

idubrov:
I like the idea of being able to drive multiple pins separately as a single atomic thing -- for the timing reasons you've mentioned.

However, for the purpose of HD44780 LCD, I think, timings on data output are less important than being able to use pins from _different_ GPIO ports (like GPIOA/GPIOB).

Even if bit operation is interrupted by, well, interrupt, it does not matter for HD44780 as data is transferred via dedicated "pulse" on EN bit after all data pins are set.

Given that flexibility of LCD, it would be tempting to assign pins to it which are "left" from the more demanding devices (like those requiring timer PWM, atomic operations on across multiple pins, etc).

Theoretically, you can design this trait to be able to span multiple ports and make it "do its best" to do as little operations as possible...

japaric:
[...]

PR #134, opened 2019-05-09
OutputPort proposal

sajattack:
Shamefully stolen from @japaric's issue #30.

I needs for my parallel LCDs.

### DIFF
--- a/src/digital/v2.rs
+++ b/src/digital/v2.rs
@@ -122,6 +122,27 @@ pub mod toggleable {
     }
 }
 
+/// A digital output "port"
+///
+/// `Width` is the size of the port; it could be `u8` for an 8-bit parallel
+/// port, `u16` for a 16-bit one, etc.
+///
+/// **NOTE** The "port" doesn't necessarily has to match a hardware GPIO port;
+/// it could for instance be a 4-bit ports made up of non contiguous pins, say
+/// `PA0`, `PA3`, `PA10` and `PA13`.
+#[cfg(feature = "unproven")]
+pub trait OutputPort<Width> {
+    /// Error type
+    type Error;
+    /// Outputs `word` on the port pins
+    ///
+    /// # Contract
+    ///
+    /// The state of all the port pins will change atomically ("at the same time"). This usually
+    /// means that state of all the pins will be changed in a single register operation.
+    fn write(&mut self, word: Width) -> Result<(), Self::Error>;
+}
+
 /// Single digital input pin
 ///
 /// *This trait is available if embedded-hal is built with the `"unproven"` feature.*


sajattack:
Here's a sample implementation:
```
impl OutputPort<u32> for Port {
    fn write(&mut self, word: u32) {
        unsafe {
            (*PORT::ptr()).$outset.write(|bits| {
                bits.bits(word)
            });
        }
    }
}
```

ryankurte:
hey thanks for the PR! looks pretty good to me.

the only question i have is how we define / configure buses of widths that are not an existing (standard integer width) type, and can this be made clear in the interface and/or documentation?

(cc. #30)

therealprof:
We talked about this a bit on IRC earlier today. I think there're few other problems with the comments and I'd certainly would like to see a *slightly* more complete impl doing something other than simply exposing the full `set` register...

sajattack:
Maybe we could have a mask parameter, or use the uX crate?

luojia65:
https://github.com/rust-embedded/embedded-hal/issues/30#issuecomment-557439768
Should the width be a type parameter in trait other than in generic type bound? That could be a more reasonable design as typically the chip only allow one width type for one port (as far as I know).

burrbull:
I've made testing implementation of `OutputPort`:
https://github.com/burrbull/embedded-hal/tree/outport rebased on 0.2.x + const generic port size (`N`)
https://github.com/burrbull/hd44780-driver/commit/cc8861bd5f8ea352472b9f42b88175f395b844a9 - using in hd44780-driver
https://github.com/stm32-rs/stm32f4xx-hal/pull/426 - `OutputPort<N, u8>` implementation example for F4

eldruin:
Thanks for your test implementation @burrbull.
For reference, here is the code:
```rust
/// A digital output "port"
///
/// `N` is number of pins in "port"
///
/// `Width` is the size of the data; it could be `u8` for an 8-bit parallel
/// port, `u16` for a 16-bit one, etc.
///
/// **NOTE** The "port" doesn't necessarily has to match a hardware GPIO port;
/// it could for instance be a 4-bit ports made up of non contiguous pins, say
/// `PA0`, `PA3`, `PA10` and `PA13`.
#[cfg(feature = "unproven")]
pub trait OutputPort<const N: u8, Width> {
    /// Error type
    type Error;
    /// Outputs `word` on the port pins
    ///
    /// # Contract
    ///
    /// The state of all the port pins will change atomically ("at the same time"). This usually
    /// means that state of all the pins will be changed in a single register operation.
    fn write(&mut self, word: Width) -> Result<(), Self::Error>;
}
```
A few questions/remarks:
1. The unconstrained `Width` parameter will be a problem for interoperability.
1. What do do when `N` and the number of bits of `Width` differ?
1. What to do with widths that do not match primitive types like `u8`/`u16`... ?
1. Little-endian vs big-endian words.
1. How to send several words back-to-back?
1. While this allows for HAL implementations to define `OutputPort`s, it is not possible for the user to define these. e.g. it is not possible for the user to get a 4-bit `OutputPort` made of `PA0`, `PA3`, `PA10` and `PA13` like stated in the docs if the HAL impl does not provide it as such. However, HALs can only provide so many pin combinations and `OutputPort` sizes without falling into macro madness.

burrbull:
What if limit trait to `u8` only? Just delete `Width`.

burrbull:
> 6\. While this allows for HAL implementations to define `OutputPort`s, it is not possible for the user to define these. e.g. it is not possible for the user to get a 4-bit `OutputPort` made of `PA0`, `PA3`, `PA10` and `PA13` like stated in the docs if the HAL impl does not provide it as such. However, HALs can only provide so many pin combinations and `OutputPort` sizes without falling into macro madness.

If you have looked at implementation example, you could see that it support all combinations of pins located on 1 real port.

burrbull:
@eldruin 
Look at this please:
```rust
#[cfg(feature = "unproven")]
pub trait OutputPort<const N: usize> {
    /// Error type
    type Error;
    /// Outputs `word` on the port pins
    ///
    /// # Contract
    ///
    /// The state of all the port pins will change atomically ("at the same time"). This usually
    /// means that state of all the pins will be changed in a single register operation.
    fn write(&mut self, word: u8) -> Result<(), Self::Error>;
}

#[cfg(feature = "unproven")]
/// Port containing array of [OutputPin]
pub struct DumbPort<PIN, const N: usize>
where
    PIN: OutputPin,
{
    pins: [PIN; N],
}

#[cfg(feature = "unproven")]
impl<PIN, const N: usize> DumbPort<PIN, N>
where
    PIN: OutputPin,
{
    /// Collect pins to port
    pub fn new(pins: [PIN; N]) -> Self {
        Self { pins }
    }
}

#[cfg(feature = "unproven")]
impl<PIN, const N: usize> OutputPort<N> for DumbPort<PIN, N>
where
    PIN: OutputPin,
{
    type Error = PIN::Error;
    fn write(&mut self, word: u8) -> Result<(), Self::Error> {
        for i in 0..N {
            if word & (1 << i) != 0 {
                self.pins[i].set_high()?;
            } else {
                self.pins[i].set_low()?;
            }
        }
        Ok(())
    }
}

```

thejpster:
How does your `write` function upload the contract:

> The state of all the port pins will change atomically ("at the same time"). This usually means that state of all the pins will be changed in a single register operation.

burrbull:
> How does your `write` function upload the contract:
> 
> > The state of all the port pins will change atomically ("at the same time"). This usually means that state of all the pins will be changed in a single register operation.

It is just dumb implementation above for HALs which don't have native support of OutputPort.

More relevant implementation example is in https://github.com/stm32-rs/stm32f4xx-hal/blob/d94044188defb82583bd561fe8fb3363932454cb/src/gpio/outport.rs

burrbull:
> 1. What about ports with more than 8 bits?

This is main question.
But what is real use-case?
Even if we solve endianess questions (which is not trivial), how to implement it considering "at the same time"? Pins must be on 1 real port for this. I can imagine only specialized chips (like shift-regiters) in this role.
Also ports in microcontrollers have limitations on power.
Maybe I missing something important?

PR #136, opened 2019-05-11
improved digital compatibility docs

ryankurte:


src/digital/v1_compat.rs
```diff
@@ -1,7 +1,39 @@
-//! v1 compatibility wrapper
-//! this module adds reverse support for v2 digital traits
-//! v2 traits must be explicitly cast to the v1 version using `.into()`,
-//! and will panic on internal errors
+//! v1 compatibility wrappers
+//! 
+//! this module provides wrappers to support use of v2 implementations with
+//! v1 consumers. v2 traits must be explicitly cast to the v1 version using 
+//! `.into()`, and will panic on internal errors
```
dbrgn: I don't want to be nitpicky, but correct capitalization and punctuation would be nice.

src/digital/v1_compat.rs
```diff
@@ -1,7 +1,39 @@
...
+//! struct OldOutputPinConsumer<T: v1::OutputPin> {
+//!     _pin: T,
+//! }
+//! 
+//! impl <T>OldOutputPinConsumer<T> 
+//! where T: v1::OutputPin {
+//!     pub fn new(pin: T) -> OldOutputPinConsumer<T> {
+//!         OldOutputPinConsumer{ _pin: pin }
+//!     }
+//! }
```
dbrgn: The docs could be a bit shorter if you'd skip the `new` function and simply make the `_pin` field `pub pin` instead, and then directly assign the value in `main`.

src/digital/v2_compat.rs
```diff
@@ -1,5 +1,36 @@
 //! v2 compatibility shims
+//! 
 //! this module adds implicit forward support to v1 digital traits
+//! allowing v1 implementations to be directly used with v2 consumers.
```
dbrgn: Same thing here regarding capitalization and punctuation.

src/digital/v2_compat.rs
```diff
@@ -1,5 +1,36 @@
...
+//! struct NewOutputPinConsumer<T: v2::OutputPin> {
+//!     _pin: T,
+//! }
+//! 
+//! impl <T>NewOutputPinConsumer<T> 
+//! where T: v2::OutputPin {
+//!     pub fn new(pin: T) -> NewOutputPinConsumer<T> {
+//!         NewOutputPinConsumer{ _pin: pin }
+//!     }
+//! }
```
dbrgn: Same thing here with the `new` function.

src/digital/v1_compat.rs
ryankurte: i think i prefer the `new` method as it shows how you'd usually use it (with the `where` clause), am okay to change it if you disagree strongly.

dbrgn:
Again, bors build fails because we deny warnings.

I'll create a PR to remove that line.

ryankurte:
bors retry

PR #145, opened 2019-07-03
Silence warning for `impl Write` in rust-nightly.

bjc:
Nightly as of, at least `0beb2ba16 2019-07-02` requires the `dyn` keyword on the Write trait impl.


ryankurte:
thanks for the PR! this appears to have already landed in an separate formatting pass.

PR #150, opened 2019-08-20
Fix error types

Disasm:
`core::convert::Infallible` was stabilized in 1.34.0. This PR replaces `Void` and `()` error types with `Infallible`.

Note that this is a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,6 @@ readme = "README.md"
 repository = "https://github.com/japaric/embedded-hal"
 version = "0.2.3"
 
-[dependencies.void]
-default-features = false
-version = "1.0.2"
-
 [dependencies.nb]
 version = "0.1.1"
 
--- a/src/digital/v2.rs
+++ b/src/digital/v2.rs
@@ -59,7 +59,7 @@ pub trait ToggleableOutputPin {
 /// ```
 /// use embedded_hal::digital::v2::{OutputPin, StatefulOutputPin, ToggleableOutputPin};
 /// use embedded_hal::digital::v2::toggleable;
-/// use core::convert::Infallible;
+/// use std::convert::Infallible;
 ///
 /// /// A virtual output pin that exists purely in software
 /// struct MyPin {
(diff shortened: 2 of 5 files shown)


thejpster:
What's our stated MSRV?

ryankurte:
@thejpster huh, looks like we haven't [documented one](https://github.com/rust-embedded/wg/blob/master/ops/msrv.md) in this repo yet?

a few thoughts without particular direction:
- i wonder what the migration path for the `void` package is, it looks like there was a suggestion they move to '!' [some time ago](https://github.com/reem/rust-void/issues/13)
- it wouldn't let void be removed completely, but, might be possible instead to add a second shim with the `Infallible` type to `digital::v2_convert` so older implementations would continue to work (until void or aliases change anyway)
- Infallible looks to be a [temporary](https://doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility) measure prior to stabilisation of `!`, does it make sense to switch now or later?

Disasm:
@ryankurte As for me, `Infallible` is a nice replacement for `Void`: it's pretty the same and doesn't require an additional crate.
Anyway, my major concern was `()` error type in `v2_compat`: it doesn't provide a hint to the compiler that error condition never happens.

no111u3:
I merge it to my fork of embedded-hal - https://github.com/no111u3/embedded-hal/pull/1

thejpster:
Given our policy:

> The MSRV shall always be strictly smaller than the latest available minor stable release. For example, if the latest stable Rust version was 1.32.1 a new minor release of the cortex-m crate can't bump the MSRV higher than 1.31.

And that this only requires 1.34, I'm happy to merge this. Infallible will alias ! when it stabilises in 1.40, so we can take a PR to move to ! after 1.41 is out.

thejpster:
@rust-embedded/hal any objections to merging?

therealprof:
Let's do this.

therealprof:
Ah, the joy of requiring a funtioning nightly build.

Disasm:
Cargo.lock is quite outdated, retrying
bors r=therealprof

ryankurte:
PR `void` to assist with interoperability while we're migrating things: https://github.com/reem/rust-void/pull/20

Issue #154, opened 2019-10-08
I2c WriteIter use Item = &u8


MabezDev:
I've been playing with an I2c EEPROM chip recently in which writes are performed in one I2c transaction, i.e:

```
START -> 16bit address -> data -> STOP
```

I figured I would use write iter to chain the address bytes and data together, like so:

```rust
pub fn write(&mut self, address: u16, buffer: &[u8]) -> Result<(), <I as WriteIter>::Error> {
    let addr_bytes = [
        ((address & 0xFF00) >> 8) as u8,
        (address & 0x00FF) as u8,
    ];
    self.i2c.write(self.dev_addr, addr_bytes.iter().chain(buffer.iter()))
}
```

Which is really clean - unfortunately because `Item = u8` to do this we have to consume the buffer. I think changin the `WriteIter` signature to
```rust
/// Blocking write (iterator version)
#[cfg(feature = "unproven")]
pub trait WriteIter {
    /// Error type
    type Error;

    /// Sends bytes to slave with address `addr`
    ///
    /// # I2C Events (contract)
    ///
    /// Same as `Write`
    fn write<'a, B>(&mut self, addr: u8, bytes: B) -> Result<(), Self::Error>
    where
        B: IntoIterator<Item = &'a u8>;
}
```
makes more sense, and also makes the iterator version more similar to the simple write version.

eldruin:
Great idea! I think then I could get rid of all the copying for page writes in the [eeprom24x](https://crates.io/crates/eeprom24x) driver.

PR #155, opened 2019-10-08
Change iterator Item of I2c iter traits to use &u8

MabezDev:
Discussion issue: #154

### DIFF
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -66,9 +66,9 @@ pub trait WriteIter {
     /// # I2C Events (contract)
     ///
     /// Same as `Write`
-    fn write<B>(&mut self, addr: u8, bytes: B) -> Result<(), Self::Error>
+    fn write<'a, B>(&mut self, addr: u8, bytes: B) -> Result<(), Self::Error>
     where
-        B: IntoIterator<Item = u8>;
+        B: IntoIterator<Item = &'a u8>;
 }
 
 /// Blocking write + read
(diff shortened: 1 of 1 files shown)


ryankurte:
hey thanks for the PR! this seems like a reasonable improvement, however, appears to be a breaking change, so will likely require some further discussion from @rust-embedded/hal

MabezDev:
> Would it be possible to structure this as an addition instead of a change? i.e. add a new WriteIterRef and WriteIterRefRead (names possibly terrible) which operate in this way?

While it would be possible, I'd be interested in the use case of consuming the iterator on write. 

I would also like to propose changing name to `write_iter` not only because it better describes what it is, but because when implementing all these traits in a hal, having multiple methods named write causes ambiguity, if both traits were imported in a users program, the following would fail to compile.

```
use hal::blocking::i2c::{WriteIter, Write}

let buffer = [0xFF; 10];
i2c.write(0xAE, &buffer).unwrap(); // fails because &buffer implements into_iter and is &[u8]
```

I may be wrong here, but given the fact that these traits are behind the `unproven` feature, it would seem breaking changes are anticipated.

jonas-schievink:
I'm not sure I understand the motivation for this. If you have an iterator that yields `&u8` you can just call `.copied()` on it to get one that yields `u8`.

MabezDev:
I wasn't aware that was possible, in which case this PR pointless.

jrmoulton:
How does using `.copied` fix the issue in #154?

I'm running into this very same problem when trying to write to an I2C EEPROM and needing to copy all of the data into a new buffer to have the address pre-pended. 

@eldruin I looked at the implementation in eeporm24x and it looks like that is still copying all of the data. 

Am I missing something?

MabezDev:
@jrmoulton With the embedded-hal 1.0-rc's the iterative API has been removed, so for now the only way to do this is to have the address and data in a singular buffer.

PR #158, opened 2019-10-31
Minor changes in doc

dkm:
Fix one typo and add few words around await! and futures making more clear that both may not work on many targets.


no111u3:
It's a good changes for me. Tests was fixed and you could be try rerun it.

PR #159, opened 2019-11-02
Update package state to buildable

no111u3:
I update package to buildable on everything platforms


thejpster:
Hi! Thanks for the PR. I'm afraid I don't really understand your opening comment, and this seems like a duplicate of https://github.com/rust-embedded/embedded-hal/pull/150 so I'd like to get that one in first. Then we can pick up the update to the stm32 crate.

no111u3:
@thejpster Sorry, my mistake. I reupdate my pull request. I include #150 because it remove unneeded dependecies from project.

PR #160, opened 2019-11-02
Update dev dependencies to actual version

no111u3:
I clean update developing dependencies to buildable state.


therealprof:
Actually `stm32f30x` is really old and not actively maintained anymore. This should probably use `stm32f3` instead.

no111u3:
@therealprof, I think yes. I update it in next commit

therealprof:
Thanks. Can you squash the commits please?

no111u3:
I did it.

PR #162, opened 2019-11-03
Update to readiness for 2018 edition

no111u3:


src/fmt.rs
```diff
@@ -3,7 +3,7 @@
 //! TODO write example of usage
 use core::fmt::{Result, Write};
 
-impl<Word, Error> Write for ::serial::Write<Word, Error=Error>
+impl<Word, Error> Write for dyn (::serial::Write<Word, Error=Error>)
```
luojia65: Is it okay if `impl<Word, Error, T> Write for T where T: ::serial::Write<Word, Error=Error>` (take another proper name for T)?

no111u3: It's ok in new version of compillers - because it replace old trait implements.

Disasm:
Ping from triage: @thejpster, could you review this? I believe this `dyn` construction was introduced more than two stables ago (in fact, over a year ago).

thejpster:
Needs a rebase after https://github.com/rust-embedded/embedded-hal/pull/167 landed.

no111u3:
> Needs a rebase after #167 landed.

I close this pr because #167 landed - 2018 edition already applied. We don't need to update to becoming 2018 edition.

PR #164, opened 2019-11-14
Make the digital::{v1, v2}::InputPin traits proven

nils-van-zuijlen:
See #41 

Previous attempt : #102

### DIFF
--- a/src/digital/v2_compat.rs
+++ b/src/digital/v2_compat.rs
@@ -76,9 +76,6 @@ where
 impl<T> v2::toggleable::Default for T where T: v1::toggleable::Default {}
 
 /// Implementation of fallible `v2::InputPin` for `v1::InputPin` digital traits
-///
-/// Available behind `"unproven"` feature because `v1::InputPin` is
-#[cfg(feature = "unproven")]
 #[allow(deprecated)]
 impl <T> v2::InputPin for T
 where
@@ -178,13 +175,11 @@ mod tests {
         assert_eq!(o.state, false);
     }
 
-    #[cfg(feature = "unproven")]
     #[allow(deprecated)]
     struct OldInputPinImpl { 
         state: bool
     }
 
-    #[cfg(feature = "unproven")]
     #[allow(deprecated)]
     impl v1::InputPin for OldInputPinImpl {
         fn is_low(&self) -> bool {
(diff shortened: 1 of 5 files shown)


nils-van-zuijlen:
@therealprof I don't understand what you said. Do you mean that the `v1` trait should be proven also?

I did not remove the `unproven` feature flag there because to me it does not make sense to prove a deprecated trait.

And keeping the `v1` trait behind a feature flag forces the `compat` traits to be behind it.

src/digital/v2_compat.rs
```diff
@@ -115,6 +117,7 @@ mod tests {
         }
     }
 
+    #[cfg(feature = "unproven")]
```
therealprof: You added this.

src/digital/v2_compat.rs
```diff
@@ -140,10 +144,12 @@ mod tests {
...
```
therealprof: ... and this

src/digital/v2_compat.rs
```diff
@@ -140,10 +144,12 @@ mod tests {
...
     struct NewToggleablePinConsumer<T: v2::ToggleableOutputPin> {
         _pin: T,
     }
 
+    #[cfg(feature = "unproven")]
```
therealprof: ... and this

src/digital/v2_compat.rs
```diff
@@ -153,6 +159,7 @@ mod tests {
         }
     }
 
+    #[cfg(feature = "unproven")]
```
therealprof: ... and this

src/digital/v2_compat.rs
nils-van-zuijlen: `v1::StatefulOutputPin` is `unproven`, so this test wouldn't build without the flag anyway.

src/digital/v2_compat.rs
nils-van-zuijlen: `v2::ToggleableOutputPin` is `unproven`, so this test wouldn't build without the flag anyway.

src/digital/v2_compat.rs
nils-van-zuijlen: `v2::ToggleableOutputPin` is `unproven`, and `NewToggleablePinConsumer` depends on it, so this test wouldn't build without the flag anyway.

src/digital/v2_compat.rs
nils-van-zuijlen: same as https://github.com/rust-embedded/embedded-hal/pull/164#discussion_r347004908

nils-van-zuijlen:
I added these `unproven` feature flags to make tests work even when that flag is not enabled.

burrbull:
> I added these `unproven` feature flags to make tests work even when that flag is not enabled.

Remove them.

PR #166, opened 2019-11-18
Update code formatting style to common

no111u3:
Update code formatting style to common rust format style.
It helps everyone to create pr without manual check coding style of project.


thejpster:
Could you add something to the Travis file which checks the formatting? Something like `cargo fmt -- --check` should do it.

no111u3:
> Could you add something to the Travis file which checks the formatting? Something like `cargo fmt -- --check` should do it.

@therealprof I update PR with adding checks the formatting on CI

thejpster:
Thanks @no111u3

PR #167, opened 2019-11-18
Change library edition to 2018

no111u3:
Crate edition updated to 2018. Any fixes are applied by `cargo fix` utility. It includes same part with #162 because in 2018 this warnings are errors.
Code formatting will be introduced in another PR(such as #166)


therealprof:
Uhm, haven't noticed but this won't work due to the use of the await keyword. Guess we need to shuffle the nb stuff around if we want to use edition 2018.

therealprof:
@no111u3 Seems there's still a conflict in `src/prelude.rs`

no111u3:
@therealprof sorry, my mistakes. I update for test.

thejpster:
bors retry

PR #171, opened 2019-11-26
Switch async code to use futures crate

no111u3:
Remove using `nb` crate for asynchronously run code - use `core::futures` for replace it.
RFC for these changes is #172

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,15 +14,12 @@ readme = "README.md"
 repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.3"
 
-[dependencies.nb]
-version = "0.1.1"
-
 [dev-dependencies]
 stm32f3 = { version = "0.8", features = ["stm32f303", "rt"] }
 futures = "0.1.17"
 
 [features]
-unproven = ["nb/unstable"]
+unproven = []
 
 [package.metadata.docs.rs]
 features = ["unproven"]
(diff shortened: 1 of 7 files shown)


thejpster:
Is there a reason this is using futures 0.1 instead of 0.3?

no111u3:
@thejpster I fixed build on last commit. It is clean compatible changes with older code - only remove `nb` but need to switch minor version (I don't create RFC because it's may be applied anytime, It's try to make code clean). For reason of prefer 0.1 - it was my try to reproduce old PR #13, now I change it to core implementation without external crates. Thanks.

korken89:
Interesting, I had no idea `Poll` had been stabilized.
We could probably replace the `nb` crate now.

therealprof:
As per 11-26 meeting this will require an RFC to move ahead.

Nemo157:
There are semantics attached to `Poll::Pending` that did not exist on `nb::Error::WouldBlock`, specifically:

> When a function returns `Pending`, the function _must_ also ensure that the current task is scheduled to be awoken when progress can be made.

Having functions that return `core::task::Poll` but do not take a `core::task::Context` and arrange to wake the task when necessary is likely to lead to deadlocks if they are used by systems built around `core::future::Future`.

no111u3:
@therealprof I added RFC on head message of this thread.

no111u3:
@Nemo157 I understand that problem we have but in current implementation of `api` i couldn't change behavior of run without more breakable changes on `api`. I'm planning to find solution in this situation. Thanks.

no111u3:
Need to review of architecture solution.

PR #173, opened 2020-01-04
Add CapturePin trait

YuhanLiin:
The `CapturePin` trait is the input capture analogue to the `PwmPin` trait, representing a single capture channel. Unlike `PwmPin`, there's no `enable` or `disable` methods, because some chips, such as the MSP430FR2355, may have trouble implementing those two methods on capture abstractions. The proposed trait has been implemented in [msp430fr2x5x-hal](https://github.com/YuhanLiin/msp430fr2x5x-hal/blob/master/src/capture.rs#L300).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -795,6 +795,26 @@ pub trait Capture {
         R: Into<Self::Time>;
 }
 
+/// A single input capture channel / pin
+///
+/// See `Capture` for details
+#[cfg(feature = "unproven")]
+pub trait CapturePin {
+    /// Type  of value returned by capture
+    type Capture;
+    /// Enumeration of `Capture` errors
+    ///
+    /// Possible errors:
+    ///
+    /// - *overcapture*, the previous capture value was overwritten because it
+    ///   was not read in a timely manner
+    type Error;
+
+    /// "Waits" for a transition in the capture `channel` and returns the value
+    /// of counter at that instant
+    fn capture(&mut self) -> nb::Result<Self::Capture, Self::Error>;
+}
+
 /// Pulse Width Modulation
 ///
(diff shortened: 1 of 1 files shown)


YuhanLiin:
The proposal for the trait is [here](https://github.com/rust-embedded/embedded-hal/issues/176), which I added after reading that I need to complete a proposal before opening a PR to add a new trait.

src/lib.rs
```diff
@@ -795,6 +795,26 @@ pub trait Capture {
...
+    ///
+    /// Possible errors:
+    ///
+    /// - *overcapture*, the previous capture value was overwritten because it
+    ///   was not read in a timely manner
+    type Error;
+
+    /// "Waits" for a transition in the capture `channel` and returns the value
+    /// of counter at that instant
+    fn capture(&mut self) -> nb::Result<Self::Capture, Self::Error>;
```
ryankurte: Thanks for the PR, I would prefer not to have two methods called `capture`, could this one be `capture_pin` or something to delineate between the two?

YuhanLiin: I don't believe having both `Capture.capture()` and `CapturePin.capture()` causes any issues. Presently, `Pwm` and `PwmPin` methods all share the same names, such as `enable`, `disable`, and `get_duty`. The reason why this is OK is because `Pwm` and `PwmPin` will never be implemented on the same type, since one represents a collection of PWM channels while the other represents a single PWM channel. As such, trait method name conflicts will never occur. `Capture` and `CapturePin` follow the same reasoning, ensuring that the two `capture()` methods will never conflict.

YuhanLiin:
Any updates on this?

eldruin:
I would say let's reevaluate this after the 1.0 release.

PR #174, opened 2020-01-06
Release 0.2.4 semver-trick for embedded-hal v0.3.x release

ryankurte:
- bump cargo version
- semver trick against embedded-hal v0.3.x

PR #175, opened 2020-01-06
[proposal] release 0.3.0

ryankurte:
Swaps digital trait default to `digital::v2` (breaking change), as the next step in the planned `digital::v1` [deprecation](https://github.com/rust-embedded/embedded-hal/issues/95), as an alternative to https://github.com/rust-embedded/wg/pull/393.

This approach is aligned with all other HAL traits (where `Infallible` is returned for infallible implementations), minimises complexity (by not introducing multiple options that must be effectively communicated to all users, or the opportunity for incompatibility due to use of the _incorrect_ trait) with the end goal of removing the version numbers and having a single set of fallible `digital::*` traits.

I had hoped to port the compatibility shims back to the v0.2 series, however, this seems impossible to do without cyclic dependencies.

## Dependencies
- embedded-hal release 0.2.4 https://github.com/rust-embedded/embedded-hal/pull/174 containing the semver-trick to maintain backwards compatibility

## Impact
- HAL consumers using `digital::*` traits directly will need to update to fallible types
- HAL consumers using `digital::v1::*` or `digital::v2::*` will observe no change

## Next steps
- HAL v0.4.x series should remove `digital::{v1, v1_compat, v2_compat}` to complete the migration


ryankurte:
Closing as replaced by #192

Issue #94, opened 2018-09-09
Allow multiple read/write calls per transaction via SPI/I2C


RandomInsano:
Both SPI and I2C have some concept of a transaction. SPI you assert an enable pin before sending data, I2C has `start` and `end` messages.

I'm finding that as I write and review more crates, the ergonomics of requiring single fixed sized buffers for communication are becoming cumbersome or require a lot of code duplication for variably large receivers. An example of the latter is the required code duplication in the [eeprom24x](https://github.com/eldruin/eeprom24x-rs/blob/master/src/lib.rs#L421).

[After some research](https://www.kernel.org/doc/Documentation/i2c/dev-interface), it looks like the buffer copy for SPI and I2C in Linux-land might be unavoidable because the `I2C_RDWR` ioctl support isn't required. That means it would need to be wrapped somehow in the [linux-embedded-hal](https://github.com/rust-embedded/linux-embedded-hal) crate.

Looking at [Arduino](https://www.arduino.cc/en/Tutorial/DigitalPotentiometer), they do have the concept of protocol transactions. I'm not sure if I agree with the API entirely, but the concept is good. I'm sure Rust opens up some interesting design improvements there.

Can someone poke holes in this idea for me? I'd like some sort of solution to allocating these extra buffers on memory constrained devices, especially when it leads to reading and writing different sized data.

Here's a list of code that I find where buffer copying was required but didn't need to be:
* [eeprom24x](https://github.com/eldruin/eeprom24x-rs/blob/master/src/lib.rs#L421): For reading / writing arbitrarily large pages
* [pscontroller](https://github.com/RandomInsano/pscontroller-rs/blob/master/src/lib.rs#L310): For changing one byte in a constant buffer

RandomInsano:
Just some quick notes here, it looks like following what Linux does with I2C_RDWR might be a great way to go if we're okay implementing a fallback:

```text
ioctl(file, I2C_RDWR, struct i2c_rdwr_ioctl_data *msgset)
  Do combined read/write transaction without stop in between.
  Only valid if the adapter has I2C_FUNC_I2C.  The argument is
  a pointer to a

  struct i2c_rdwr_ioctl_data {
      struct i2c_msg *msgs;  /* ptr to array of simple messages */
      int nmsgs;             /* number of messages to exchange */
  }

  The msgs[] themselves contain further pointers into data buffers.
  The function will write or read data to or from that buffers depending
[...]

Issue #52, opened 2018-03-02
[Discussion] `Read` and `Ignore` traits for blocking SPI operations


austinglaser:
Currently, there are the `blocking::spi::Transfer` and `blocking::spi::Write` traits. However, there times when one might want to read data out of a slave, without caring about what gets transmitted -- a `blocking::spi::Read`

This could be done with the `blocking::spi::Transfer` trait, but a separate trait provides two advantages:

- No need to allocate a dummy buffer to pass in as `words`
- An implementer of `blocking::spi::Read` could allow the MOSI pin to be optional

Going even further, there are times when a driver simply requires some clock cycles on the bus -- but will not be shifting meaningful data either in or out. This may seem like a corner case, but I've encountered this exact situation when configuring an ice40 FPGA. For reference, see page 19 of [the programming and configuration manual](www.latticesemi.com/-/media/LatticeSemi/Documents/ApplicationNotes/IK/iCE40ProgrammingandConfiguration.ashx?la=en). It's also a requirement for some commodity SPI flash devices.

This could be served by a `blocking::spi::Ignore` trait, which just takes a number of bytes to ignore.

ryankurte:
This was discussed in https://github.com/rust-embedded/embedded-hal/pull/178#discussion_r365008818 (note that this was closed in favour of #191 with only the SPI component), with the conclusion that `Ignore` is equivalent to `Write` and `Read` is equivalent to `Transfer` with a zeroed array which is preferable as it matches the physical implementation on the device.

Unless there is something else outstanding I believe that resolves this issue?

PR #178, opened 2020-01-09
Add transactional traits for SPI and I2C

ryankurte:
Related to #94, PR for review / discussion

Demonstrated in:
- https://github.com/rust-embedded/linux-embedded-hal/pull/33
- https://github.com/ryankurte/rust-embedded-spi/pull/3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.3"
 
 [dependencies.nb]
-version = "0.1.2"
+version = "0.1.1"
 
 [dev-dependencies]
 stm32f3 = { version = "0.8", features = ["stm32f303", "rt"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.3"
 
 [dependencies.nb]
-version = "0.1.1"
+version = "0.1.2"
 
 [dev-dependencies]
 stm32f3 = { version = "0.8", features = ["stm32f303", "rt"] }
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -127,3 +127,28 @@ pub trait WriteIterRead {
     where
         B: IntoIterator<Item = u8>;
 }
+
+/// Actions for transactional I2C trait
+/// 
+/// This allows composition of I2C operations into a single bus transaction
+#[cfg(feature = "unproven")]
+#[derive(Debug, PartialEq)]
+pub enum Actions<'a> {
+    /// Read data into the provided buffer, write data undefined
+    Read(&'a mut [u8]),
+    /// Write data from the provided buffer, discarding read data
+    Write(&'a [u8]),
+}
+
+/// Transactional trait allows multiple actions to be executed
+/// as part of a single I2C transaction
+#[cfg(feature = "unproven")]
+pub trait Transactional {
+    /// Associated error type
+    type Error;
+    
+    /// Execute the provided actions
+    fn exec(&mut self, addr: u16, actions: &[Actions]) -> Result<(), Self::Error>;
+}
+
+
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -148,7 +148,5 @@ pub trait Transactional {
     type Error;
 
     /// Execute the provided actions against the provided I2C address
-    fn exec<'a, O>(&mut self, address: u8, operations: O) -> Result<(), Self::Error>
-    where
-        O: AsMut<[Operation<'a>]>;
+    fn exec<'a>(&mut self, address: u8, operations: &mut[Operation<'a>]) -> Result<(), Self::Error>;
 }
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -129,7 +129,7 @@ pub trait WriteIterRead {
 }
 
 /// Actions for transactional I2C trait
-/// 
+///
 /// This allows composition of I2C operations into a single bus transaction
 #[cfg(feature = "unproven")]
 #[derive(Debug, PartialEq)]
@@ -146,9 +146,7 @@ pub enum Actions<'a> {
 pub trait Transactional {
     /// Associated error type
     type Error;
-    
+
     /// Execute the provided actions against the provided I2C address
     fn exec(&mut self, addr: u8, actions: &[Actions]) -> Result<(), Self::Error>;
 }
-
-
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -148,5 +148,5 @@ pub trait Transactional {
     type Error;
 
     /// Execute the provided actions against the provided I2C address
-    fn exec(&mut self, addr: u8, actions: &[Actions]) -> Result<(), Self::Error>;
+    fn exec(&mut self, addr: u8, actions: &mut [Actions]) -> Result<(), Self::Error>;
 }
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -128,12 +128,12 @@ pub trait WriteIterRead {
         B: IntoIterator<Item = u8>;
 }
 
-/// Transaction for transactional I2C trait
+/// Operation for transactional I2C trait
 ///
 /// This allows composition of I2C operations into a single bus transaction
 #[cfg(feature = "unproven")]
 #[derive(Debug, PartialEq)]
-pub enum Transaction<'a> {
+pub enum Operation<'a> {
     /// Read data into the provided buffer
     Read(&'a mut [u8]),
     /// Write data from the provided buffer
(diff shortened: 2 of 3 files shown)


src/blocking/spi.rs
```diff
@@ -104,3 +104,31 @@ pub mod write_iter {
...
+
+
+/// Actions for transactional SPI trait
+/// 
+/// This allows composition of SPI operations into a single bus transaction
+#[cfg(feature = "unproven")]
+#[derive(Debug, PartialEq)]
+pub enum Actions<'a> {
+    /// Read data into the provided buffer, write data undefined
+    Read(&'a mut [u8]),
```
austinglaser: Is there a concrete advantage to having `Read` in addition to `Transfer`? In both cases, you have to create and initialize a mutable buffer -- at least if you're calling from safe code. But I could see "write data undefined" being a low-caliber footgun. Maybe you _do_ care about the outgoing bytes, but the HAL implementation you're initially testing with happens to do the right thing.

What about `Read(u8, &'a mut [u8])`, where the first byte is clocked out repeatedly for the duration of the transaction?

src/blocking/spi.rs
```diff
@@ -104,3 +104,31 @@ pub mod write_iter {
...
+    /// Write data from the provided buffer, discarding read data
+    Write(&'a [u8]),
+    /// Transfer data from the provided buffer, overwriting the buffer with read data
+    Transfer(&'a mut [u8]),
+}
```
austinglaser: What about an `Ignore(u8, usize)`, which clocks out a number of identical bytes and discards the read data?

I've worked with FPGAs (specifically: the ICE40 series) which require some dummy cycles at certain points in their initialization sequence. While this _could_ be done with `Write`, that would require allocating an output buffer when not truly necessary.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.3"
 
 [dependencies.nb]
-version = "0.1.1"
+version = "0.1.2"
```
eldruin: nitpick: this is a separate issue.

src/blocking/spi.rs
ryankurte: > Is there a concrete advantage to having Read in addition to Transfer? 

Some devices support read operations without transferring data out to the peripheral.

> I could see "write data undefined" being a low-caliber footgun

Yep, unfortunately when _only_ reading neither [spidev](https://github.com/spotify/linux/blob/master/include/linux/spi/spidev.h) or the [CP2130](https://www.silabs.com/documents/public/application-notes/AN792.pdf) device I've been playing with support configuring what data is shifted out.

As it's hardware dependent, I don't think this is in scope for the hal. Would a more explicit warning on the `Read` operation mitigate this concern?

src/blocking/spi.rs
ryankurte: I think this is probably out of scope for this crate, requiring all hal implementations to add an `Ignore` method (which is not something actually supported by hardware afaik) seems like more trouble than allocating a buffer if you need this functionality.

src/blocking/spi.rs
austinglaser: > Alternately Write and Transfer would cover most use cases, so, Read could just be dropped.

I would be in favor of that

src/blocking/spi.rs
austinglaser: That's a fair point

src/blocking/i2c.rs
```diff
@@ -127,3 +127,26 @@ pub trait WriteIterRead {
...
+
+/// Transactional trait allows multiple actions to be executed
+/// as part of a single I2C transaction
+#[cfg(feature = "unproven")]
+pub trait Transactional {
+    /// Associated error type
+    type Error;
+
+    /// Execute the provided actions against the provided I2C address
+    fn exec(&mut self, addr: u8, actions: &[Actions]) -> Result<(), Self::Error>;
```
ryankurte: do we like the name `exec` here, or would something else be better?

eldruin: I think it is fine, except someone else has a better idea.

src/blocking/i2c.rs
```diff
@@ -127,3 +127,26 @@ pub trait WriteIterRead {
     where
         B: IntoIterator<Item = u8>;
 }
+
+/// Actions for transactional I2C trait
+///
+/// This allows composition of I2C operations into a single bus transaction
+#[cfg(feature = "unproven")]
+#[derive(Debug, PartialEq)]
+pub enum Actions<'a> {
```
jonas-schievink: Enums usually have a singular name, so `Action`. Also, why not `Operation`?

ryankurte: Yeah really good point, I'm not sure why I didn't just use `Transaction` tbh, does that work for you?

jonas-schievink: Sounds good!

src/blocking/spi.rs
ryankurte: relates to https://github.com/rust-embedded/embedded-hal/issues/52

src/blocking/spi.rs
```diff
@@ -104,3 +104,28 @@ pub mod write_iter {
         }
     }
 }
+
+/// Operation for transactional SPI trait
+///
+/// This allows composition of SPI operations into a single bus transaction
+#[cfg(feature = "unproven")]
+#[derive(Debug, PartialEq)]
+pub enum Operation<'a, W: 'static> {
```
ryankurte: is the static bound here okay? I can't think of a case it's _not_, but, could use `'a` instead if it was preferred

jonas-schievink: Why use `'static` here?

This should ideally use a `Word` bound, but `Word` doesn't currently exist which makes it effectively impossible to portably use non-`u8` word sizes since there are no universally agreed-upon types for that.

ryankurte: > Why use 'static here?

i was trying to be clear that it's a different bound on the type `W` vs. the references to the data in `Operation`, `'a` is also _correct_, but i couldn't think of a case where the `W` type wouldn't be static?

this is why the question ^_^

ryankurte: the alternative also foists the lifetime bound to the trait, which is okay, but it's also kind of nice to have it in the function declaration (`fn exec<'a, O: AsMut<[Operation<'a, W>]>>(&mut self, operations: O) -> Result<(), Self::Error>`)

src/blocking/spi.rs
```diff
@@ -104,3 +104,28 @@ pub mod write_iter {
...
+    /// Write data from the provided buffer, discarding read data
+    Write(&'a [W]),
+    /// Write data from one buffer while reading into the second buffer
+    WriteRead(&'a [W], &'a mut [W]),
```
ryankurte: WriteRead taking two arrays (as is consistent with the linux `spidev`, though one is also allowable) makes it difficult to map between `WriteRead` objects and the `Transfer` function, it's probably more useful to run a single array here instead...

eldruin: By `Transfer` you mean `exec` or something else?

src/blocking/spi.rs
eldruin: In my limited experience, lifetime bounds are less problematic when attached to functions rather than traits as their scope is smaller but we probably will not add further functions to the trait which would need a different lifetime binding so it probably does not make much of a difference.

src/blocking/spi.rs
ryankurte: the [existing](https://github.com/rust-embedded/embedded-hal/blob/371251e03142f969d671b28f563efa695c5bc64a/src/blocking/spi.rs#L4) `spi::Transfer` trait. I have now swapped this to match (because otherwise polyfilling `Transactional` requires allocation).

ryankurte: right, a side effect of this is that it becomes nearly impossible to wrap the linux spidev transactions :cry:, so, possibility of allocation it is i guess?

see [here](https://github.com/rust-embedded/linux-embedded-hal/pull/33/files#diff-b4aea3e418ccdb71239b96952d9cddb6R307):
``` rust
// Map types from generic to linux objects
let mut messages: Vec<_> = operations.as_mut().iter_mut().map(|a| {
  match a {
    SpiOperation::Write(w) => SpidevTransfer::write(w),
    SpiOperation::WriteRead(w, r) => SpidevTransfer::read_write(w, r) ,
  }
}).collect();
```

ryankurte: It's possible to hack around this with:
``` rust
let w = unsafe {
  let p = r.as_ptr();
  std::slice::from_raw_parts(p, r.len())
};

SpidevTransfer::read_write(w, r)
```

however, it's not super clear from the [docs](https://www.kernel.org/doc/html/v4.14/driver-api/spi.html#c.spi_transfer) whether sharing tx_buffer and rx_buffer is okay

eldruin: Hmm... it does not look very nice. For the Linux kernel [implementation](https://github.com/torvalds/linux/blob/aae1464f46a2403565f75717438118691d31ccf1/drivers/spi/spidev.c#L198) I think it would work, as the buffers are copied in and out.

ryankurte:
Closing out in favour of #191 for now

Issue #182, opened 2020-01-26
Pwm enable and disable methods only work for channels


justacec:
In a review of the Pwm trait, I noticed that there is no way to disable a timer completely.  The existing `enable` and `disable` methods are designed for specific channels.  I think there ought to be some method to generally disable the entire timer.  I think that the current methods should be re-purposed as the generate timer disable/enable and then two new methods could be introduced to support channel enable/disable.  These new methods could be called `enable_channel`/`disable_channel`.  The remaining methods would not need to be altered as they are understood by context (ie, one cannot specify the duty for all of the channels at once, it is understood that is a per-channel operation).

PR #183, opened 2020-01-26
Add methods to enable/disable the entire timer

justacec:
I am submitting this simple PR to start discussion on the recently opened issue #182.  I have simply renamed the existing `enable` and `disable` methods to be `channel_enable` and `channel_disable`.  I have also added two new methods, `enable`/`disable`, which would then be expected to be used to enable/disable the entire timer.

I think that this functionality is useful so that one does not need to temporarily store the current duty for each channel in order to "disable" the timer.  The state would be maintained and it results in a single read/write to the timer control registers.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -862,11 +862,17 @@ pub trait Pwm {
     /// (e.g. `0.0 .. 1.0`) or an integer representation (e.g. `0 .. 65535`)
     type Duty;
 
+    /// Disables a PWM timer (all channels)
+    fn disable(&mut self);
+
+    /// Enables a PWM timer (all channels)
+    fn enable(&mut self);
+
     /// Disables a PWM `channel`
-    fn disable(&mut self, channel: Self::Channel);
+    fn channel_disable(&mut self, channel: Self::Channel);
 
     /// Enables a PWM `channel`
-    fn enable(&mut self, channel: Self::Channel);
+    fn channel_enable(&mut self, channel: Self::Channel);
 
     /// Returns the current PWM period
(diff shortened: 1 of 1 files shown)


ryankurte:
cc. @rust-embedded/hal, if this change is desirable we should break this in the v1 migration?

therealprof:
I think the desire to have this came up in https://github.com/stm32-rs/stm32f1xx-hal/pull/176 .

therealprof:
@justacec Would you mind fixing the conflicts?

therealprof:
Ping @justacec.

therealprof:
@justacec Are you still around?

justacec:
@therealprof Working on it now.

justacec:
@therealprof Ok, I think I synced this up.  As I was updating this PR, I was not sure if the other "channel" specific methods (such as `try_get_duty`) should be changed to `try_get_duty_channel` when called from the Pwm object.  That name makes sense for the PwmPin, but maybe not intuitive enough for the Pwm object.

therealprof:
Looks good to me. Any objections @ryankurte?

ryankurte:
Looks good to me! Even though it is rather straightforward I think it would still be good to demonstrate an implementation / driver, as well as to add this change to the changelog.

justacec:
@ryankurte I just updated the changelog as requested.  As for the implementation, I suppose that would need to be done in the `stm32f1xx_hal` crate.

ryankurte:
has anyone had a chance to implement this? would be excellent to get it merged ^_^

Issue #177, opened 2020-01-07
Ship a 1.0 release


thejpster:
Part of https://github.com/rust-embedded/wg/issues/383

Settled blockers:

* ~#169~ (closed for later consideration)
* #163
* ~#172~
* ~[[RFC] digital::v3 interface](https://github.com/rust-embedded/wg/pull/393) + moving to a single interface~
* #234 
* #98 PR: #222 
* #140
* #147 PR: #230
* #156
* #110
* #244 PR: #246
* #298  
* ~#249~
* #229 PR: #296 
* ~#182 PR: #183~
* #287
* #321
* #351 
* #365 
* Alternative SPI word lengths: #320
* ~#340~
* #398 
* ~#387~ 
* #394 
* #392
* #432 
* #440 
* #443
* #445
* #367 
* ~https://github.com/rust-embedded/linux-embedded-hal/issues/87~ discussed in WG meetings, we decided we're OK with the solutions in https://github.com/rust-embedded/linux-embedded-hal/issues/87#issuecomment-1267641344
* #349 
* #480
* #283 PR: #284 
* #482 
* #483 
* ~#539~ docs in #559, will add CS delays to `embedded-hal-bus` later. 
* #541 
* #552
* #455 

Open blockers:

* #547
* #553 ?
* #555
* Show that MCU and linux implementations work fine
* Show that drivers can work fine with the new changes
* Your issue here!

Feel free to nominate issues also by commenting below

jamesmunns:
Here is a [link to the video call] discussion had by myself and most of the members of @rust-embedded/hal team. This includes initial decisions made, and justification/reasoning for each point. The next steps for this would be to codify this into an RFC, and get approval for it (though there is already general consensus within the embedded-hal team).

[link to the video call]: https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary

by the way, this was the "issue with the list of big topics" we were looking for during the meeting :)

eldruin:
I believe the only unresolved issue as of the list above is https://github.com/rust-embedded/embedded-hal/issues/172
I would also add the dependency on https://github.com/japaric/nb/issues/20
Then we need to think about the dev-dependencies which are at the moment `stm32f3` and `futures`. I guess specially these will hold us back for a while.

Disasm:
[...]

PR #105, opened 2018-10-16
Make digital traits fallible

eldruin:
This makes the digital traits fallible as discussed in #97, #95, #100 and #41 implemented as a clean break.

PR #192, opened 2020-03-11
[towards 1.0]: Fallible proven traits

eldruin:
Here some work towards 1.0 following https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-597682592.


Disasm:
We agreed on `try_` prefix on fallible methods to prevent name collision.

eldruin:
Oops, I misread that. Here some more work towards what was discussed and thus 1.0.
- Rename all methods `try_*` and make them fallible as appropriate.
- Use `core` instead of `std`.
- Remove `unproven` feature

I did not rename the adc::Channel::channel() method. This method exists at all due to https://github.com/rust-lang/rust/issues/54973, which is fixed as of Rust 1.35.
What do we want to do about this?
- Keep as is.
- Rename method.
- Update MSRV to 1.35 and make the method a constant.

thejpster:
Personally I'm OK with upping the MSRV to 1.35. That is now 7 versions (or 42 weeks) behind the current stable.

src/blocking/delay.rs
```diff
@@ -12,15 +12,21 @@
 /// `UXX` denotes the range type of the delay time. `UXX` can be `u8`, `u16`, etc. A single type can
 /// implement this trait for different types of `UXX`.
 pub trait DelayMs<UXX> {
+    /// Enumeration of `DelayMs` errors
+    type Error;
+
     /// Pauses execution for `ms` milliseconds
-    fn delay_ms(&mut self, ms: UXX);
+    fn try_delay_ms(&mut self, ms: UXX) -> Result<(), Self::Error>;
```
thejpster: So, this is interesting. Can a busy-wait delay ever actually fail? If not, should this remain an infallible API?

src/timer.rs
```diff
@@ -1,13 +1,12 @@
 //! Timers
 
-use core::convert::Infallible;
 use nb;
 
 /// A count down timer
 ///
 /// # Contract
 ///
-/// - `self.start(count); block!(self.wait());` MUST block for AT LEAST the time specified by
```
thejpster: Like with the busy-wait functions, can a timer ever actually fail? Should this be infallible? I'm less convinced than with the busy-waits, but I could still go either way on this.

src/blocking/delay.rs
eldruin: Good point. I also wondered about these traits. However, blocking behavior is not the same as busy-waiting.

src/timer.rs
eldruin: For completeness I repeat here: I also wondered about these traits. However, blocking behavior is not the same as a busy-wait implementation.

Disasm:
I think that adding `try_` prefix only applies to traits that are already fallible. I'm not sure if we should convert all infallible traits into fallible ones as this can be done later by adding another fallible trait.

eldruin:
> I think that adding `try_` prefix only applies to traits that are already fallible. I'm not sure if we should convert all infallible traits into fallible ones as this can be done later by adding another fallible trait.

Hmm, this would mean we would be back to `PwmPin` vs. `FalliblePwmPin`, which was extensively discussed and discarded for the `OutputPin` case (#95, #97, #100, #105).

Here a relevant excerpt from [the meeting protocol](https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary):
> 3. For the next major release, we will ONLY be offering Fallible traits, e.g. traits that return Results. All trait method names will be prefixed with try_ for consistency, e.g. try_send.

src/blocking/delay.rs
thejpster: Could you elaborate on that last statement?

eldruin: Sure.
I understand blocking behavior as the fact that the further execution of the caller's code is stopped until the operation that was called has finished. i.e. the contract to the caller.
(In contrast to e.g. returning `Error::WouldBlock`).
I wanted to highlight that in the case of the delay this does not necessarily mean busy-waiting. For example, one could also think of an implementation à la cooperative multitasking which would still be an implementation which holds this contract (of course errors in this case need to be reported to the appropriate caller, it would be completely wrong to make the delay function fallible in order to return errors from another one, that is _not_ what I am saying).
The point being that maybe a different implementation which I do not know of exists for which it may make sense to fail.

Going back to my main statement, I made these methods fallible as well because the [meeting notes](https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary) explicitly stated:
> 3. For the next major release, we will ONLY be offering _Fallible_ traits, e.g. traits that return `Result`s. All trait method names will be prefixed with `try_` for consistency, e.g. `try_send`.

Nevertheless, I also wondered if failing would actually be possible for a delay implementation or if this was decided only for consistency (I was not at the meeting, maybe I should join some time).
It would be interesting to know if someone can think of a failing delay.

thejpster: > It would be interesting to know if someone can think of a failing delay.
I've given it some thought and although I can't think of one, I can't rule it out. This this is the correct approach.

thejpster:
LGTM. Any reason we can't merge this?

Cargo.toml
```diff
@@ -14,15 +14,10 @@ readme = "README.md"
 repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.3"
 
-[dependencies.nb]
-version = "0.1.1"
+[dependencies]
+nb = { version = "0.1.1", features = ["unstable"] }
```
therealprof: `unstable` ? That has a weird sound to it considering that we're aiming to stabilize this.

eldruin: This is because currently the `unproven` feature requires `nb/unstable`. The state of the `unstable` feature should be discussed for `nb`'s 1.0 release.

ryankurte:
Sorry to raise this a bit late, but would it make sense for us to batch all the breaking change PRs into a separate branch so that we can release bug fixes (for example #194) prior to the next major release? Adds to workload in keeping the branch up to date but avoids this stalling us if this takes a little while.

thejpster:
I'm a big fan of git flow for release management, but that didn't go down well in a recent meeting. Shall we put something about rules for branch management on the agenda for the next meeting?

thejpster:
For now if you need to make a hotfix, I suggest you create a branch from the release tag.

therealprof:
> Sorry to raise this a bit late, but would it make sense for us to batch all the breaking change PRs into a separate branch so that we can release bug fixes (for example #194) prior to the next major release? Adds to workload in keeping the branch up to date but avoids this stalling us if this takes a little while.

I agree we should try to keep the individual PRs as compact and self-contained as possible to allow for cherry-picking if necessary for e.g. a bugfix release.

> I'm a big fan of git flow for release management, but that didn't go down well in a recent meeting.

Well, it would be sufficient if the reviewer insisted on broken down PRs before accepting a wholesale crate PR. 😅However in this particular case we're good because the author took good care and split the change into plenty of individual commits.

PR #194, opened 2020-03-14
InputPin instead of OutputPin for compatibility shim

AdminXVII:
The OldInputPin required OutputPin instead of InputPin for the new impl. Fix the trait bound.

### DIFF
--- a/src/digital/v1_compat.rs
+++ b/src/digital/v1_compat.rs
@@ -113,7 +113,7 @@ pub struct OldInputPin<T> {
 
 impl<T, E> OldInputPin<T>
 where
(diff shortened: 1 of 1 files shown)


ryankurte:
Thanks for opening this! We're doing a big refactor of master at the moment which is going to remove the previous pin versions, but should probably backport the fix.

I had a got at just swapping branches, but, this didn't work because it needs a rebase. If you'd like to land this could you open the same PR with just this change against the `v0.2.x` branch, and add this to the changelog?

AdminXVII:
Done in https://github.com/rust-embedded/embedded-hal/pull/199.

Closing this in favor of the other MR.

PR #195, opened 2020-03-16
Separate Capture, Pwm and Qei traits into modules

eldruin:
The `lib.rs` file is already _very_ long. I propose separating these traits which are just at the end of the file into their own modules.
I have kept the modules private and reexported the traits and types to maintain the same interface.
However, for consistency with the other traits maybe it would be worth it to expose these new modules as well and not reexport the traits in `lib.rs`.
What do you think?


ryankurte:
Seems like a good idea to me!
> However, for consistency with the other traits maybe it would be worth it to expose these new modules as well and not reexport the traits in lib.rs.

As part of the breaking changes happening in #192 I think we may as well fix this for consistency.

eldruin:
@ryankurte done.

eldruin:
@ryankurte sorry there were some errors in the documentation. They should be solved now.
On macOS `cross test --target=arm-unknown-linux-musleabihf` does not run the doc tests. Any idea why?

eldruin:
Is anything preventing this being merged?

ryankurte:
thanks for the reminder, looks good to me (though we're going to need to do a lot of changelog updating)

bors r+

PR #198, opened 2020-04-06
Add MSRV configuration to CI

eldruin:
The compatibility with the MSRV should be checked by the CI. At the moment that is Rust 1.35.0 due to [this issue](https://github.com/rust-lang/rust/issues/54973).
The MSRV and its update process should probably also be documented in the README.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,11 @@ matrix:
       rust: nightly
       if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
+    # Minimum supported Rust version
+    - env: TARGET=x86_64-unknown-linux-gnu
+      rust: 1.35.0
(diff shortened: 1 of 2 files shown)


ryankurte:
thanks for the PR! seems like a good addition to me, can we also add a note to the readme as per [here](https://github.com/rust-embedded/wg/blob/master/ops/msrv.md)?

eldruin:
@ryankurte Sure. It is done now.
What do you think of a rustc compatibility badge (also a documentation badge and so on) like [here](https://crates.io/crates/syn)?

eldruin:
Ok, badges will be discussed in https://github.com/rust-embedded/wg/pull/443.
We can add that in a separate PR after the discussion has settled.
Is there anything else preventing the merge of this?

PR #199, opened 2020-04-07
InputPin instead of OutputPin for compatibility shim

AdminXVII:
The OldInputPin required OutputPin instead of InputPin for the new impl. Fix the trait bound.


ryankurte:
bors retry

ryankurte:
Failed due to `dyn` keyword requirement that is / was fixed in 961dfb87687bf0fd07b0b126db6d0083242323ca, i need to update v0.2.x to the latest non-breaking change (rather than a mirror of 0.2.3)

PR #202, opened 2020-04-13
Add triagebot configuration

LeSeulArtichaut:
This enables [assignment](https://github.com/rust-lang/triagebot/wiki/Assignment) through triagebot on this repository, in preparation for the migration from highfive to triagebot for PR assignment.

cc rust-lang/highfive#258 rust-lang/triagebot#433


therealprof:
Is there a newer nightly that includes the rustfmt component?

ryankurte:
I don't think we actually [specify](https://github.com/rust-embedded/embedded-hal/blob/cad3b0c19098bb89159890e40391db72950c67db/.travis.yml#L17) a nightly version? (in which case, this seems to me it should be an allowed failure?)

therealprof:
Yeah it should. However it seems there's fixed nightly build so now we can also r+ all the other repos. ;)

Issue #203, opened 2020-04-16
I2S interface


maxekman:
First some background: I'm planning to use embedded Rust in a sound-oriented project and am currently experimenting with the STM32 F4 Discovery which has an onboard DAC with amp and headphone jack. The DAC connects to the MCU with I2C for control commands (volume etc) and I2S for the audio data. Currently I have just started to implement a embedded-hal driver for the DAC, a Cirrus Logic CS43L22, beginning with setting up the device and I2C control channel. In the end I will use a more capable DAC with 4-8 channels, but it will still use I2S for audio data.

Is there a plan to add a I2S interface in the embedde-hal project yet? If not, how would I go about contributing that to embedded-hal? I have read some guidelines where it says to create a new interface in an external crate before it's stable, any example of that process so far?

The I2S standard uses mostly the same internal components and control registers as a SPI interface, but with a bit more complex clock setup to my understanding.

ryankurte:
hi, sounds like an interesting project!

> Is there a plan to add a I2S interface in the embedded-hal project yet?

As far as I know there is not yet.

> I have read some guidelines where it says to create a new interface in an external crate before it's stable, any example of that process so far?

You can either create a new crate with new traits, or fork embedded-hal and add traits (and use cargo patches to test), we've just switched to this approach so don't have a great example of the former yet, though #191 shows the latter.

eldruin:
Interesting project. I also have several I2S devices (including a CS4344 :) I could write a driver for but I always focused on other projects.
To start the discussion I have created a PR at #204.

maxekman:
Thanks for all the input! I’m fairly new to Rust and will follow along the proposal and contribute what I can to the process.

eldruin:
I published I2S traits in a separate [`embedded-i2s`](https://github.com/eldruin/embedded-i2s-rs) crate.
I would encourage anybody to continue the discussion over there.
A discussion about integration into `embedded-hal` would come much later.

KOEGlike:
Any updates?

PR #204, opened 2020-04-17
[Proposal] I2S traits

eldruin:
I added some I2S traits to kick off the discussion started in #203.
The traits accept two words for left and right and are otherwise similar to SPI.

An interleaved version can be found in the [`i2s-interleaved`](https://github.com/eldruin/embedded-hal/tree/i2s-interleaved) branch. The data should be passed interleaved. e.g. `[left0,right0,left1,right1,...]`

A version of these traits for the embedded-hal 0.2.x version can be found in the [`i2s-v0.2.x`](https://github.com/eldruin/embedded-hal/tree/i2s-v0.2.x) branch.

There is another MCU implementation in [`stm32f4xx-hal`](https://github.com/stm32-rs/stm32f4xx-hal/pull/265)

TODO:
- [X] Demo microcontroller HAL implementation:
    - [X] [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal/pull/99) - [discussion issue](https://github.com/stm32-rs/stm32h7xx-hal/issues/90)
    - [ ] [[WIP] stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/pull/145)
- [x] [Bit-banging HAL implementation supporting 16-bit and 32-bit packets](https://github.com/eldruin/bitbang-hal/blob/i2s-v0.2.x/src/i2s.rs)
    - [x] [Working example tested using both MAX98357A and PCM5102A](https://github.com/eldruin/driver-examples/blob/i2s-bitbang-bp/stm32f1-bluepill/examples/i2s-audio-bitbang.rs)
- [X] [Working example of saw wave](https://github.com/maxekman/stm32f407g-disc/blob/dac-support-with-cs43l22/examples/audio.rs)
    - Uses driver: https://github.com/maxekman/cs43l22
- [X] Full-duplex trait implementation
    - [X] [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal/pull/99) - [discussion issue](https://github.com/stm32-rs/stm32h7xx-hal/issues/90)
- [x] Changelog entry.
- [ ] Remove `try_` prefix
- [ ] Move files according to #278 after that is merged

Thoughts?
cc. @maxekman

### DIFF
--- a/src/blocking/i2s.rs
+++ b/src/blocking/i2s.rs
@@ -1,0 +1,39 @@
+//! Blocking I2S API
+
+/// Blocking read
+pub trait Read<W> {
+    /// Error type
+    type Error;
+
+    /// Reads enough bytes from the slave to fill `left_words` and `right_words`.
+    fn try_read<'w>(
+        &mut self,
+        left_words: &'w mut [W],
+        right_words: &'w mut [W],
+    ) -> Result<(), Self::Error>;
+}
+
+/// Blocking write
+pub trait Write<W> {
+    /// Error type
+    type Error;
+
+    /// Sends `left_words` and `right_words` to the slave.
+    fn try_write<'w>(
+        &mut self,
+        left_words: &'w [W],
+        right_words: &'w [W],
+    ) -> Result<(), Self::Error>;
+}
+
+/// Blocking write (iterator version)
+pub trait WriteIter<W> {
+    /// Error type
+    type Error;
+
+    /// Sends `left_words` and `right_words` to the slave.
+    fn try_write<LW, RW>(&mut self, left_words: LW, right_words: RW) -> Result<(), Self::Error>
+    where
+        LW: IntoIterator<Item = W>,
+        RW: IntoIterator<Item = W>;
+}
--- a/src/blocking/i2s.rs
+++ b/src/blocking/i2s.rs
@@ -32,7 +32,11 @@ pub trait WriteIter<W> {
     type Error;
 
     /// Sends `left_words` and `right_words` to the slave.
-    fn try_write<LW, RW>(&mut self, left_words: LW, right_words: RW) -> Result<(), Self::Error>
+    fn try_write_iter<LW, RW>(
+        &mut self,
+        left_words: LW,
+        right_words: RW,
+    ) -> Result<(), Self::Error>
     where
         LW: IntoIterator<Item = W>,
         RW: IntoIterator<Item = W>;
--- a/src/i2s.rs
+++ b/src/i2s.rs
@@ -1,0 +1,17 @@
+//! I2S - Inter-IC Sound Interface
+
+use nb;
+
+/// Full duplex
+pub trait FullDuplex<Word> {
+    /// Error type
+    type Error;
+
+    /// Reads the left word and right word available.
+    ///
+    /// The order is in the result is `(left_word, right_word)`
+    fn try_read(&mut self) -> nb::Result<(Word, Word), Self::Error>;
+
+    /// Sends a left word and a right word to the slave.
+    fn try_send(&mut self, left_word: Word, right_word: Word) -> nb::Result<(), Self::Error>;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- I2S (Inter-IC Sound) traits.
 
 ### Changed
 
--- a/src/blocking/mod.rs
+++ b/src/blocking/mod.rs
@@ -6,6 +6,7 @@
 
 pub mod delay;
 pub mod i2c;
+pub mod i2s;
 pub mod rng;
 pub mod serial;
 pub mod spi;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -412,6 +412,7 @@ pub mod blocking;
 pub mod capture;
 pub mod digital;
 pub mod fmt;
+pub mod i2s;
 pub mod prelude;
 pub mod pwm;
 pub mod qei;
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -28,6 +28,7 @@ pub use crate::digital::InputPin as _embedded_hal_digital_InputPin;
 pub use crate::digital::OutputPin as _embedded_hal_digital_OutputPin;
 pub use crate::digital::StatefulOutputPin as _embedded_hal_digital_StatefulOutputPin;
 pub use crate::digital::ToggleableOutputPin as _embedded_hal_digital_ToggleableOutputPin;
+pub use crate::i2s::FullDuplex as _embedded_hal_i2s_FullDuplex;
 pub use crate::pwm::Pwm as _embedded_hal_Pwm;
 pub use crate::pwm::PwmPin as _embedded_hal_PwmPin;
 pub use crate::qei::Qei as _embedded_hal_Qei;
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -14,6 +14,9 @@ pub use crate::blocking::i2c::{
     WriteIterRead as _embedded_hal_blocking_i2c_WriteIterRead,
     WriteRead as _embedded_hal_blocking_i2c_WriteRead,
 };
+pub use crate::blocking::i2s::{
+    Read as _embedded_hal_blocking_i2s_Read, Write as _embedded_hal_blocking_i2s_Write,
+};
 pub use crate::blocking::rng::Read as _embedded_hal_blocking_rng_Read;
 pub use crate::blocking::serial::Write as _embedded_hal_blocking_serial_Write;
 pub use crate::blocking::spi::{


maxekman:
Thanks @eldruin for opening the proposal. I also got contacted by @jkristell about drivers for the CS43L22 (which uses I2S) which he also had been experimenting with. Maybe @jkristell also has some input to the discussion?

src/blocking/i2s.rs
```diff
@@ -0,0 +1,39 @@
...
+    ) -> Result<(), Self::Error>;
+}
+
+/// Blocking write (iterator version)
+pub trait WriteIter<W> {
+    /// Error type
+    type Error;
+
+    /// Sends `left_words` and `right_words` to the slave.
+    fn try_write<LW, RW>(&mut self, left_words: LW, right_words: RW) -> Result<(), Self::Error>
```
therealprof, line 35: Do we need `LW` and `RW` separate for some reason?

maxekman: For PCM mode it’s not needed at all, for the other modes it’s not technically needed but could be semantically useful. For the other modes first left then right is written to the data register, or for Philips mode it’s synced to the WS clock. But I’m very new to this and have little practical knowledge. I guess the final decision depends on what abstraction level is the right one for the embedded-hal lib. The more high level left/right channel semantics (or even multichannel) can definitely be left to individual drivers to supply IMO.

eldruin: Separate `LW` and `RW` would allow different types for the buffers as long as they can be converted to a `W` iterator. I would rather not require them to be the same type if it is not necessary.

maxekman: Just out of curiosity, when would it make sense to have different types for the left and right channel? To my understanding they would always be interleaved when written and an application would most likely write both a left and right word at the same time.

ryankurte: > Separate LW and RW would allow different types for the buffers as long as they can be converted to a W iterator.

Is there a situation in which this is useful? from my _extremely limited_ i2s experience the left and right channels are always interleaved and usually going to be run at the same sampling rate / bit depth?

Afaik we usually have separate methods for iterator based approaches, rather than trying to be generic over both. For example, something like `fn try_write<W>(&mut self, left_words: &[W], right_words: &[W]) -> Result<(), Self::Error>` vs. `fn try_write_iter<W, I: Iterator<W>>(&mut self, left_words: I, right_words: I)`

maxekman:
I would like to bring up another topic; clock control.

I know that setting up a device etc is outside the scope of the embedded-hal project, but I would like to gather the opinions on clock setup for I2S. Is there a way it can be done in the scope of embedded-hal?

The clock setup plays such a major role for I2S often involving both a master clock for external ADCs and a word clock for the channels in addition to the ordinary serial clock. Any other thoughts on this topic?

src/blocking/i2s.rs
maxekman: I would prefer a single interleaved buffer and would be perfectly happy with having a higher level abstraction (with separate left/right etc) in the driver crate instead. That way stereo and multichannel DACs can be supported with better semantics.

eldruin: I thought separate types would improve support for 1-channel "mono" communication. e.g. maybe one wants to have a struct that generates numbers on the fly for the left channel and just loop over a zero-filled buffer for the right channel.

However, I just had a look at the multichannel description (TDM, although it is not standard) and it seems anything higher level than a raw buffer would not be a good fit.
I will update this PR soon.

maxekman: Sounds good. The new interface looks much cleaner BTW.
I'll try to play with an implementation of it in the stm32f4xx-hal crate I get some time, mostly for me to learn the ecosystem better.

maxekman:
BTW, I have started work on the CS43L22 driver, which could be an example/reference implementation. I'll hopefully get it into the stm32f407g-disc crate eventually.

maxekman:
I have now also started to experiment with a I2S HAL implementation for the stm32f4xx-hal crate (https://github.com/stm32-rs/stm32f4xx-hal).

maxekman:
I have working sound output from the F4 discovery board now! The next step will do the same thing but via this HAL proposal and see how that feels. Will feedback any discoveries on the way.

src/blocking/i2s.rs
maxekman: Now when I have some more knowledge with I2S the separate left and right words might not be such a bad idea after all. What supports this is the fact that a full frame is considered transmitted only after both left and right has been sent. The word size can be 16, 24 or 32 bits. In the STM32F4 there is a special status flag (CHSIDE) which can be used together with TXE to determine which word to send next. It would be good to be able to take advantage of that.

I have also read the data sheet for the CS4385 which is a 8 channel DAC. In TDM mode it can receive all 8 channels over a SAI interface, which is basically a more modular version of I2S. In that case it can send 256 bits after only one WS trigger, but that behavior is not possible with standard I2S. My conclusion is that it’s perfectly ok for the HAL to focus on providing a 2 channel abstraction, which is all that standard I2S will be able to do.

Could you revert back the last change @eldruin? Sorry for the extra rounds 🙂

eldruin: @maxekman Very interesting :) Sure, no problem. I just reverted the last change in the i2s branch. The interleaved version can be found in the `i2s-interleaved` branch.

maxekman: Thanks @eldruin, will give it a go later today.

maxekman:
I have published an early draft of a MCU HAL implementation: https://github.com/stm32-rs/stm32f4xx-hal/pull/145

If you want you can update the todo in the description @eldruin.

maxekman:
I have also published a WIP of a driver implementation for the CS43L22 DAC found on the STM32F4 Discovery board: https://github.com/maxekman/cs43l22

@eldruin You can add this to the description also if you want.

maxekman:
Here is a running example of a simple frequency sweep of a inverted saw wave: https://github.com/maxekman/stm32f407g-disc/blob/dac-support-with-cs43l22/examples/audio.rs

eldruin:
Cool stuff @maxekman! I updated the description accordingly.

eldruin:
I backported this PR to embedded-hal 0.2.x for testing ease in the [`i2s-v0.2.x`](https://github.com/eldruin/embedded-hal/tree/i2s-v0.2.x) branch.
I wrote an initial bit-banging implementation of the I2S traits [here](https://github.com/eldruin/bitbang-hal/blob/i2s-v0.2.x/src/i2s.rs).
Next I will test it on hardware and then iterate adding further capabilities.

maxekman:
@eldruin That is cool, do you have some DAC chip to test with?

maxekman:
@eldruin Do you think I should also implement the v2.x version of the HAL for stm32f4xx-hal? I'm not familiar with the timelines for when the current master of embedded-hal will be released. I did hear something that it may be released as v1..

eldruin:
Yes, the current `embedded-hal` master state will be released as v1.0 but I am not sure when that will happen (see #177). Maybe it would make sense to release this in a 0.2.x release to get more testing done before 1.0.
What do you think @therealprof?

I have a couple I2S DACs but had no chance to try them so far. I want to test my implementation with a MAX98357A which uses standard I2S (the only mode I implemented so far). If you want give it a try as well that would be great :)

maxekman:
@eldruin Sounds good, I'll work on the v2.x HAL implementation first then, in any case.

eldruin:
I tested my bitbanging implementation on a bluepill and it works fine with both MAX98357A and PCM5102A using 16-bit audio. See [here](https://github.com/eldruin/driver-examples/blob/i2s-bitbang-bp/stm32f1-bluepill/examples/i2s-audio-bitbang.rs)
I started a 24-bit implementation as well but did not finish it yet.
It would be good to see an implementation of the full-duplex trait. Otherwise we could also remove it for now to move this forward.

RaoulHC:
Is it just the stm32f4xx-hal implementation that's needed for this to go in? I've been playing around with audio on the stm32f407g-disc too, so it would be good to see this get merged and not use a selection of half finished branches. Happy to help that go in if you're no longer working on it @maxekman, as I see there hasn't been movement in a while.

maxekman:
@RaoulHC I would love to finish this, I might be able to do it today. I have indeed not worked on this in a while due to the day job. If I don't get anything up soon, feel free to help out.

maxekman:
@eldruin Could you rebase and push your i2s-v0.2.x branch (and possibly the i2s also)? Thanks!

eldruin:
@maxekman I have rebased both branches now. FYI, I think it would be best to release this as part of `embedded-hal` version 1.1.0 to avoid getting anything else on the way of 1.0.0. Version 1.1.0 could be quickly released after 1.0 and if this is ready by then, I believe it could be merged.

maxekman:
@eldruin Are you suggesting that I drop the work towards v0.2.x and completely focuses on v1.x?

eldruin:
> @eldruin Are you suggesting that I drop the work towards v0.2.x and completely focuses on v1.x?

That is not what I meant. I just wanted to let you know that this does not need to be done before the 1.0 release.
Regarding the `i2s-v0.2.x` version, I only added it to ease testing and development of I2S support since using the `i2s` branch required changing everything to be 1.0-compatible first and that was a lot of work.
As of now, the `embedded-hal` `v0.2.x` branch is only receiving some maintenance fixes and I2S should only be merged to the master (1.x) branch after the 1.0 release.
Does this clarify?

maxekman:
Yep, thanks for the clarification @eldruin!

maxekman:
Got some progress on the ref implementation / POC today but could not get any sensible audio to play using the trait version (the non-trait version `send` works and plays the saw tooth) from the example in stm32f407g-disco (https://github.com/stm32-rs/stm32f407g-disc/pull/16). Must be something with the generics that is not done correctly (I'm pretty new to Rust and the type system). Here is the trait implementation for `try_write`, any advice is welcome: https://github.com/stm32-rs/stm32f4xx-hal/pull/212/files#diff-d277d6e47b64bf2a37ad0427748af56dR999

src/i2s.rs
```diff
@@ -0,0 +1,17 @@
...
+    /// Error type
+    type Error;
+
+    /// Reads the left word and right word available.
+    ///
+    /// The order is in the result is `(left_word, right_word)`
+    fn try_read(&mut self) -> nb::Result<(Word, Word), Self::Error>;
+
+    /// Sends a left word and a right word to the slave.
+    fn try_send(&mut self, left_word: Word, right_word: Word) -> nb::Result<(), Self::Error>;
```
luojia65, line 16: Should this be try_write? There's a `read` so a `write` is maybe okay here

eldruin: I named it similarly to `spi::FullDuplex`, where it is also called try_read/[try_send](https://github.com/rust-embedded/embedded-hal/blob/6f172879c7f2a82d97f28f7631b19902f748435d/src/spi.rs#L26)

maxekman:
Maybe this is ready to be merged soon? I have not been able to finish my reference implementation but @samcrow did finish a similar implementation in https://github.com/stm32-rs/stm32f4xx-hal/pull/265.

eldruin:
@maxekman cool!
What would be left would be implementing the traits in this PR on top of https://github.com/stm32-rs/stm32f4xx-hal/pull/265. I think it should be fairly trivial but maybe some details come up then.

YruamaLairba:
Hi, i just discovered this I2S proposal. samcrow mentionned it because i'm rewriting I2s implementation for stm32f4xx. Is the discussion still open ? I have remarks, comments and suggestions.

The first is the following: i see many times the terms "slave" to refer to the other I2s device. I think referring in this way is an error, "master" and "slave" just define who provide clocks, and the "master" is not even necessarily one of the I2s interface implied in data communication.

Issue #172, opened 2019-11-27
[RFC] Use core language async primitives instead external crates


no111u3:
# Summary
This proposal proposes to change the way how `async api` does implemented. It makes `async api` more clearly for use and extend.

# Motivation
Our current approach to implement trait primitives as depends on `nb` crate. It was actually on start of this project but after version `1.36` rust already have `async traits` and methodology for work with it. So in 2018 edition of rust lang we have language words as `async`, `await`. The marcosses exported by `nb` are conflicted with key words, we need to use `r#` prefix for each macros use.

This RFC attempts to change implementation to `core types` based implementation. It will be more effective and more conform rust language ideas.

# Detailed design
This RFC inspired of PR #13, but after apply external crate I redesign for use core functionality because it will be better than external depends.

For example we have interface for serial:
```rust
use core::task::Poll;

/// A serial interface
pub trait Serial {
    /// Error type associated to this serial interface
    type Error;
    /// Reads a single byte
    fn read(&mut self) -> Poll<Result<u8, Self::Error>>;
    /// Writes a single byte
    fn write(&mut self, byte: u8) -> Poll<Result<(), Self::Error>>;
}
```
It use `core::task::Poll` as point of `async` run status. It was easy for implement and it isn't depends on `nb` or another external crates.

# Alternatives
Don't implement this RFC.

# Uresolved questions:
 - More clear examples of right use `async` code.
 - How we need to update depend crates?
 - Do we need to rewrite `nb` prefer than rewrite `async` api of `embedded-hal`?

Nemo157:
There are semantics attached to `Poll::Pending` that did not exist on `nb::Error::WouldBlock`, specifically:

> When a function returns `Pending`, the function _must_ also ensure that the current task is scheduled to be awoken when progress can be made.

Having functions that return `core::task::Poll` but do not take a `core::task::Context` and arrange to wake the task when necessary is likely to lead to deadlocks if they are used by systems built around `core::future::Future`.

Nemo157:
> @Nemo157 I understand that problem we have but in current implementation of api i couldn't change behavior of run without more breakable changes on api. I'm planning to find solution in this situation. Thanks.

[...]

PR #206, opened 2020-04-23
Add asynchronous versions of most HAL traits

dflemstr:
This attempts to take a stab at addressing #172 by implementing asynchronous versions of most HAL traits, using the language-native async/await support.  The thinking here is that it's better to have *some* suggestion for what an async HAL could look like, as a starting point for discussion.

The code was mostly taken from https://github.com/dflemstr/embedded-platform where I've had a chance to battle test it for a while, but the code should not be considered polished by any means.

Any device should be able to support async/await now that Rust no longer relies on TLS for scheduling.  Even if a device lacks fancy capabilities like a heap, you can always use an interrupt-driven scheduling model using my https://github.com/dflemstr/direct-executor executor crate for example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,12 @@ repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.3"
 
 [dependencies]
+futures = { version = "0.3.4", optional = true }
 nb = { version = "0.1.1", features = ["unstable"] }
 
 [dev-dependencies]
 stm32f3 = { version = "0.8", features = ["stm32f303", "rt"] }
 futures = "0.1.17"
 
+[features]
+asynchronous = ["futures"]
--- a/src/asynchronous/gpio.rs
+++ b/src/asynchronous/gpio.rs
@@ -1,0 +1,201 @@
+//! General input/output pins.
+//!
+//! The [`InputPin`] and [`OutputPin`] traits define pins that can be read and written digitally
+//! (i.e. either in a low or high state).
+//!
+//! There are additionally various `Into*` traits that allow users to re-configure pins to switch
+//! between different modes of operation, e.g. [`IntoFloatingInputPin`] turns a pin into an
+//! [`InputPin`] that does not employ any pull-up or pull-down resistors.
+use core::pin;
+use core::task;
+
+pub mod get;
+pub mod set;
+
+/// A generic pin that can't be interacted with.
+pub trait Pin {
+    /// The common error type for all pin operations.
+    ///
+    /// A single error type for all operations is enforced for simplicity.
+    type Error;
+}
+
+/// A pin that can be read from.
+pub trait InputPin: Pin {
+    /// Polls a read operation of this pin to completion.
+    fn poll_get(
+        self: pin::Pin<&mut Self>,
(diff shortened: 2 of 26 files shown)


Disasm:
Alternative: https://github.com/Disasm/async-trait-poc/tree/master/embedded-async-sandbox (requires unstable `generic_associated_types` feature)

dflemstr:
Any updates or feedback on this? @ryankurte @thejpster @ilya-epifanov :)

ryankurte:
hi @dflemstr, thanks for the PR (and the reminder)!

it seems like an interesting approach, however, i think there's both too much ongoing discussion around how we're going to achieve it (#172 as you've seen) and that this is too large of a changeset to land without more discussion / documentation. 

i'm not _entirely_ sure what the path forward from here is, will try and arrange a chat with the hal team about how we plan to proceed, but, if you're interested in continuing with this (and per our updated hal guidelines) i would prefer to see a single trait (maybe just serial?) implemented in an async manner with both example implementations (i believe the `linux-embedded-hal` crate has async methods now) and consumers.

@Disasm also interesting! looks like the `generic_associated_types` feature fixes the cursed lifetime issues my attempt at this for my radio hal had (cc. @jamesmunns), and that this would also benefit from the switch from nb to poll as per #172?

also cc. @therealprof

therealprof:
@dflemstr Nice work! I have to agree with @ryankurte about having some example implementations to look and also the way forward is very much under discussion.

Since it would be nice to have a way forward and there's obviously some initiative, how about we turn this into a "community" project where interested people can collaborate and experiment; we might call this `async-embedded-hal`.

@jamesmunns Do we have pointers to our community efforts? Can we put this on the next meeting agenda?

dflemstr:
@ryankurte @therealprof thanks for getting back to me! I think that to get the ball rolling, we need to break out of this sort of catch-22, that I guess lots of "standards bodies" struggle with quite commonly :)

Ideally I think it would be nice to have a shared "playground" as part of embedded-hal (e.g. a feature flag that is considered outside of the public API for semver purposes, or a separate crate that is re-exported) so that different device implementations could try out the shared APIs easily.  I do think it makes sense to start with the entire API as a whole rather than start with serial and continue with SPI etc, since there's value in having them share characteristics (e.g. a common `Write` trait).

Given your suggestions above: We could start from the implementation end, and convince one of the device HAL maintainer to invest in this.  However, since there would be no embedded-hal "blessed" direction I fear there will be a risk that the implementation ends up too device-specific, or that there won't be any buy-in for creating async versions of the API to begin with.  See for example my work [here](https://github.com/nrf-rs/nrf-hal/pull/129) where I think we got quite far but I felt like I am unsure about what the next steps should be; there are a bunch of discussions I'd like to have with the entire embedded-hal community that don't really belong/can't be answered in that specific PR.

dflemstr:
@ryankurte @therealprof @jamesmunns I think I will need your help in creating a shared location for this work, either in this repo or else in some other repo (e.g. under the `rust-embedded` org).  What would be your preferences?

Dirbaio:
This can probably be closed, now that we have `embedded-hal-async`.

eldruin:
Thank you everybody for the discussion!

PR #209, opened 2020-05-13
Proposal for Configure baudrate trait

unizippro:
This trait lets the implementer choose the type for `Baudrate` as this is done in different ways in different crates.
Returning a result lets the implementer sort out bad, incompatible or impossible bauds, or simply return an unsupported error.
An implementation can be seen at [stm32l4xx-hal/feature/change-baud/](https://github.com/MathiasKoch/stm32l4xx-hal/blob/feature/change-baud/src/serial.rs#L814)
In this implementation the clock freq is saved with the `Uart` struct, and the used to change the baud. If the clock frequency were to change the Uart would have to be reinitialized anyway.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -25,3 +25,13 @@ pub trait Write<Word> {
     /// Ensures that none of the previously written words are still buffered
     fn try_flush(&mut self) -> nb::Result<(), Self::Error>;
 }
+
+/// Enable changing the baudrate after initiation of serial interface
+pub trait ConfigureBaud {
+    /// Baudrate type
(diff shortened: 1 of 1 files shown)


timokroeger:
Thank you for tackling this issue.
Could you please provide an example on how to use the new Trait in a generic driver?

unizippro:
I see now that this function should be able to be called from tx and rx. But since they don't hold any reference to registers I'll leave this as a draft for now.
A solution may be to return `(tx, rx, config)` from split, but i won't be investigating this for now.

Issue #21, opened 2017-09-24
CAN/CAN-FD API?


kjetilkjeka:
From working with the [uavcan rust implementation](https://github.com/UAVCAN/uavcan.rs) I see the convenience of having a standardized CanFrame and CAN interface in Rust. Perhaps this is something this crate should aim to have?

I think it would make sense to both have a ```trait Can``` and a ```struct CanFrame```.

robomancer-or:
+1 on this.  My one request would be to not differentiate between frames that have been received and frames that are intended to be transmitted (this seems to be what you're proposing, I'm just making it explicit).  ChibiOS does it this way and I find it creates a lot of duplicate code.  Is this feature only lacking a champion?  If so, I'd be glad to champion it.

therealprof:
Please mock up the interface you have in mind for discussion. A PoC implementation would be even better but is not necessarily required.

kjetilkjeka:
Here's my attempt https://docs.rs/embedded_types/0.3.2/embedded_types/can/index.html

i think a trait for sending these types is required as well.

therealprof:
Those types seem to be just the framing, those should probably live in their own crate anyway. What is needed for a hal interface are generic methods for setting up CAN transfers (and doing so) over the CAN peripherals of a typical MCU. I don't know particularly much about CAN transmissions but from reading the STM32F042 reference manual it pretty much seems to build down to:

- CAN configuration
- Switching between different modes
- Stuffing data into a mailbox
- Getting data out of a mailbox

Maybe there's more to it and/or other MCUs do it in a complete different way, don't know.

robomancer-or:
Yeah, CAN configuration is pretty straightforward.  I'm going to take my own stab at writing a proposal over the next day or two.  I do have a question before I dive in though,

 In one of @japaric 's hal implementations (F3xx I want to say?) he has his own implementation of units, but the uom crate seems to already be for this purpose.  Would it be preferable to roll my own representation of a frequency type, or would it be better to depend on uom?

dunmatt:
Quick preview for you guys:  https://github.com/dunmatt/embedded-hal/commit/24f492295d5856a21ed03542e77dc17f00d30575

It still needs documentation, obviously, but even without it I think it's pretty clear what's going on (for the most part).  One thing I wasn't sure of (again, very new to rust) was if I'm going the move correctly in CanInterface::receive... would it be better to use a mutable ref?

What do you all think?

[...]

PR #212, opened 2020-05-23
Controller Area Network (CAN) Take 3

timokroeger:
## Usage Example
[stm32-fwupdate](https://github.com/timokroeger/pcan-basic-rs/blob/master/pcan-basic/examples/stm32-fwupdate.rs)

## Prototype Implementations
* [pcan-basic](https://github.com/timokroeger/pcan-basic-rs) on top of an existing software API
* [stm32fxx-hal](https://github.com/timokroeger/stm32f1xx-hal/tree/can-embedded-hal) as part of a HAL crate

## Previous Discussion
* [WIP] CAN Interface take 2 #77
* CAN/CAN-FD API? #21
* added a HAL for CAN interfaces #53


josko7452:
Hey! I am just dropping by I am taking up Rust and I was thinking that I would do CAN layer for **embedded-hal** since I have quite some encounter with CAN on embedded, but I see I come a bit late.

But now looking at it I was thinking wouldn't it make sense to have some generic timing structure akin to what is in Linux kernel: 
https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/can/netlink.h#L31
namely structs `can_bittiming` and `can_bittiming_const`

The idea is that struct as `can_bittiming` is generic description of timing settings of a CAN hardware and `can_bittiming_const` defines the capability of given driver.
While I would not add implementation to calculate `struct can_bittiming` in rust-hal I think it makes sense to have trait that allows you to set generic timing and get the timing capabilities from CAN hardware.

If this is not done then setting timing will be a a bit bypass of HAL layer because the values are specific for each CAN IP core..

I saw that timing was discussed before, but is a bit different way of doing it.. its not implementing anything directly, only defining the interface common to both CAN driver and an application siting on top of **embedded-hal** this way the drivers can provide concise way to show timing capabilities while application layer can use those to set the timing.

timokroeger:
I added traits for receive filter configuration. There is quite some variation between CAN controllers:
* Not all filters have a mask
* A shared mask for multiple filters (MPC2515)
* Some controllers only support standard 11bit identifiers (STM32 bxCAN is configurable)
* Some do not include the RTR bit in the filter (Microchip MCP2515, Peak PCAN-Basic)
* Some have the RTR bit but no mask for it (Bosch C_CAN, Infineon MultiCAN)

I tried to cover all those configurations with an interface that exposes capabilities similar to the approach that @josko7452 described for timing. `filter_groups()` returns an iterator that describes the controllers filter capabilities.

@josko7452  Regarding the bit timings I’m not quite sure if they are in scope of `embedded_hal`. I’m borrowing this ASCII graphic from @pftbest:
````
    (1)                 (2)               (3)               (4)            (5)
+--------+         +----------+  +-------------------+  +--------+  +---------------+
|  Some  |   CAN   |   CAN    |  |       HAL         |  |  HAL   |  | Device Dirver |
| Device |<=======>| Hardware |<-|  implementation   |<-| Traits |<-|      for      |
|        |   BUS   | (in MCU) |  | (hardware driver) |  |        |  | "Some Device" |
+--------+         +----------+  +-------------------+  +--------+  +---------------+
                                           ^                                ^
                                           |                                |
                                           |                        +---------------+
                                           |                        |      User     |
                                           +------------------------|    Program    |
                                                                    |               |
                                                                    +---------------+
                                                                           (6)
````
Usually the system integrator configures the bit timings. I would consider bit timing configuration to be part of the init procedure in (6).
In contrast the HAL only provides traits for the device drivers (4), (5). So even if a generic device driver can query the timing capabilities how would it use that information? The only use case I can come up with: Baud-rate adjustments during runtime (similar to #79). Given that the CAN bus is usually used for reliable applications software based baud-rate adjustment looks rather fragile to me.

**Comment by josko7452** [comment]
> @josko7452 Regarding the bit timings I’m not quite sure if they are in scope of `embedded_hal`. I’m borrowing this ASCII graphic from @pftbest:

Yes I have seen that ACII graphics. My point is that I don't know whose scope that would be otherwise.
The problem is that as you probably know **CAN timing** parameters are given by CAN standards ([like this one](http://www.semiconductors.bosch.de/pdf/can2spec.pdf)) and therefore can be calculated generically disregarding the specific implementation of a given CAN IP core.

To me if you are an init procedure in (6) you would only like to pass something **generic** to the driver which all drivers understand rather than having to deal at (6) which register sets what.. 

That's the point of the driver to abstract the register accesses or am I wrong? And this is exactly what `struct can_bittiming` describes. Instead of dealing with registers these are values as specified in the **BOSH** document so they are common to _all_ CAN hardware.
`struct can_bittiming {
	__u32 bitrate;		/* Bit-rate in bits/second */
	__u32 sample_point;	/* Sample point in one-tenth of a percent */
	__u32 tq;		/* Time quanta (TQ) in nanoseconds */
	__u32 prop_seg;		/* Propagation segment in TQs */
	__u32 phase_seg1;	/* Phase buffer segment 1 in TQs */
	__u32 phase_seg2;	/* Phase buffer segment 2 in TQs */
	__u32 sjw;		/* Synchronisation jump width in TQs */
	__u32 brp;		/* Bit-rate prescaler */
};`
The driver can take this and translate it to IP core specific registers that set what this describes.
This is usually very thin layer of translation just subtracting 1 or adding 1 or summing some of these.

I am not saying there is need to include this calculation in **embedded-hal** this can be done elsewhere (even application layer, or pre-calculated constants), but what is needed is a concise **interface** to set the timing.

Also by these examples I am not saying we should copy what is in the **Linux kernel**.
But on the other hand there were several takes on CAN HW abstraction there so why to reinvent the wheel.

Just for completness this is how these structs are used:
https://elixir.bootlin.com/linux/latest/source/drivers/net/can/dev.c#L118 (this is the **generic** calculation of timing heuristic)
And this is how it's used in a driver:
https://elixir.bootlin.com/linux/latest/source/drivers/net/can/ti_hecc.c#L175
and
https://elixir.bootlin.com/linux/latest/source/drivers/net/can/ti_hecc.c#L265

As you can see it _doesn't add a lot of burden_ to the driver. What it does is that instead of having
**implementation specific** method to set bit timings we would end up with a **generic one**.

And then when you would port your application from one CAN HW to another you wouldn't need to replace your init function in (6) for the CAN HW, because it would still apply (be it constants or a function that calculates `struct can_bittiming` on the fly.

If there is another place where such a generic interface can be defined _for all_ CAN drivers then I welcome the suggestions.

**Comment by timokroeger** [comment]
Rebased to resolve the conflict in `Cargo.toml`.
Any comments on this?
Looking at this again with some distance I would consider dropping/postponing the second commit with filter traits to get a basic interface going.

File: src/can.rs [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+        !self.is_remote_frame()
+    }
+
+    /// Returns the frame identifier.
+    fn id(&self) -> u32;
+
+    /// Returns the data length code (DLC) which is in the range 0..8.
+    ///
+    /// For data frames the DLC value always matches the lenght of the data.
+    /// Remote frames no not carry any data, yet the DLC can be greater than 0.
```
marcelbuesing, line 36: Small typo I assume this should be "**d**o not carry"

src/can.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn is_data_frame(&self) -> bool {
+        !self.is_remote_frame()
+    }
+
+    /// Returns the frame identifier.
+    fn id(&self) -> u32;
+
+    /// Returns the data length code (DLC) which is in the range 0..8.
+    ///
+    /// For data frames the DLC value always matches the lenght of the data.
```
schultetwin1, line 35:
```suggestion
    /// For data frames the DLC value always matches the length of the data.
```

timokroeger:
Thank you for the comments and suggestions!

> * I had to [unwrap()](https://github.com/marcelbuesing/socketcan-rs/blob/embedded-hal/src/lib.rs#L678) when creating the CanFrame because the identifier may be too large to fit into the frame u32 > 29bit for EFF and u32 > 11bit for SFF. So I assume it's probably a good idea to make both `new_extended` and `new_standard` return a `Result`.

Makes sense looking at your socketcan implementation. I did not put a `Result` because I could not come up with a recovery scenario. When an application wants to create a frame with an invalid identifier something must be going wrong already. Anyway, I’m not feeling strongly about this.

> * Unable to implement `with_rtr` without modification, I assume this is not really critical.

socketcan should [be able to support this](https://github.com/mbr/socketcan-rs/blob/3fa01909d084ece79438b7b91795dc66c97c5c46/src/lib.rs#L570). The rust wrapper should be extended with a method to set that bit.

> * The Ok return type documentation for Transmitter::transmit was  a little unclear to me. I assume the returned frame would be one with lower priority already in the buffer right?

You understood that correctly: The returned frame is the one with the lowest priority of all pending messages. As a non-native speaker I’m glad to take suggestions for better a description.

reneherrero:
Wondering what's up with this PR?  Looking forward to having this available...

src/can.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn dlc(&self) -> usize;
+
+    /// Returns the frame data (0..8 bytes in length).
+    fn data(&self) -> &[u8];
+}
+
+/// A CAN interface that is able to transmit frames.
+pub trait Transmitter {
+    /// Associated frame type.
+    type Frame: Frame;
```
reneherrero, line 46: Is this going to work for device crates? (be advised that I may very well be off)

Here's what happens when I tried doing so:

1. Because it's a device crate, there's no transmitter implementation, I just use the traits like this:
```
pub struct Device<T: Receiver + Transmitter> {
    bus: T
}
```
2. When I go to use the 'Transmitter::transmit' function, I only have a `Frame` trait to work with so I wrote a `CanFrame` struct that implements the `Frame` trait.
3. Problem is I can't for the life of me pass my `CanFrame` struct to the `transmit` function.  Here's what my implementation looks like: (cropped for readability) 
```
impl<T, E, F> Device<T>
where
    T: Receiver<Error = E, Frame = F> + Transmitter<Error = E, Frame = F>,
    F: crate::hal::can::Frame,
    E: Debug,
{
...
```

Note also that I tried creating a Mock Transmitter and I could include the types and that works fine:
```
    impl Transmitter for MockCanBus {
        type Frame = CanFrame;
        type Error = ();

        fn transmit(&mut self, frame: &CanFrame) -> nb::Result<Option<Self::Frame>, Self::Error> {
            self.frames.push(frame.clone());
            Ok(Option::None)
        }
    }
```

Is there problem between the chair and the keyboard?

timokroeger: > Is this going to work for device crates?

This is a very valid question! Actually I started this PR because the previous attempts for CAN traits failed to realize this.

For embedded-hal traits there are usually two perspectives:
1. HAL implementations (like `stm32f1xx-hal` or `lpc8xx-hal`) which actually provide and have the `impl Trait for xxx`.
2. Device drivers which only consume the trait interfaces (but do not implement them)

Now to your points regarding the implementation of a device driver:
```` rust
pub struct Device<T: Receiver + Transmitter> {
    bus: T
}
````
This will work fine for crates that implement the `Receiver` and `Transmitter` trait on the same type. This might not always be the case. To make the driver work in the general I recommend following code (note how the traits bound can also be omitted here):
``` rust
struct Device<Rx, Tx> {
    rx: Rx,
    tx: Tx,
}
````

> When I go to use the 'Transmitter::transmit' function, I only have a Frame trait to work with so I wrote a CanFrame struct that implements the Frame trait

Now comes the tricky part and you were almost there. Crucial point is that you do not need to include the `Frame` in the generic parameter list. To use the `Frame` trait you access it through the `Tx` type parameter instead like `Tx::Frame`.

```` rust
impl<Tx, ...> Device<Tx, ...>
where
    Tx: Transmitter,
    ...
{
    pub fn example() {
        let tx_frame = Tx::Frame::new_standard(0x123, &[]);
        block!(self.tx.transmit(&tx_frame)).unwrap();
    }
}
````

Hope that clarifies some things!

For details you can have a look at this [example device driver](https://github.com/timokroeger/pcan-basic-rs/blob/6a4868bb341ff1c5807a77e484e01cda02398352/pcan-basic/examples/stm32-fwupdate.rs).

Comment (reneherrero) [reply src/can.rs:46]: Yep, that works.  Thanks a million Timo for the heads up.  Very elegant approach I must say (once you wrapped your head around it).

It might be a good idea to note this in the `Frame`'s and the associated type in `Transmitter` (above `type Frame: Frame;`) documentation?  One is usually tempted to go straight to what he needs, but in this case, there's a little detour to take... something tells me I won't be the only one to make this mistake.

FYI: I'm going to take a stab at implementing a Linux implementation.  Only thing I found out there is socketcan-rs, but that's terribly outdated and includes too many dependencies for my liking.

Again, compliments on your implementation, very nice work

Comment (reneherrero) [reply src/can.rs:46]: Made some progress.  I have a first draft of a Linux implementation loosely based on socketcan-rs that only uses libc here: https://github.com/reneherrero/candev

A couple of comments:
1. Would it not be a good idea to have `Frame::new_standard` and `Frame::new_extended` return a result?  Unless I'm missing something, as it stands, it is possible to pass and incorrect id or too much data?
2. I made a few examples to test things out and can transmit and receive data fine using Timo's traits.  But when I try to implement a trivial [driver](https://github.com/reneherrero/candev/blob/master/examples/driver.rs), it does not recognize that the `Receiver::Frame` and the `Transmitter::Frame` are the same type and hence doing the same as @timokroeger does in the `stm32f1xx-hal` `echo` example gives me an `mismatched types` error.  Not clear on the why, but pretty sure it's something I'm doing wrong.
```
if let Ok(frame) = self.t.receive() {
  self.t.transmit(&frame);
                  ^^^^^^ expected embedded_hal::can::Transmitter::Frame, found embedded_hal::can::Receiver::Frame
}
```

Thoughts?

Comment (timokroeger) [reply src/can.rs:48]: Good to see more implementations to verify and improve the traits!

1. Pushed a commit to fix this. As noted previously I do not feel like it makes sense to recover from an error during frame creation. That’s why I did not add an additional `Error` associate type. I opted for a simple `()` error case to keep complexity low.

2. To make this work rust needs to support https://github.com/rust-lang/rust/issues/20041. I have think more about this one. Maybe merging receiver and transmitter traits makes sense after all.

Comment (reneherrero) [reply src/can.rs:48]: Took a stab at the filters yesterday.

Implementing the `Frame` trait went fairly well: just need to do some testing to ensure correctness...  the Soketcan documentation is kind of bare.

Where I did run into a wall is with the `FilteredReceiver`: Socketcan doesn't seem to have the notion of filter groups and I'm not sure how to implement the 'FilterGroup' and `FilterGroups` traits along with the `filter_groups` method which 'FilteredReceiver' needs.

Multiple ways to see this:
1. Would be possible to just not implement `FilteredReceiver` then create and configure the filters on the Socket object and pass it along to whatever.  Problem with that is that anything using `FilteredReceiver` wouldn't work: not good.
2. Other is to just stub it out the parts that don't make sense which would at least enable creating filters directly with the hal.  Better than (1) but still not great.
3. Another I can think of is a blend of (1) and (2), separating the filter group pieces from `FilteredReceiver`.

Tough call.  I hate the fact that looking for the lowest common denominator often results in less than optimal results.

Any suggestions?  Are partial implementations an ok thing?

Comment (timokroeger) [reply src/can.rs:48]: For time reasons only abrief comments. I’ll revisit later.

SocketCAN handles filtering in SW so theoretically it has no limit on the number of filters.  
When all filters have the same capability (which I think they have for SocketCAN) then there is only one `FilterGroup`.  
Because there is just one `FilterGroup` you can simple use `iter::Once<FilterGroup>` for the `FilterGroups` associate type  and return `iter::once(…)` from the `filter_groups()` method.

Comment (reneherrero) [reply src/can.rs:48]: I understand.  I looked at your stm32f1xx-hal implementation for inspiration but couldn't see how to translate it to SocketCAN.

In the meantime I added a basic CI, docs and updated the dependencies to use your latest changes.  Will ponder on the idea...

Thanks Timo

Comment (timokroeger) [reply src/can.rs:48]: Ref: https://www.kernel.org/doc/html/v5.4/networking/can.html#raw-protocol-sockets-with-can-filters-sock-raw

The filter group trait implementation for SocketCAN can look like:
```` rust
pub struct FilterGroup;

impl embedded_hal::can::FilterGroup for FilterGroup {
    fn num_filters(&self) -> usize {
        // CAN_RAW_FILTER_MAX
        // https://github.com/torvalds/linux/blob/v5.4/include/uapi/linux/can.h#L218
        512
    }

    fn extended(&self) -> bool {
        // Filter works for extended (29bit) identifiers
        true
    }

    fn mask(&self) -> Option<MaskType> {
        // There is a configurable mask for each filter
        Some(MaskType::Individual)
    }

    fn rtr(&self) -> RtrFilterBehavior {
        // RTR bit is part of the the filter and the mask
        RtrFilterBehavior::Configurable
    }
}

impl embedded_hal::can::FilteredReceiver for SocketCanDevice {
    type Filter = Filter;
    type FilterGroup = FilterGroup;
    type FilterGroups = std::iter::Once<FilterGroup>;

    fn filter_groups(&self) -> Once<FilterGroup> {
        // All filters have the same capability: There is only one group.
        std::iter::once(FilterGroup)
    }

    // ...
}
````

Comment (timokroeger) [reply src/can.rs:48]: Found another error in the filter description and added a commit to clarify.
Thank you for implementing the traits for SocketCAN! It really helps to polish the interfaces.

Comment (marcelbuesing) [reply src/can.rs:48]: Might also help to look at the [Broadcast Manager](https://www.kernel.org/doc/Documentation/networking/can.txt) part for the filtering part. I would still split filters and basic receive send the faster we get the basic stuff the better, I think that already makes a lot of things possible.

Comment (reneherrero) [reply src/can.rs:48]: No problem Timo... thank you for your patience/tolerance, I'm getting pretty decent at rust, but still learning.  Same goes for the CAN echo system.

I put all your suggestions in place, although I need to read up some more on filter groups to make sure what I put in did mimics well what's at the hardware level.  Mind you that changes would be internal and not break anything.

Ok, still needs quite a bit of testing and polishing (don't like the candev name either -> suggestions welcome), but it is a decent start I think.

Comment (reneherrero) [reply src/can.rs:48]: There is one thing I should have mentioned, and it only happens when implementing a driver: using the `unwrap` on `transmit` throws a compile error.  Tried to fix this multiple ways but no luck.  Works just fine on the receive and transmit examples.
```
no method named `unwrap` found for enum `std::result::Result<std::option::Option<<T as embedded_hal::can::Transmitter>::Frame>, nb::Error<<T as embedded_hal::can::Transmitter>::Error>>` in the current scope

method not found in `std::result::Result<std::option::Option<<T as embedded_hal::can::Transmitter>::Frame>, nb::Error<<T as embedded_hal::can::Transmitter>::Error>>`

note: the method `unwrap` exists but the following trait bounds were not satisfied:
      `nb::Error<<T as embedded_hal::can::Transmitter>::Error>: std::fmt::Debug`
```

Thoughts?  Line of code is [here](https://github.com/reneherrero/candev/blob/167a201f42963448f7f93c38ee617ca5a22c1519/examples/driver.rs#L26)

Comment (timokroeger) [reply src/can.rs:48]: That is due to the nature of [`nb::Result`](https://docs.rs/nb/1.0.0/nb/index.html).  
`block!(tx.transmit(&frame)).unwrap()` should work.

Comment (reneherrero) [reply src/can.rs:48]: Tried that too: doesn't work either.
```
block!(self.t.transmit(&frame)).unwrap();
   |                            ^^^^^^ method not found in `std::result::Result<std::option::Option<<T as embedded_hal::can::Transmitter>::Frame>, <T as embedded_hal::can::Transmitter>::Error>`
   |
   = note: the method `unwrap` exists but the following trait bounds were not satisfied:
           `<T as embedded_hal::can::Transmitter>::Error: std::fmt::Debug`
```

I did finally find a way to fix this, but not sure it's the right way of going about it: by changing `type Error` in [`Transmiter`](https://github.com/timokroeger/embedded-hal/blob/2c935c1e2c1c6229ef3fc590da519a13b90faae6/src/can.rs#L51) to `type Error: core::fmt::Debug;`

You can reproduce by creating an `examples` folder in your copy of `embedded-hal` and adding an `rs` file with this:
```
use embedded_hal::can::{Frame, Receiver, Transmitter};
use nb::block;

#[derive(Debug)]
struct Driver<T> {
    pub id: u32,
    pub t: T,
}

impl<T> Driver<T>
where
    T: Receiver + Transmitter,
{
    pub fn echo(&mut self) {
        let mut count = 0;
        while count < 3 {
            /* This doesn't work:
            if let Ok(frame) = self.t.receive() {
                self.t.transmit(&frame);
            }
            */
            if let Ok(frame) = self.t.receive() {
                let frame = <T as Transmitter>::Frame::new_standard(self.id, frame.data()).unwrap();
                block!(self.t.transmit(&frame)).unwrap();
                // match self.t.transmit(&frame) {
                //     Ok(f) => match f {
                //         None => {}
                //         Some(_) => panic!("Transmit returned a queued frame"),
                //     },
                //     Err(_) => panic!("Failed to transmit frame"),
                // }
            }

            count += 1;
        }
    }
}

fn main(){}
```

Comment (timokroeger) [reply src/can.rs:48]: For `unwrap()` to work you need to add the `Debug` trait bound for `Transmitter::Error` ([example](https://github.com/timokroeger/pcan-basic-rs/blob/master/pcan-basic/examples/stm32-fwupdate.rs#L15-L22)).
To keep things simple your commented out manual error handling is perfectly fine.

Comment (reneherrero) [reply src/can.rs:48]: Thanks to pointing me to that example, got a variation of that to work:
```
impl<T> Driver<T>
where
    T: Receiver + Transmitter,
    <T as Transmitter>::Error: core::fmt::Debug,
{ ... }
```

But to your point, any real world scenario probably wouldn't use `unwrap`, or send back the same exact message it received for that matter.

Hope this helps move this PR forward: looking pretty solid from where I'm standing.

@ilya-epifanov Can you comment?

Comment (timokroeger) [reply src/can.rs:48]: @reneherrero
Found more elegant way to achieve frame loopback which actually works on stable:
```` rust
impl<T, F, E> Driver<T>
where
    F: Frame,
    E: core::fmt::Debug,
    T: Receiver<Frame = F, Error = E> + Transmitter<Frame = F, Error = E>,
{
    pub fn echo(&mut self) {
        let frame = block!(self.0.receive()).unwrap();
        self.0.transmit(&frame).unwrap();
    }
}
````

Comment (reneherrero) [reply src/can.rs:48]: WOW!  Thanks for the Associated Types 301 class @timokroeger 🥇

**Comment by reneherrero** [comment]
Hey @timokroeger,

You *might* be asked to rename several trait methods with the [`try_` prefix for fallible methods](https://github.com/rust-embedded/embedded-hal/pull/192) to prevent name collisions.

I noticed this while trying to submit a PR in linux-embedded-hal... complicates things a little.

Thoughts?

**Comment by reneherrero** [comment]
@thejpster  I'm guessing a) you got lots on your plate or b) you're out on vacation and very intoxicated (hope it's b)...  looking forward to getting your input.

Seems to me like adding CAN support would be a pretty big deal, and Timo's implementation is very solid.

**Comment by timokroeger** [comment]
I'm on holidays. Give me another week to write a summary about the state of this pull request und outstanding changes/ideas.
The `try_` prefix is certainly something we should consider.
AFAIK thejpster has hibernated his embedded-hal activities. It's just the automatic reviewer list that is out of date.

**Comment by reneherrero** [comment]
Gotcha.  Enjoy your off-time and stay safe.

**Comment by timokroeger** [comment]
Added `candev` to the issue description and included a list of open items.

### New requirements

Protocol libraries like [UAVCAN](https://uavcan.org/) are the intended consumers of the CAN traits.
Luckily UAVCAN has a list of [requirements](https://forum.uavcan.org/t/rtos-can-api-requirements/706) for the CAN interface.
We are are missing two of them:
  * Transmission deadline
  * RX and TX timestamps

I’m not sure if it possible to add TX deadlines in nice generic way (SocketCAN for example cannot implement it).

RX timestamps, in contrast, can IMO easily be added with a `fn timestamp() -> Option<Timestamp>` method on the `Frame` trait.
For TX a common approach is to make use of loopback frames.

### Combine `Receiver` and `Transmitter` traits

To make the loopback approach work we need to merge the  `Receiver` and `Transmitter` traits.
This comes with additional benefits of simpler drivers as @reneherrero noted.

```` rust
impl<Can, F, E> Driver<Can>
where
    F: Frame,
    E: core::fmt::Debug,
    Can: Receiver<Frame = F, Error = E> + Transmitter<Frame = F, Error = E>,
````

vs.

```` rust
impl<Can> Driver<Can>
where
    Can: embedded_hal::can::Can,
    Can::Error: core::fmt::Debug,
````

I will prototype those ideas with PCAN and stm32f1xx-hal to see how well they work.
What do you think?

**Comment by marcelbuesing** [comment]
I think adding `fn timestamp() -> Option<Timestamp>` makes sense to the `Frame` trait.  I just updated the [socketcan-fork](https://github.com/mbr/socketcan-rs/commit/9b69cc048ab3ebff5214225c03f56f42f4d63281). Seems to work fine. I think it might be good to have an easy way for extracting the id from the enum. E.g. an `Into<u32> for Id`.

File: src/can.rs [hunk 4]
```diff
@@ -0,0 +1,192 @@
...
+
+    /// Associated error type.
+    type Error;
+
+    /// Puts a frame in the transmit buffer.
+    ///
+    /// If the buffer is full, this function will try to replace a lower priority frame
+    /// and return it. This is to avoid the priority inversion problem.
+    /// Transmits frames of equal identifier in FIFO fashion.
+    fn transmit(&mut self, frame: &Self::Frame) -> nb::Result<Option<Self::Frame>, Self::Error>;
```
ryankurte, line 76: as predicted above, we're now prefixing fallible methods with `try_` (sorry for the hassle!)

src/can.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// For data frames the DLC value always matches the length of the data.
+    /// Remote frames do not carry any data, yet the DLC can be greater than 0.
+    fn dlc(&self) -> usize;
+
+    /// Returns the frame data (0..8 bytes in length).
+    fn data(&self) -> &[u8];
+}
+
+/// A CAN interface that is able to transmit and receive frames.
+pub trait Can {
```
ryankurte, line 64: is it useful to have a blocking implementation to go along with this? (see [1](https://github.com/rust-embedded/embedded-hal/blob/master/src/spi.rs) [2](https://github.com/rust-embedded/embedded-hal/blob/master/src/blocking/spi.rs)

timokroeger:
Changes
* Removed the filter traits (I will open a seperate pull request for them)
* Added `try_` prefixes
* Added default blocking implementation

@marcelbuesing 
I also prototyped some timestamping. I think the best solution is to go with [`embedded-time`](https://github.com/FluenTech/embedded-time) in the long run. Gonna open another PR soon to show the progress here.

I’m hesitant to add the `Into<u32>` / `From<Id>` impls. In systems which use both standard and extended identifiers they convert into the same number even though the are distinct. Having the user explicitly check the type prevents bugs in the case of mixed IDs.

timokroeger:
Please have a good look at the blocking traits! I’m not overly confident, because I’m not fully familiar with the pattern yet.
Is it a bad thing for both traits to have the exact same function names?

Example blocking trait implementations and usages:
* [PCAN-Basic](https://github.com/timokroeger/pcan-basic-rs/commit/0f25fa7d9c7447590daa7802af103fee58b19a67)
* [stm32f1xx-hal](https://github.com/timokroeger/stm32f1xx-hal/commit/7249d2e6034323bd83849e3625d2236337eb53d8)

marcelbuesing:
> Please have a good look at the blocking traits! I’m not overly confident, because I’m not fully familiar with the pattern yet.
> Is it a bad thing for both traits to have the exact same function names?
> 
> Example blocking trait implementations and usages:
> 
>     * [PCAN-Basic](https://github.com/timokroeger/pcan-basic-rs/commit/0f25fa7d9c7447590daa7802af103fee58b19a67)
> 
>     * [stm32f1xx-hal](https://github.com/timokroeger/stm32f1xx-hal/commit/7249d2e6034323bd83849e3625d2236337eb53d8)

Well it seem mostly analogous to the serial implementation, although serial seems to add a `b` to the default blocking implementation fns. I assume this might become relevant once there are non blocking default implementations.

Regarding the timestamp, not sure what the policy is regarding dependencies. Maybe this could be put behind a feature flag making the dependency to embedded-time optional.

> I’m hesitant to add the Into<u32> / From<Id> impls. In systems which use both standard and extended identifiers they convert into the same number even though the are distinct. Having the user explicitly check the type prevents bugs in the case of mixed IDs.

Yes I assume you are right, it might be a little inconvenient but it probably makes more sense this way.

Btw. not really relevant but since I mostly worked with a Socketcan so far, I'd consider naming the fns `try_read` and `try_write`.

marcelbuesing:
I'm attempting to use the fork for an ISOTP implementation but I'm stuck with Error not enforcing being std::fmt::Debug. Any ideas ;)?

```Rust
use thiserror::Error as ThisError;

#[derive(ThisError, Debug)]
pub enum Error<TCan> where TCan: Can, <TCan as Can>::Error: std::fmt::Debug {
     // ...
    #[error("CAN error")]
    Can(#[from] TCan::Error),
}
```

```
#[derive(ThisError, Debug)]
   |          ^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T> std::convert::From<T> for T;
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

timokroeger:
@marcelbuesing
Looks like the error described in https://github.com/dtolnay/thiserror/issues/85.
Probably removing `#[from]` fixes it.

Sympatron:
What do you think about starting with a seperate crate for can traits for now? That way it could already be used by other crates.

timokroeger:
Great idea! The traits now live at: https://github.com/timokroeger/embedded-can and there is a v0.1 release on crates.io

reneherrero:
Just want to point out that @timokroeger 's crate has 39,887 downloads.  I think that's a pretty clear indicator that CAN support should be included here.

eldruin:
Here an update about the changes that have happened `embedded-hal`:
- This should be in a new `can` module with `blocking`/`nb`
- We decided against including default implementations (#289)
- We removed all `try_` prefixes.
- You can propose error variants for this over at #296 

It is great to see these traits have so many users and 2 implementations. I would welcome these traits on `embedded-hal`.
Given the field experience:
1. What are the open design questions?
2. Do you foresee any breaking changes?

rene-at-dell:
Appreciate you looking into this @eldruin Timo's implementation has matured over time.  `0.3.0` is quite a bit simpler than what's here: I like it, much easier to consume.  

@timokroeger Why did you switch the 'Id' trait to an actual implementation?  The one piece I'm really not crazy about are the [constructors](https://github.com/timokroeger/embedded-can/blob/b66e70867eadde9cbd49c07243930ed0e540eaaf/src/id.rs#L18).  If the value is out of range, shouldn't that be an error?  Take 4? :)

timokroeger:
Thank you bringing the PR to my attention again.
The comment section on this PR has become rather big. I created a new take 4 PR with the changes suggested by @eldruin.

@reneherrero
I found that all implementations of an `Id` trait would look almost the same and with a concrete implementation we do not need another associate trait for the `Can` interface.

I still have to catch up #296 on the error handling. If we have concrete error types I agree that it might make sense for the `Id` constructors to return a `Result`

PR #138, opened 2019-05-12
Provide v2_compat for ToggleableOutputPin

little-arhat:
fixes #137

PR #214, opened 2020-05-27
[0.2.x]: Fix warning, remove #[deny(warnings)] but keep check on CI

### LINKED ISSUE #213: error: trait objects without an explicit `dyn` are deprecated
Hi,

I'm getting the following error when trying to compile the library with rust 1.43.1:
```
error: trait objects without an explicit `dyn` are deprecated
   --> src/fmt.rs:6:29
    |
6   | impl<Word, Error> Write for ::serial::Write<Word, Error=Error>
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn (::serial::Write<Word, Error=Error>)`
    |
```

Only way I can see to make it compile is by removing `#![deny(warnings)]` from `lib.rs`, but that doesn't seem right?

I'm also not getting why I can reference the embedded-hal crate from another project (`embedded-hal = "0.2.3"`) without issues, but if I reference the code (`embedded-hal = { path="../../embedded-hal/" }`) I get the above mentioned error?

Thanks in advance,

eldruin:
This fixes #213. Additionally I removed #[deny(warnings)] but added a check for it on CI.
I think we should cherry pick #138 to the 0.2.x branch as well and release a 0.2.4 version, since 1.0 will still take a while.
I can do that in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: rust
 
+env:
+  global:
+    RUSTFLAGS='-D warnings'
+
 matrix:
   include:
     - env: TARGET=x86_64-unknown-linux-gnu
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ version = "1.0.2"
 version = "0.1.1"
 
 [dev-dependencies]
-stm32f30x = "0.6.0"
+stm32f30x = "0.8.0"
 futures = "0.1.17"
 
 [features]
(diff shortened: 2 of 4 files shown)


eldruin:
CI tests failed on rust nightly due to untagged-option. I have updated the `stm32f30x` package now.

adamgreig:
bors retry

ryankurte:
bors retry

PR #216, opened 2020-06-02
Switch from Travis to GitHub actions

therealprof:


therealprof:
@ryankurte This would be ready now.

PR #217, opened 2020-06-02
Restrict ADC ID type to Copy types

therealprof:
This fixes a deny-by-default clippy lint:

```
error: a `const` item should never be interior mutable
  --> src/adc.rs:47:5
   |
47 |     const CHANNEL: Self::ID;
   |     ^^^^^^^^^^^^^^^--------^
   |                    |
   |                    consider requiring `<Self as adc::Channel<ADC>>::ID` to be `Copy`
   |
   = note: `#[deny(clippy::declare_interior_mutable_const)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
```


therealprof:
Need to be applied before #216 will work. ;)

Issue #98, opened 2018-09-30
Consider improving watchdog API using move semantics


hannobraun:
@astro [suggested the following](https://github.com/rust-embedded/embedded-hal/pull/76#issuecomment-417413406):

```rust
/// Feeds an existing watchdog to ensure the processor isn't reset. Sometimes
/// commonly referred to as "kicking" or "refreshing".
#[cfg(feature = "unproven")]
pub trait Watchdog {
    /// Triggers the watchdog. This must be done once the watchdog is started
    /// to prevent the processor being reset.
    fn feed(&mut self);
}

/// Enables A watchdog timer to reset the processor if software is frozen or 
/// stalled.
pub trait WatchdogEnable {
    /// Unit of time used by the watchdog
    type Time;
    /// The started watchdog that should be `feed()`
    type Target: Watchdog;
    /// Starts the watchdog with a given period, typically once this is done 
    /// the watchdog needs to be kicked periodically or the processor is reset. 
    ///
    /// This consumes the value and returns the `Watchdog` trait that you must
    /// `feed()`.
    fn start<T>(self, period: T) -> Self::Target where T: Into<Self::Time>;
}

/// Disables a running watchdog timer so the processor won't be reset.
#[cfg(feature = "unproven")]
pub trait WatchdogDisable {
    type Target: WatchdogEnable;
    /// Disables the watchdog
    ///
    /// This stops the watchdog and returns the `WatchdogEnable` trait so that
    /// it can be started again.
    fn disable(self) -> Self::Target;
}
```

I think this is an interesting approach that's worth thinking about.

luojia65:
This would work well on my HAL crate: https://github.com/gd32v-rust/gd32vf103-hal/blob/68a9d3b594038480504e3b52abea1a48aefaf148/src/wdog.rs
There would be two conditions:
1. The Watchdog after enabled is turned into another struct, like Watchdog<Enabled> or something. This could take ownership of old struct to prevent double `start` calls in compile time, and prevent `disable` calls if the watchdog is already disabled.
2. If we don't need to implement different traits and functions for a disabled watchdog and an enabled one, we may use `type Target = Self`, which can also be adapted to the older WatchdogEnable trait.

eldruin:
PR #222 by @luojia65 was merged and will be released in v1.0.0.
Can this be closed now?

hannobraun:
Yes, I think so.

PR #222, opened 2020-06-16
Improve watchdog API design using move semantics

luojia65:
This pull request improved watchdog API design. When starting watchdog, we may convert it into another type. We may implement different functions for this type. Or downstream developers can implement `Watchdog` for only an enabled type, to prevent feed to disabled watchdogs or forget to enable before feeding. If we are able to stop this watchdog, it can be converted into the former type.

If current design still need a same type after the watchdog is enabled, they may use `Target = Self`. In this way we create a fallback for earlier designs. 

A simple proof of concept: [here](https://github.com/gd32v-rust/gd32vf103-hal/blob/new-watchdog-design/src/wdog.rs#L155-L169) (L120-L153 for its `Enable` implementation, and there is `Disable` implementation)

Related issue: https://github.com/rust-embedded/embedded-hal/issues/98
Earlier discussion: https://github.com/rust-embedded/embedded-hal/pull/76#issuecomment-417413406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- The watchdog API now uses move semantics. See [PR](https://github.com/rust-embedded/embedded-hal/pull/222).
+
 ## [v1.0.0-alpha.1] - 2020-06-16
 
 *** This is an alpha release with breaking changes (sorry) ***
--- a/src/watchdog.rs
+++ b/src/watchdog.rs
@@ -1,7 +1,7 @@
 //! Traits for interactions with a processors watchdog timer.
 
 /// Feeds an existing watchdog to ensure the processor isn't reset. Sometimes
-/// commonly referred to as "kicking" or "refreshing".
+/// the "feeding" operation is also referred to as "refreshing".
 pub trait Watchdog {
     /// An enumeration of `Watchdog` errors.
     ///
@@ -21,23 +21,28 @@ pub trait Enable {
     /// For infallible implementations, will be `Infallible`
     type Error;
 
-    /// Unit of time used by the watchdog
+    /// Unit of time used by the watchdog.
     type Time;
 
-    /// The started watchdog that should be `feed()`
+    /// The started watchdog that should be `feed()`.
     type Target: Watchdog;
 
-    /// Starts the watchdog with a given period, typically once this is done 
-    /// the watchdog needs to be kicked periodically or the processor is reset. 
+    /// Starts the watchdog with a given period, typically once this is done
+    /// the watchdog needs to be `feed()` periodically, or the processor would be
+    /// reset.
     ///
     /// This consumes the value and returns the `Watchdog` trait that you must
     /// `feed()`.
-    fn try_start<T>(&mut self, period: T) -> Result<Self::Target, Self::Error>
+    fn try_start<T>(self, period: T) -> Result<Self::Target, Self::Error>
     where
         T: Into<Self::Time>;
 }
 
 /// Disables a running watchdog timer so the processor won't be reset.
+///
+/// Not all watchdog timers support disable operation after they've been enabled.
+/// In this case, hardware support libraries would not implement this trait
+/// and hardware-agnostic libraries should consider not requiring it.
 pub trait Disable {
     /// An enumeration of `Disable` errors.
     ///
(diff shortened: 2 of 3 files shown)


therealprof:
Thanks for the PR. It looks good to me on first glance but I need to find time for a deeper review.

Two comments off the bat since you're touching this:
- We don't want to kick dogs anymore. 😅 Can we maybe change the kick into pet or something more friendly?
- Most watchdogs cannot be disabled once they've been armed/enabled so it would be great if we could mention that in the docs to not raise wrong expectations

luojia65:
Thank you! I've modified texts to avoid this word on dogs 😅 And additional documents are added to make it clear that, most watchdogs do not support disable process.

eldruin:
A shortcoming of this approach is that a naive implementation of `start` or `disable` may drop the watchdog instance if the operation failed. I am unsure if that is always fine.
I have encountered this "fallible transformations" issue in several drivers where some methods transform the driver struct type (via a mode marker type) (e.g. from config status to operation status, which expose different methods) and my solution so far is to return the original instance through the error type.
Here is [an example](https://github.com/eldruin/ads1x1x-rs/blob/a302082c2fbeff0403c0abd9bb3931ae3f744b3c/src/devices/mode/oneshot.rs#L15).
This would still be possible to do with this approach but in my attempts non-trivial error recovery code starts getting messier.

luojia65:
@eldruin Thanks for comment! :) You're right, it's somehow tricky in error handling for failable type state classes. Seems like the only way is to return in error type, but this is common on Rust's std crate. For example if a `RwLock` or `Mutex` is poisoned when trying to lock, a [`PoisonError`](https://doc.rust-lang.org/stable/std/sync/struct.PoisonError.html) is returned; this type contains an `into_inner` function to return the guard back. As far as I know, there's rarely an error when enabling or disabling watchdogs; but if there are, embedded drivers could still take this design.

src/watchdog.rs
```diff
@@ -15,29 +15,46 @@ pub trait Watchdog {
...
         T: Into<Self::Time>;
 }
 
 /// Disables a running watchdog timer so the processor won't be reset.
-pub trait WatchdogDisable {
-    /// An enumeration of `WatchdogDisable` errors.
+///
+/// Not all watchdog timers support disable operation after they've been enabled.
+/// In this case, hardware abstract libraries would not consider implementing this
+/// trait.
```
eldruin, line 45:
```suggestion
/// In this case, hardware support libraries would not implement this trait
/// and hardware-agnostic libraries should consider not requiring it.
```

src/watchdog.rs
```diff
@@ -15,29 +15,46 @@ pub trait Watchdog {
...
 
-    /// Disables the watchdog
-    fn try_disable(&mut self) -> Result<(), Self::Error>;
+    /// Disabled watchdog instance that can be enabled.
+    type Target: Enable;
+
+    /// Disables the watchdog.
+    ///
+    /// This stops the watchdog and returns the `Enable` trait so that
+    /// it can be started again.
```
eldruin, line 58:
```suggestion
    /// This stops the watchdog and returns an instance implementing the
    /// `Enable` trait so that it can be started again.
```

src/watchdog.rs
luojia65: Thank you! That looks more fluent.

src/watchdog.rs
luojia65: Thank you!

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- A new watchdog API design using move semantics from [this pull request](https://github.com/rust-embedded/embedded-hal/pull/222).
```
eldruin, line 13:
```suggestion
- The watchdog API now uses move semantics. See [PR](https://github.com/rust-embedded/embedded-hal/pull/222).
```

src/watchdog.rs
```diff
@@ -1,7 +1,7 @@
 //! Traits for interactions with a processors watchdog timer.
 
 /// Feeds an existing watchdog to ensure the processor isn't reset. Sometimes
-/// commonly referred to as "kicking" or "refreshing".
+/// the "feeding" operation is commonly referred to as "refreshing".
```
therealprof, line 4:
```suggestion
/// the "feeding" operation is referred to as "refreshing".
```

therealprof: Or "also".

CHANGELOG.md
luojia65: That's nicer :)

src/watchdog.rs
luojia65: Thanks! 'Also' may be better here; pushed in a new commit.

PR #223, opened 2020-06-17
Add transactional I2C interface

eldruin:
An example where a transactional I2C interface would be an improvement is when facing the problem of sending an array of data where the first item is the destination address.
At the moment this requires copying the data into a bigger array and assigning the first item to the address. With a transactional I2C two `write` operations could be chained into one transaction so that it is possible to send the address and data without an extra copy.
This is specially problematic if the data length is unknown e.g. because it comes from the user.

You can see the problem [here in the eeprom24x driver](https://github.com/eldruin/eeprom24x-rs/blob/f75770c6fc6dd8d591e3695908d5ef4ce8642566/src/eeprom24x.rs#L220). In this case I am lucky enough to have the page upper limit so that the copy workaround is possible.

With this PR a bunch of code and macros could be replaced by doing something similar to this:
```rust
let user_data = [0x12, 0x34, ...]; // defined somewhere else. length may be unknown.
let target_address = 0xAB;
let mut ops = [
  Operation::Write(&[target_address]),
  Operation::Write(&user_data),
];
i2cdev.try_exec(DEV_ADDR, &ops)
```

I added a PoC [here in linux-embedded-hal](https://github.com/eldruin/linux-embedded-hal/blob/7512dbcc09faa58344a5058baab8826a0e0d0160/src/lib.rs#L211) including [an example](https://github.com/eldruin/linux-embedded-hal/blob/transactional-i2c/examples/transactional-i2c.rs) of a driver where a classical combined write/read is performed through the transactional interface.

Note: A default implementation of the `Transactional` trait like in #191 is not possible because STOPs would always be sent after each operation. What is possible is to do is the other way around. This includes an implementation of the `Write`, `Read` and `WriteRead` traits for `Transactional` implementers.

This is based on previous work from #178 by @ryankurte and it is similar to #191 

TODO:
- [x] Add changelog entry

### DIFF
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -270,3 +270,54 @@ pub trait Transactional {
         operations: &mut [Operation<'a>],
     ) -> Result<(), Self::Error>;
 }
+
+/// Default implementation of `blocking::i2c::Write`, `blocking::i2c::Read` and
+/// `blocking::i2c::WriteRead` traits for `blocking::i2c::Transactional` implementers.
+pub mod transactional {
+    use super::{Operation, Read, Transactional, Write, WriteRead};
+
+    /// Default implementation of `blocking::i2c::Write`, `blocking::i2c::Read` and
+    /// `blocking::i2c::WriteRead` traits for `blocking::i2c::Transactional` implementers.
+    pub trait Default<E> {}
+
+    impl<E, S> Write for S
+    where
+        S: self::Default<E> + Transactional<Error = E>,
+    {
+        type Error = E;
+
+        fn try_write(&mut self, address: u8, bytes: &[u8]) -> Result<(), Self::Error> {
+            self.try_exec(address, &mut [Operation::Write(bytes)])
+        }
+    }
+
+    impl<E, S> Read for S
+    where
+        S: self::Default<E> + Transactional<Error = E>,
+    {
+        type Error = E;
+
+        fn try_read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Self::Error> {
+            self.try_exec(address, &mut [Operation::Read(buffer)])
+        }
+    }
+
+    impl<E, S> WriteRead for S
+    where
+        S: self::Default<E> + Transactional<Error = E>,
+    {
+        type Error = E;
+
+        fn try_write_read(
+            &mut self,
+            address: u8,
+            bytes: &[u8],
+            buffer: &mut [u8],
+        ) -> Result<(), Self::Error> {
+            self.try_exec(
+                address,
+                &mut [Operation::Write(bytes), Operation::Read(buffer)],
+            )
+        }
+    }
+}
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -232,3 +232,39 @@ pub trait WriteIterRead<A: AddressMode = SevenBitAddress> {
     where
         B: IntoIterator<Item = u8>;
 }
+
+/// Transactional I2C operation.
+///
+/// Several operations can be combined as part of a transaction.
+#[derive(Debug, PartialEq)]
+pub enum Operation<'a> {
+    /// Read data into the provided buffer
+    Read(&'a mut [u8]),
+    /// Write data from the provided buffer
+    Write(&'a [u8]),
+}
+
+/// Transactional I2C interface.
+///
+/// This allows combining operation within an I2C transaction.
+pub trait Transactional {
+    /// Error type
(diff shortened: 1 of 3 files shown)


src/blocking/i2c.rs
```diff
@@ -125,3 +125,39 @@ pub trait WriteIterRead {
...
+    /// - After executing the last operation an SP is sent automatically.
+    /// - If the last operation is a `Read` the master does not send an acknowledge for the last byte.
+    ///
+    /// - `ST` = start condition
+    /// - `SAD+R/W` = slave address followed by bit 1 to indicate reading or 0 to indicate writing
+    /// - `SR` = repeated start condition
+    /// - `SP` = stop condition
+    fn try_exec<'a, O>(&mut self, address: u8, operations: O) -> Result<(), Self::Error>
+    where
+        O: AsMut<[Operation<'a>]>;
```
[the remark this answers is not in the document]
eldruin: Interesting. Are you sure this is also a problem for the I2C `Operation` type, which is not generic over the word type like in the SPI case?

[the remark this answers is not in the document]
ryankurte: I'm not sure, and unfortunately I squashed the commits on #191 so I can't even go back and find it. I think either way it's probably still better to be consistent?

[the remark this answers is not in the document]
eldruin: I see, that's unfortunate. Ok, the `AsMut` case would anyway be a relaxation of the slice type so this would not really be a breaking change if someone needs that in the future.

src/blocking/i2c.rs
```diff
@@ -125,3 +125,92 @@ pub trait WriteIterRead {
...
+pub enum Operation<'a> {
+    /// Read data into the provided buffer
+    Read(&'a mut [u8]),
+    /// Write data from the provided buffer
+    Write(&'a [u8]),
+}
+
+/// Transactional I2C interface.
+///
+/// This allows combining operation within an I2C transaction.
```
therealprof, line 142:
```suggestion
/// This allows combining operations within an I2C transaction.
```

therealprof:
Doesn't the trait need an entry in the prelude?

eldruin:
> Doesn't the trait need an entry in the prelude?

Done.

src/blocking/i2c.rs
```diff
@@ -125,3 +125,92 @@ pub trait WriteIterRead {
...
+    /// - `SR` = repeated start condition
+    /// - `SP` = stop condition
+    fn try_exec<'a>(
+        &mut self,
+        address: u8,
+        operations: &mut [Operation<'a>],
+    ) -> Result<(), Self::Error>;
+}
+
+/// Default implementation of `blocking::i2c::Write`, `blocking::i2c::Read` and
```
ryankurte, line 167: oh interesting, I guess it makes sense to have the default trait be this way around in the I2C case? (the SPI one is opposite).

src/blocking/i2c.rs
```diff
@@ -125,3 +125,92 @@ pub trait WriteIterRead {
...
+///
+/// This allows combining operations within an I2C transaction.
+pub trait Transactional {
+    /// Error type
+    type Error;
+
+    /// Execute the provided operations on the I2C bus.
+    ///
+    /// Transaction contract:
+    /// - Before executing the first operation an ST is sent automatically. This is followed by SAD+R/W as appropriate.
```
ryankurte, line 150: is any customisation of this contract required, or is this order and combinations of operations always valid?

src/blocking/i2c.rs
eldruin: Yes. The same way as SPI is not possible because the `Write`, etc traits will always send a STOP after each operation.

src/blocking/i2c.rs
eldruin: From my understanding, any order and combination should be valid. See I2C specification sections [3.2.4](https://www.nxp.com/docs/en/user-guide/UM10204.pdf#G1672579), [3.2.7](https://www.nxp.com/docs/en/user-guide/UM10204.pdf#G1672934) and also [Wikipedia](https://en.wikipedia.org/wiki/I%C2%B2C#Transaction_format).

eldruin:
Is there anything left to do here? Since we have merged transactional SPI already, adding also transactional I2C to the same release would be nice.

therealprof (approved):
Excellent, let's get this in then.

bors r+

therealprof:
... after @ryankurte removed his concern...

ryankurte (approved):

Issue #147, opened 2019-07-12
Support 7-bit and 10-bit I2C addressing


BroderickCarlin:
The current I2C traits assume 7-bit addressing. I recently encountered a situation where I needed to talk to a device that utilized a 10-bit address and while it was possible to manipulate the current trait into working for 10-bit addressing it does require an abstraction layer capable of converting the 10-bit address into a corresponding 7-bit address + payload. While it would be a breaking change, I believe it would be beneficial to support both 7-bit and 10-bit addressing in the trait definitions itself to ease usage for developers that are attempting to interface with objects that implement these traits. One potential solution I'd like to put forward is defining an enum such as:
```
enum I2cAddress {
    SevenBit(u8),
    TenBit(u16),
}
```
which would be used in a trait definition such as:
```
pub trait Read {
    type Error;

    fn read(&mut self, address: I2cAddress, buffer: &mut [u8]) -> Result<(), Self::Error>;
}
```

A quick search showed this topic has been brought up in the past but was not actively discussed and no solutions put forward. Even though the thread has been quiet recently I believe the conversation in https://github.com/rust-embedded/embedded-hal/issues/100 would be relevant to a change such as this.

Let me know thoughts; I'd be happy to do some work and throw up a PR with these changes!

PS: If we are making breaking changes we might want to also discuss standardizing on `addr` or `address` in the trait definitions to avoid the mismatch that there is now 😄

Rahix:
Not all I2C bus drivers support 10-bit addressing so I don't think it is a good idea to allow 10-bit addressing in general.  Instead, I'd suggest having separate traits for 7 and 10-bit addressing with an optional default impl of 7-bit addressing for 10-bit implementations.

ryankurte:
> I'd suggest having separate traits for 7 and 10-bit addressing

Good call i think, what about genericising the i2c traits (ie. `read<Word>`)  to be consistent with the [spi](https://github.com/rust-embedded/embedded-hal/blob/a5c0e3c63e4e431807d491d35dd67d1d43c07c92/src/spi.rs#L15) traits?

BroderickCarlin:
>Not all I2C bus drivers support 10-bit addressing 

Can you provide an example of any hardware that would not be able to support 10-bit addressing? 

[...]

PR #230, opened 2020-07-04
Make I2C compatible with multiple address sizes

eldruin:
This adds I2C 7-bit and 10-bit address mode compatibility as roughly described [here](https://github.com/rust-embedded/embedded-hal/issues/147#issuecomment-511703503).
Discussion issue: #147 

I have also added the `SevenBitAddress` as the default address mode to the traits so this is not even a breaking change.

Usage broken down per use case:
* **Device driver which only supports 7-bit addressing mode:**
       The driver looks exactly the same as now since the default address mode is 7-bit.

```rust
 impl<I2C, E> MyDriver<I2C>
 where I2C: i2c::Write<Error = E> {
   pub fn do_cool_stuff(&mut self) // ...
 }
 ```

* **Device driver which only supports 10-bit addressing mode:**
       The only difference to a 7-bit-address-only driver is one additional parameter in the I2C trait bound.
```rust
 impl<I2C, E> MyDriver<I2C>
 where I2C: i2c::Write<TenBitAddress, Error = E> {
   pub fn do_cool_stuff(&mut self) // ...
 }
 ```
 
* **Driver for device supporting both addressing modes:**
       Complexity can be abstracted away into additional internal traits which can handle the addressing stuff. Driver code stays clean.
       **This is nothing new**. We already do this on drivers for devices compatible with both I2C and SPI. No need for duplicated code.
       Here a real example: [usage](https://github.com/eldruin/bmi160-rs/blob/3af5637f1df047bb811a4885525cfbe8b44d8ede/src/device_impl.rs#L43), [traits](https://github.com/eldruin/bmi160-rs/blob/master/src/interface.rs)
  
 ```rust
 impl<DI, E> MyDriver<DI>
 where DI: WriteData<Error = E> {
   pub fn do_cool_stuff(&mut self) {} // ...
 }
 
 pub trait WriteData {
 // ...
 }
 
// it is also possible to just leave the `SevenBitAddress` type out here,
// since it is the default.
 impl<I2C, E> WriteData for I2cInterface<I2C, SevenBitAddress>
 where
     I2C: i2c::Write<SevenBitAddress, Error = E>,
 {
   // ...
 }
 
 impl<I2C, E> WriteData for I2cInterface<I2C, TenBitAddress>
 where
     I2C: i2c::Write<TenBitAddress, Error = E>,
 {
   // ...
 }
 ```
 
* **Bus controller impl supporting only 7-bit addressing mode:**
       Code stays almost the same, just adding one addressing mode parameter. Additionally, _if desired_:
    * 10-bit addressing can be software-emulated:
         Emulate by extending and copying payload in separate `TenBitAddress` implementation. Total flexibility to do whatever is necessary in this case since the code is independent.
    * 10-bit addressing cannot be software-emulated:
         Implementation does not offer implementation for `TenBitAddress` variant. The user gets a compilation error and everything is clear.
 
* **Bus controller impl supporting both addressing modes:**
       No problem. Two separate implementations guarantee as much flexibility as necessary. At the same time, sharing generic code is possible.
  
 Additional benefits: 
* No runtime performance cost 
* No runtime switching, duplicated code or panics for unsupported modes. 
* Consistent with what we do for code paths that can be determined statically by the compiler.
* To my taste elegant, simple and very descriptive.

See [here](https://github.com/rust-embedded/embedded-hal/issues/147#issuecomment-647157906) for a comparison to other alternatives.

I have also sealed the trait.

## Proof
* A HAL implementation of both modes: [bitbang-hal](https://github.com/eldruin/bitbang-hal/tree/i2c-multi-address-mode). [code changes](https://github.com/eldruin/bitbang-hal/compare/embedded-hal-1.0.0-alpha.1...eldruin:i2c-multi-address-mode)
* Drivers supporting only 7-bit addresses need **no changes**.
For demonstration purposes, explicitly including the `SevenBitAddress` would look like this: [OPT300x](https://github.com/eldruin/opt300x-rs/tree/i2c-multi-address-mode). [code changes](https://github.com/eldruin/opt300x-rs/compare/i2c-multi-address-mode).
This would be similar to the case of a 10-bit-only device driver.

### DIFF
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -1,12 +1,119 @@
 //! Blocking I2C API
 //!
-//! Slave addresses used by this API are 7-bit I2C addresses ranging from 0 to 127.
+//! This API supports 7-bit and 10-bit addresses. Traits feature an `AddressMode`
+//! marker type parameter. Two implementation of the `AddressMode` exist:
+//! `SevenBitAddress` and `TenBitAddress`.
 //!
-//! Operations on 10-bit slave addresses are not supported by the API yet (but applications might
-//! be able to emulate some operations).
+//! Through this marker types it is possible to implement each address mode for
+//! the traits independently in `embedded-hal` implementations and device drivers
+//! can depend only on the mode that they support.
+//!
+//! Additionally, the I2C 10-bit address mode has been developed to be fully
+//! backwards compatible with the 7-bit address mode. This allows for a
+//! software-emulated 10-bit addressing implementation if the address mode
+//! is not supported by the hardware.
+//!
+//! Since 7-bit addressing is the mode of the majority of I2C devices,
+//! `SevenBitAddress` has been set as default mode and thus can be omitted if desired.
+//!
+//! ## Examples
+//!
+//! ### `embedded-hal` implementation for an MCU
+//! Here is an example of an embedded-hal implementation of the `Write` trait
+//! for both modes:
+//! ```
+//! # use embedded_hal::blocking::i2c::{SevenBitAddress, TenBitAddress, Write};
+//! /// I2C0 hardware peripheral which supports both 7-bit and 10-bit addressing.
+//! pub struct I2c0;
+//!
+//! impl Write<SevenBitAddress> for I2c0
+//! {
+//! #   type Error = ();
+//! #
+//!     fn try_write(&mut self, addr: u8, output: &[u8]) -> Result<(), Self::Error> {
+//!         // ...
+//! #       Ok(())
+//!     }
+//! }
+//!
+//! impl Write<TenBitAddress> for I2c0
+//! {
+//! #   type Error = ();
+//! #
+//!     fn try_write(&mut self, addr: u16, output: &[u8]) -> Result<(), Self::Error> {
+//!         // ...
+//! #       Ok(())
+//!     }
+//! }
+//! ```
+//!
+//! ### Device driver compatible only with 7-bit addresses
+//!
+//! For demonstration purposes the address mode parameter has been omitted in this example.
+//!
+//! ```
+//! # use embedded_hal::blocking::i2c::WriteRead;
+//! const ADDR: u8  = 0x15;
+//! # const TEMP_REGISTER: u8 = 0x1;
+//! pub struct TemperatureSensorDriver<I2C> {
+//!     i2c: I2C,
+//! }
+//!
+//! impl<I2C, E> TemperatureSensorDriver<I2C>
+//! where
+//!     I2C: WriteRead<Error = E>,
+//! {
+//!     pub fn read_temperature(&mut self) -> Result<u8, E> {
+//!         let mut temp = [0];
+//!         self.i2c
+//!             .try_write_read(ADDR, &[TEMP_REGISTER], &mut temp)
+//!             .and(Ok(temp[0]))
+//!     }
+//! }
+//! ```
+//!
+//! ### Device driver compatible only with 10-bit addresses
+//!
+//! ```
+//! # use embedded_hal::blocking::i2c::{TenBitAddress, WriteRead};
+//! const ADDR: u16  = 0x158;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/blocking/i2c.rs [hunk 1]
```diff
@@ -5,8 +5,32 @@
...
+/// Address mode (7-bit / 10-bit)
+///
+/// Note: This trait is sealed and should not be implemented outside of this crate.
+pub trait AddressMode: private::Sealed {
+    /// Address value type
+    type Address;
+}
+
+/// 7-bit address mode type
+pub struct SevenBitAddress(());
```
therealprof, line 19: Why the unit type inside the struct?

eldruin: It is a private data member. That way users cannot create an instance. It should only be a marker type.

therealprof: I'm aware what it does, just not sure about the purpose. Which bad things coul happen if someone decided to instantiate a `SevenBitAddress`?

eldruin: It is possible to implement a trait for `SevenBitAddress` but with a private data member it is at least not possible to create an instance and call stateful methods.
I do not know if this would be useful for somebody, though.

eldruin:
I have now also added the `SevenBitAddress` as the default address mode to the traits so this is not even a breaking change.

ryankurte:
seems generally good to me! the trait indirection approach is clever, though, feels a touch complex (and i have been bitten by trying to do overly clever things with the type system 🙃)

does this give us a strong benefit over providing a couple of public type aliases and documentation of the convention (`pub type I2cAddrSevenBit = u8;`, `pub type I2cAddrTenBit = u16;`)? i suppose the perk is you cannot implement the I2C interfaces over external types, with the disadvantage it makes bounds moderately more complex, and either way this must be communicated by documentation.

(i'm not opposed to this approach, but, would like to know we have considered all possibilities)

eldruin:
> does this give us a strong benefit over providing a couple of public type aliases and documentation of the convention (`pub type I2cAddrSevenBit = u8;`, `pub type I2cAddrTenBit = u16;`)?
> i suppose the perk is you cannot implement the I2C interfaces over external types, with the disadvantage it makes bounds moderately more complex, and either way this must be communicated by documentation.

Yes, technically without a trait bound it would be possible to misuse the I2C traits and pass some random object through the address parameter if both HAL implementation and driver agree on a complex "address" type.

It would also be possible to do this simplification:
```rust
pub trait AddressMode: private::Sealed {}

pub type SevenBitAddress = u8;
pub type TenBitAddress = u16;

impl AddressMode for SevenBitAddress {}
impl AddressMode for TenBitAddress {}
```
I think this would be fine, since we probably will not need any additional type other than `Address` in the `AddressMode` trait.

Would you prefer that?

I have added some documentation now on how this would work for a HAL implementation and device drivers. I could also write some documentation about how to write a driver for a device compatible with both modes but that might be too much. Opinions?

burrbull:
Looks good to me. :+1:

ryankurte:
> It would also be possible to do this simplification:

ooh, i was under the impression you can't implement traits on type aliases? though, this may be mitigated by the trait being defined in the same place.

> Would you prefer that?

hmm i think so? it seems a bit simpler, neat that the `Address` type is actually the type of the address (rather than a marker), and i _think_ this will simplify/clarify the docs too. so, weak preference unless there's a good reason for the more complex approach.

eldruin:
> > Would you prefer that?
> 
> hmm i think so? it seems a bit simpler (which is excellent imo), neat that the `Address` type is actually the type of the address (rather than a marker), and i _think_ this will simplify/clarify the docs too. so, weak preference unless there's a good reason for the more complex approach.

Done!

ryankurte:
i think this looks good to me, @therealprof any thoughts?

therealprof:
Haven't had time to play with it in person but looks and sounds great.

PR #236, opened 2020-07-21
PwmPin: allow get_duty to return the old value for some time

eldruin:
This is #140 rebased to master.
However, I suppose the same thing applies to the `Pwm::try_get_duty()` method. Shouldn't the same note be added to that method as well?

Closes #140


therealprof:
> However, I suppose the same thing applies to the `Pwm::try_get_duty()` method. Shouldn't the same note be added to that method as well?

Sure, why not. The more documentation the better. ;)

Issue #141, opened 2019-06-01
Add built-in no-op implementations for GPIO


mvirkkunen:
Dummy `OutputPin`/`InputPin` implementations would be useful for implementing optional I/O pins. A dummy `OutputPin` would simply discard any value written to it, while a dummy `InputPin` would always return a constant value.

# Motivation

One very common use for a dummy `OutputPin` would be SPI chip select pins, which for some chips are optional if you only have one chip on the bus. It could also be useful for data/command, read/write etc pins that the particular circuit doesn't need and has them tied to a constant voltage.

Currently the neatest way to implement an optional I/O pin I can think of is to use a no-op trait implementation.

Even though it would be very simple to just code this, I'm opening an issue because I think the following things should be discussed first:

## Is this efficient?

This should be pretty efficient for the common use case of chip select or read/write pins. The calls into empty functions should be easy to optimize out, there's no need to use an `Option` which may increase memory use or cause runtime overhead, and driver implementations don't even have to take the optionalness into account and can just assume a pin is there.

Additionally what comes to using an `Option` there's also an issue with using it for an optional pin, because `Option<T>` requires *some* type for `T` even if you only ever plan on having a `None`, so you need a dummy `OutputPin` type anyway.

However this probably doesn't make it easy for a driver to "detect" whether a pin has been dummied out or not, so if that makes a difference (for instance if the driver can omit an expensive operation that's needed to figure out what to write to the pin), that case needs to be handled differently.

## Is this safe?

This in theory makes it possible for users to end up with a broken driver instance because they dummied out a pin that is actually required to use the driver. However this is pretty much equivalent to forgetting a PCB trace, and users should be aware of what pins are really needed for their thing to function. A driver could signal optionalness by for instance having the pin as an optional argument in a builder style interface.

## Does this belong in embedded-hal?

All drivers could also have their own `struct DummyOutputPin`, but I think it would make sense to have the dummy implementation directly in some crate. This would probably be the first real "trait implementation" in this crate, so I'm not sure if it's out of scope. Having an entirely new crate just for a couple of dummy implementations sounds like a hassle though.

robamu:
Thanks @eldruin for providing this crate. Are there any plans to make this built-in in the `embedded-hal` ? This seems like a very useful feature commonly needed.
I'm writing a driver for a SPI ADC, and I actually have to add chip selects even though I dont need them for my test board (chip select is managed by the peripheral)

Issue #180, opened 2020-01-21
SPI CS Traits / Implementations


ryankurte:
At the moment we aren't super explicit about management of the chip select pin for SPI transactions, which means drivers have to either manage these themselves or make big assumptions about the underlying SPI implementation.

With #178 we have the ability to execute multiple operations in a single transaction which brings this to the forefront. (@jonas-schievink this is why I didn't call them `Transaction` because the whole thing is a bus transaction, whoops.)

To mitigate this, I'm proposing the addition of a `ChipSelect` marker trait for SPI devices that manage SPI themselves, As well as a wrapper object that takes an SPI object GPIO output and implements `ChipSelect` to simplify the use of devices with manual CS management.

A downside of this is that we now need to return distinct types where `ChipSelect` or `!ChipSelect` in cases where either is possible, for example, the linux embedded hal driver may have CS managed automatically _or_ be used with manual CS management. This could be mitigated by splitting the constructors for `no_cs` and `with_cs` versions, however, will be required for many SPI devices.

thejpster:
How would it work with multiple chip selects? Would you have multiple wrappers sharing the same SPI driver object (in some kind of Cell wrapper) each with a unique GPIO?

ryankurte:
Yeah good question, depends on the implementation (Linux you could just create multiple devices `/dev/spidev0.N` and leave them to manage CS themselves), but in the more embedded case i think [shared-bus](https://github.com/Rahix/shared-bus) could be extended to do the job as you describe.

This is not quite as trivial as using a shared-bus object in the proposed wrapper as you really need to take ownership of the bus _before_ asserting CS, so shared-bus or similar would need to be extended with support for all the CS stuff.

There's also a related issue in shared-bus https://github.com/rahix/shared-bus/issues/8 discussing this exact problem.

Dirbaio:
Solved in #351, #444

PR #237, opened 2020-07-21
Add dummy GPIO pins

eldruin:
This is an implementation of #141 to get the ball rolling.
I put everything into a single `DummyPin` since the level can be put into a typestate for zero cost and then it can be used for cases where the requirement is `InputPin + OutputPin`.
Alternatives:
- Separate `DummyOutputPin`, `DummyInputPin`, `DummyStatefulOutputPin`.
- Accept a bool on creation time for a `DummyInputPin` and `DummyStatefulOutputPin`.


src/digital.rs
```diff
@@ -123,3 +123,112 @@ pub trait InputPin {
...
+        pub fn new() -> Self {
+            DummyPin { _l: PhantomData }
+        }
+
+        /// Create new instance
+        ///
+        /// When checked it will always behave as being low
+        pub fn new_low() -> Self {
+            DummyPin { _l: PhantomData }
+        }
```
therealprof, line 164: Why two identical constructors `new` and `new_low`?

src/digital.rs
```diff
@@ -123,3 +123,112 @@ pub trait InputPin {
...
+
+    impl StatefulOutputPin for DummyPin<level::High> {
+        fn try_is_set_high(&self) -> Result<bool, Self::Error> {
+            Ok(true)
+        }
+
+        fn try_is_set_low(&self) -> Result<bool, Self::Error> {
+            Ok(false)
+        }
+    }
```
therealprof, line 206: Shouldn't a stateful pin really be stateful?

src/digital.rs
eldruin: I thought it would be appropriate to have `new_high()` and `new_low()` if the user is interested in using it as an `InputPin` since this is more explicit than the default `new()`.
Nevertheless, somebody that just wants an `OutputPin` can rather expect a `new()` method.
I should define one by using the other, though.

therealprof: `new_low()` and `new_high()` seem perfectly fine to me. I was confused by `new()`. 😅

src/digital.rs
eldruin: The pin could be made to remember what the last state set but then a boolean needs to be added to the struct and the methods are not zero-cost anymore.
If we want to keep the output pin implementation zero-cost, we could also create a separate `DummyStatefulOutputPin` which adds the boolean.

src/digital.rs
eldruin: ok, sure.

src/digital.rs
therealprof: I don't think it makes sense to implement `StatefulOutputPin` on something which is not stateful as that's really the only purpose of that trait.

eldruin: Ok, I could also just remove the `StatefulOutputPin` implementation. The original motivation issue did not include it anyway.

therealprof: And conversely `toggleable` doesn't make sense on a stateless pin either.

eldruin: alright I removed them.

therealprof: Works for me, alternatively we could do as you suggested and add another non-zero-cost stateful pin although I'm not super convinced whether that is really useful.

eldruin:
I added a changelog entry now, albeit quite short.

src/digital.rs
eldruin: Yeah I think we can go with just this and add a stateful pin later on if somebody asks for it.

eldruin:
I mentioned that it is zero-cost and improved the documentation a bit now. Sorry for the re-review.

therealprof (dismissed):
No worries, looks even better now. 🎉

ryankurte:
hmm i am not so convinced by this approach, it would seem to me to introduce ambiguity that is better expressed via the type system (which does perhaps require more exploration and documentation).

as an example, if a driver _can use_ a pin but does not _require_ it, i believe it should be implemented over both conditions in the manner that is correct for the device, rather than via a dummy pin that the driver cannot be aware of. i have an example of this planned, but, haven’t yet had a chance to implement it...

for the mentioned SPI case i believe further work is required to accurately represent and abstract over transactions and busses with managed or unmanaged `CS` pins (#180).

therealprof:
>  rather than via a dummy pin that the driver cannot be aware of

Why would the application need to be aware of a dummy pin? In the end it's always the user who needs to decide how to wire the hardware and how to configure it.

We do have dummy pins of the required type/implementing the required traits pretty in pretty much any HAL impl which is okay but not great. The larger issue here is that in generic drivers you cannot make use those so either you require the user to provide the correct types or implement the dummy pin themselves. I think having these here provides a nice and coherent addition to the traits which might also be useful to write tests, etc.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - 10-bit addressing mode for I2C traits.
+- Dummy GPIO pin (No-op, zero cost) struct `DummyPin`, useful when dealing
```
dbrgn, line 13: Nitpicking...

```suggestion
- Dummy GPIO pin (no-op, zero cost) struct `DummyPin`, useful when dealing
```

src/digital.rs
```diff
@@ -123,3 +123,82 @@ pub trait InputPin {
...
+
+    /// Pin level marker types for usage of `DummyPin` as an `InputPin`.
+    pub mod level {
+        /// `DummyPin` will always behave as being high when checked.
+        pub struct High;
+        /// `DummyPin` will always behave as being low when checked.
+        pub struct Low;
+    }
+
+    /// Dummy (No-op, zero-cost) pin
```
dbrgn, line 143: Nitpicking...

```suggestion
    /// Dummy (no-op, zero-cost) pin
```

src/digital.rs
```diff
@@ -123,3 +123,82 @@ pub trait InputPin {
...
+    ///
+    /// This will discard any value set to it and when checked always behave
+    /// according to the value provided at construction time (high/low).
```
dbrgn, line 146: I think this wording would be a bit clearer.

```suggestion
    /// The implementation will discard any value written to it. When read,
    /// it will always behave according to the value provided at construction
    /// time (high or low).
```

src/digital.rs
```diff
@@ -123,3 +123,82 @@ pub trait InputPin {
...
+    pub struct DummyPin<L = level::Low> {
+        _l: PhantomData<L>,
+    }
+
+    impl DummyPin<level::Low> {
+        /// Create new instance
+        ///
+        /// When checked it will always behave as being low.
```
dbrgn, line 154: I think "read" fits better here.

```suggestion
        /// When read it will always behave as being low.
```

src/digital.rs
```diff
@@ -123,3 +123,82 @@ pub trait InputPin {
...
+        pub fn new_low() -> Self {
+            DummyPin { _l: PhantomData }
+        }
+    }
+
+    impl DummyPin<level::High> {
+        /// Create new instance
+        ///
+        /// When checked it will always behave as being high.
```
dbrgn, line 163: I think "read" fits better here.

```suggestion
        /// When read it will always behave as being high.
```

dbrgn (dismissed):
Looks useful to me. Some docs-only nitpicking below.

ryankurte:
i wrote a _whole_ thing but i don't have the time / energy to edit it and the majority of the issues are already raised by the author in #141. the cliff-notes follow, i apologise if it comes off as brisk but it's what i can manage at the moment 🙃

- optional pins should be represented _in the driver API_
  - providing an unexpected `dummypin` to a driver violates the driver type constraints
  - designing drivers in this manner precludes compile or run-time decisions about pin availability (eg. use of optional IRQs _if_ they are available)
  - in pretty much all existing drivers, an application providing a dummy pin would break the driver in unpredictable and difficult to debug ways, especially where the CS pin is involved
- this stated issue already be expressed in a type-safe manner
  - via `Option<T>` (as detailed in #141, though still requiring a dummy type)
  - via trait defaults and specialisation over type constraints
  - both approaches (as with all the driver-author-tricks) really need to be documented

i am uncomfortable with the current PR as it would appear to me that the issues with this approach raised in #141 are not adequately addressed, and that there are better mechanisms to represent this without providing a blessed approach to functionally violate the driver type constraints. i would _really_ like to see more discussion of the possible approaches and limitations of each in #141 before landing this (or similar), but my current feeling is:

- if the issue is providing a type for use instantiating drivers that take an `Option<T>` there is no need for the dummy pin type to be instantiable and calling any methods could cause a panic
  - this seems to me to address #141, be less surprising, and brings the scope back to _application_ use rather than impacting driver authors (who see only `Some` or `None`)
  - i think this would be a good candidate for inclusion here
- if there is a _clear_ reason to need this to be instantiable, then it is expressly being provided to violate type constraints which imo should be justified and treated as such
  - construction should be marked as `unsafe`, documentation should be full of warnings and should not suggest usage as except for _very specific_ situations this will break the driver
  - this is unfortunately incompatible with causing a fault if it were used in an `Option<T>` then actually called :-(
  - i am unsure whether this would be better located here or in a third party crate

therealprof:
We discussed this controversely in todays meeting and opinons are wildly varying. It doesn't seem like we can easily find a solution which pleases everyone. Unnominating for now but we need to find a good way to resolve such deadlocks in the future.

eldruin:
Alright, I put this implementation into its own crate: [dummy-pin](https://crates.io/crates/dummy-pin) so that people that are interested can use it (compatible with `embedded-hal` `0.2` as of version `0.1`).
I would suggest moving the discussion about a "blessed" approach back to #141. Sorry for the annoyance.
@ryankurte Would you be able to put your thoughts there as well?

PR #239, opened 2020-07-22
Add try_set_state method for OutputPin

### LINKED ISSUE #200: OutputPin trait usability -- add set_value
Right now the v2 ouput pin looks like this:

```
/// Single digital push-pull output pin
pub trait OutputPin {
    /// Error type
    type Error;

    /// Drives the pin low
    ///
    /// *NOTE* the actual electrical state of the pin may not actually be low, e.g. due to external
    /// electrical sources
    fn set_low(&mut self) -> Result<(), Self::Error>;

    /// Drives the pin high
    ///
    /// *NOTE* the actual electrical state of the pin may not actually be high, e.g. due to external
    /// electrical sources
    fn set_high(&mut self) -> Result<(), Self::Error>;
}
```

This assumes 2 methods called set_low and set_high are how people want to interact with it. I think it might be nice to offer some flexibility of also having a set or set_value method that accepts a boolean.

The reason I think that this might be helpful is that when using the OutputPin, I nearly always already have a bool, so I always write the following code:
```
if val_to_set {x.set_high()?;} else {x.set_low()?;}
```

And this is probably affecting a lot of users.

Maybe this issues has already been beaten to death, in which case, I am sorry. I looked and could not find any other discussion on this.

For the InputPin, there is no need for ```get_value()```, as ```is_high()``` is essentially ```get_value()```, and ```is_low()``` essentially returns the invert of the value.

Any concerns with this idea? If no, I could put in a pull request.

eldruin:
This is an implementation of #200 to gather some opinions and so we can either accept it or close the issue.
This was earlier discussed at #44.

I added a conversion from `bool` following the usual convention as well as an `ops::Not` implementation as suggested in #200, which seemed appropriate.

I also added a default implementation for the `try_set_state` method. This bears the question whether a default implementation for `try_set_high()` / `try_set_low()` by using `try_set_state()` would be useful, so that potential implementors can choose to implement less methods.

It should be noted that adding a default implementation for all 3 methods has the somewhat amusing property of generating an endless loop if none is overwritten.

Closes #200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - 10-bit addressing mode for I2C traits.
+- `try_set_state` method for `OutputPin` using an input `PinState` value.
 
 ### Changed
 
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -1,5 +1,45 @@
 //! Digital I/O
 
+use core::{convert::From, ops::Not};
+
+/// Digital output pin state
+///
+/// Conversion from `bool` and logical negation are also implemented
+/// for this type.
+/// ```rust
+/// # use embedded_hal::digital::PinState;
+/// let state = PinState::from(false);
+/// assert_eq!(state, PinState::Low);
+/// assert_eq!(!state, PinState::High);
+/// ```
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum PinState {
+    /// Low pin state
+    Low,
+    /// High pin state
+    High,
+}
+
+impl From<bool> for PinState {
+    fn from(value: bool) -> Self {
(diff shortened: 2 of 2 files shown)


eldruin:
Seems like rust-highfive has not run here
r? therealprof

ryankurte:
idea seems good to me!

> It should be noted that adding a default implementation for all 3 methods has the somewhat amusing property of generating an endless loop if none is overwritten.

definitely think we only want to default impl _one_ of these (probably `try_set_state`) because this situation is, v cursed to come across.

eldruin:
@therealprof What do you think about this?

therealprof (approved):
LGTM

eldruin:
Seems we all agree.

@bors merge r=therealprof

eldruin:
Seems we all agree.

bors r=therealprof

PR #241, opened 2020-08-01
Storage Traits for on and off board storage

### LINKED ISSUE #28: Thoughts on Flash API?
Do you mind taking a quick look at https://github.com/idubrov/stm32-hal/blob/master/src/flash.rs to see if Flash trait I created is a good candidate to be included into this embedded-hal crate?

It appears to me that the trait itself could go into embedded-hal and implementation (line #97 and below) would go into device-specific crate (which in my case is stm32f103xx, but there is not stm32f103xx-hal yet, as far as I can see).

I also have an EEPROM emulation crate (https://github.com/idubrov/eeprom) which uses Flash trait, but, I guess, it should stay as a separate "driver" crate, which will depend on embedded-hal, right?

tl8roy:
Closes #28
These are a set of traits that can be used for on and off board storage.

Each function is split into its own trait to deal with the different types of memory (Particularly Read Only vs Read/Write)

MCU Flash impl:
https://github.com/tl8roy/alt-stm32f30x-hal/blob/master/src/storage.rs

I2C EEPROM impl:
https://github.com/tl8roy/eeprom24x-rs/blob/master/src/eeprom24x.rs

SPI PSRAM implementation:
https://github.com/tl8roy/esp_psram
This is a completely new implementation using SPI.

Feedback has come from #28, https://github.com/diondokter/device-driver/issues/1 and twitter https://twitter.com/tl8roy/status/1287254634512310272

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,6 +1,6 @@
 //! Storage
-//! The Read and Write traits are seperate to allow for Read Only Memory as well as Read and Write
-//! Example implementations include:
+/// The Read and Write traits are seperate to allow for Read Only Memory as well as Read and Write
+/// Example implementations include:
 
 use nb;
 
@@ -12,7 +12,7 @@ pub struct AddressOffset<U>(U);
 use core::ops::Add;
 
 /// Implement add for the Address and AddressOffset Types
-impl Add for Address<U> {
+impl<U> Add<AddressOffset<U>> for Address<U> {
     type Output = Self;
 
     fn add(self, other: AddressOffset<U>) -> Self {
@@ -96,9 +96,15 @@ pub trait ErasePage<U> {
     /// Erase the page of memory
     /// For flash devices, this sets the whole page to 0xFF
     /// Implementations should mask the address as required to get the page to erase
-    fn try_erase(&mut self, page: Page<U>) -> nb::Result<(), Self::Error>;
+    fn try_erase_page(&mut self, page: Page<U>) -> nb::Result<(), Self::Error>;
+
+    /// Erase the page of memory at the address. Note: The only valid address is the start of the page
+    /// For flash devices, this sets the whole page to 0xFF
+    /// Implementations should mask the address as required to get the page to erase
+    fn try_erase_address(&mut self, address: Address<U>) -> nb::Result<(), Self::Error>;
 }
 
+/// This trait allows for checking that data can fit before writing to the device. As some devices have limits on writing accross pages, the page size is also included
  pub trait StorageSize<Word,U> {
     /// An enumeration of Storage errors
     type Error;
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,0 +1,79 @@
+//! Storage
+//! The Read and Write traits are seperate to allow for Read Only Memory as well as Read and Write
+//! Example implementations include:
+
+use nb;
+
+/// Read a single word from the device
+/// `Word` type allows any word size to be used
+pub trait SingleRead<Word> {
+    /// An enumeration of Storage errors
+    type Error;
+
+    /// Reads the word stored at the address
+    /// ```
+    /// pub fn try_read(&mut self, address: usize) -> nb::Result<u8, Self::Error>
+    ///     let address = address as *const _;
+    ///     unsafe {
+    ///         Ok(core::slice::from_raw_parts::<'static, u8>(address,length)) 
+    ///     }
+    /// ```
+    fn try_read(&mut self, address: usize) -> nb::Result<Word, Self::Error>;
+
+}
+
+/// Write a single word to the device
+/// `Word` type allows any word size to be used
+pub trait SingleWrite<Word> {
+    /// An enumeration of Storage errors
+    type Error;
+
+    /// Writes the word to the address
+    fn try_write(&mut self, address: usize, word: Word) -> nb::Result<(), Self::Error>;
+
+}
+
+/// Read multiple bytes from the device
+/// Intended to be used for when there is a optimized method of reading multiple bytes
+/// 
+/// Iterating over the slice is a valid method to ```impl``` this trait
+pub trait MultiRead<Word> {
+    /// An enumeration of Storage errors
+    type Error;
+
+    /// Reads the words stored at the address to fill the buffer
+    /// ```
+    /// pub fn try_read_slice(&mut self, address: usize,  buf: &mut [Word]) -> nb::Result<Word, Self::Error>
+    ///     let address = address as *const _;
+    ///     unsafe {
+    ///         buf = core::slice::from_raw_parts::<'static, Word>(address,buf.len())
+    ///     }
+    ///     
+    ///     Ok() 
+    /// }
+    /// ```
+    fn try_read_slice(&mut self, address: usize, buf: &mut [Word]) -> nb::Result<(), Self::Error>;
+}
+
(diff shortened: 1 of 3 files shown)


ryankurte:
hey thanks for the PR, this looks like a super useful abstraction! (and i needed it this week 😂)

i am not completely sure whether this is the right level to include in the embedded-hal or not, though it would be great to have a standard flash API for HAL implementations which is a serious point in favour. as it is a reasonably complex abstraction i suspect that there remains some room for experimentation (and some changes may occur during this), so i _think_ my preference at this time would be to create a `flash-hal` crate to build out and evaluate this api prior to either adopting it into the WG or merging it into embedded-hal.

@rust-embedded/hal how do y'all feel like this, and is it worth a discussion at the next meeting about formalising the experimental-crate approach a bit more?

a few miscellaneous thoughts:

- is there an advantage to an API for page / flash sizes rather than associated constants? i guess flash size might be runtime detected but page size is probably not?
- how would you implement multiple erase methods (often devices support chip, page, half-page, and sector erase)?
- is it important to communicate the _erase state_? in the comments you mention setting the memory to 0, however, this depends on the underlying technology and _may_ impact the api consumer
- is it expected that the flash-hal will fail when writing across some boundaries and thus this must be handled by the application, or that the driver will manage this? it is worth noting that this may not be a _page_ boundary (many chips i have used can only write 255 bytes at a time as well as the standard page boundary limit). it seems to me the latter would be a neater abstraction for consumers?
- is it worth having a mechanism for locking/unlocking memory as is required for writing to most internal flash or is that outside the scope of this abstraction? (and is it important to signal that writing to internal flash is going to stop the processor doing anything or is that a hal implementation concern).

tl8roy:
> 
> 
> hey thanks for the PR, this looks like a super useful abstraction! (and i needed it this week 😂)
> 
> i am not completely sure whether this is the right level to include in the embedded-hal or not, though it would be great to have a standard flash API for HAL implementations which is a serious point in favour. as it is a reasonably complex abstraction i suspect that there remains some room for experimentation (and some changes may occur during this), so i _think_ my preference at this time would be to create a `flash-hal` crate to build out and evaluate this api prior to either adopting it into the WG or merging it into embedded-hal.

While the inspiration is that issue, the goal is absolutely a generic storage traits that aren't flash specific.

src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// For Flash storage, the write functions can't set a bit to 1.
+///
+/// For non flash devices, this trait is not required, but it can be used to erase data as recommended by the device (EG set all fields to 0).
+pub trait ErasePage<U> {
+    /// An enumeration of Storage errors
+    type Error;
+
+    /// Erase the page of memory
+    /// For flash devices, this sets the whole page to 0xFF
```
eldruin, line 112: I also think the reset value is probably better left to implementations. Anything written in the documentation of the trait would be part of a contract. (also below)

src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Allow for checking that data can fit before writing to the device.
+///
+/// As some devices have limits on writing accross pages, the page size is also included.
+pub trait StorageSize<Word, U> {
+    /// An enumeration of Storage errors
+    type Error;
+
+    /// Returns the start address and the maximum size that can be stored by the device
+    fn try_total_size(&mut self) -> nb::Result<(Address<U>, AddressOffset<U>), Self::Error>;
```
eldruin, line 130: I think two different methods may be better, since it would be fairly easy to make a mistake.

src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Address represents an unsigned int. This allows for devices that have bigger or smaller address spaces than the host.
+pub struct Address<U>(pub U);
+/// Address Offset represents an unsigned int that is used as an optional offset from the base address.
+pub struct AddressOffset<U>(pub U);
+
+use core::ops::Add;
+
+/// Implement add for the Address and AddressOffset Types.
+impl<'a, 'b, U> Add<&'b AddressOffset<U>> for &'a Address<U>
```
eldruin, line 14: Subtraction of an offset from an address is probably also interesting. And if we already at it, maybe addition and subtraction of address offsets. However I start wondering where to stop :) and then if it makes sense to include these implementations at all since the user already needs to deal with the inner values.

src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+{
+    type Output = Address<U>;
+
+    fn add(self, other: &'b AddressOffset<U>) -> Address<U> {
+        Address(self.0 + other.0)
+    }
+}
+
+/// Page represents an unsigned int that is a Page ID in the device memory space.
+pub struct Page<U>(U);
```
eldruin, line 26: I am wondering about a conversion from `Page` to `Address` but I guess the best way would be adding a `From` implementation and that is something that needs to be done in the implementation and not nicely enforcible here.

tl8roy:
I think the SPI flash chip library is non blocking. I'll have a crack at that.

MathiasKoch:
I think it would make sense to make `Address`, `AddressOffset` and `Page` implement/derive a bunch of basic traits.
Of the top of my head it would make sense with something like `Clone, Copy, Debug, PartialEq, PartialOrd`?

Also i think it makes sense to be able to sub two addresses? very common way of getting a length?
and an `Address % AddressOffset` might also give value? I know the last couple here are easy to do myself, as the inner is public, but still applicable i think.

Sympatron:
What do you think about starting with a new crate to get this going? I think something like `embedded-storage` would be fitting. That way people could start using it immediately.

After some time it could still be merging into embedded-hal.

MathiasKoch:
@Sympatron I think that is the right approach! Currently i am forcing all of my crates into separate branches just to use a storage branch here, and it doesn't quite make sense. 

Maybe it would make sense to get @eldruin to create it in `embedded-rust-community`? I think that would be a great fit!

tl8roy:
I'm happy to support which ever solution works best for the community. I do have a bit of time this weekend.

eldruin:
`embedded-storage` sounds good to me. As previously discussed, it is probable that this takes a while to get ready for `embedded-hal`. It should be noted as well that there are two proposals for storage traits: this one and also #248.
Maybe both can live within `embedded-storage` until it is settled which one is best or so.
Adding it to the `rust-embedded-community` might be possible. I am preparing an overhaul of the org's mission.

MathiasKoch:
@tl8roy @eldruin 

I have created this: https://github.com/BlackbirdHQ/embedded-storage in an attempt to bring this forward, knowingly that it does not address all our issues, but at least it is a starting point for discussions.. It is currently a 1:1 copy of #248 

I would love to see it become part of the embedded-rust-community, but had to start it somewhere else..? @eldruin If you would be okay with it, i can transfer it?

eldruin:
@MathiasKoch sure thing. It would be great if this triggers a development impulse like with `embedded-nal`. If you run into problems transferring it, let me know.

MathiasKoch:
Awesome!

Let me introduce: https://github.com/rust-embedded-community/embedded-storage 

@eldruin can you close #28 with a reference to the new repo?
Then i will close #248, leaving only this PR that @tl8roy could maybe move to embedded-storage for further discussion?

tl8roy:
Now moved to https://github.com/rust-embedded-community/embedded-storage/pull/8 so closing off for now.

PR #242, opened 2020-08-04
Make ADC Channel trait use a stateful method to get the IDs

### LINKED ISSUE #110: ADC API  does not allow non-unit channels
I'm just implementing the unproven ADC traits for RIOT and found an issue with Channel::channel:

As the channel() function does not even take a &self reference, there can't be implementations of it that are not fully contained in the type (ie. only really works for zero-sized types). Now the zero-sized types are a major plus of embedded Rust, but this signature precludes the use of eg. numeric channels as would be convenient when wrapping an operating system's ADCs.

I suggest adding a &self or &mut self argument to the .channel() function to make it usable as a method.

I don't understand the rationale behind there not being a self argument in or the existence of the channel method at all there in the first place, so I can't tell what it breaks. The example in the OneShot trait would need to say `_pin.channel()` instead of `PIN::channel()` -- but also there I don't see the rationale for not doing that in the first place.

(For the RIOT wrappers, until I can publish the code I'd like to point to, things look like this:

    pub struct ADCLine(i32); // You'd get that from the OS on the long run; right now the int is pub
    impl Channel<ADC> for ADCLine {
        type ID = i32;
        fn channel() -> i32 { unimplemented!() }
    }
    
    impl OneShot<ADC, i32, ADCLine> for ADC {
        fn read(&mut self, pin: &mut ADCLine) {
            unsafe { riot_sys::adc_init(pin.0) };
            ...
        }
    }

)

eldruin:
Closes #110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   has been renamed `try_write_iter` for consistency.
 - Updated `nb` dependency to version `1`.
 - The watchdog API now uses move semantics. See [PR](https://github.com/rust-embedded/embedded-hal/pull/222).
+- The ADC `Channel` trait now uses a stateful method to get the IDs.
 
 ## [v1.0.0-alpha.1] - 2020-06-16
 
@@ -39,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   in trait implementations where methods cannot fail.
 - A new [process](https://github.com/rust-embedded/embedded-hal#how-to-add-a-new-trait)
   has been adopted for the addition of traits to the embedded-hal.
+- The ADC `Channel` trait now uses a constant to represent the IDs.
 - The minimum supported Rust version is 1.35 due to [this issue](https://github.com/rust-lang/rust/issues/54973).
 
 ## [v0.2.3] - 2019-05-09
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -39,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   in trait implementations where methods cannot fail.
 - A new [process](https://github.com/rust-embedded/embedded-hal#how-to-add-a-new-trait)
   has been adopted for the addition of traits to the embedded-hal.
+- The ADC `Channel` trait now uses a constant to represent the IDs.
 - The minimum supported Rust version is 1.35 due to [this issue](https://github.com/rust-lang/rust/issues/54973).
```
david-sawatzke, line 44: This can now also be removed, since the associated feature isn't used anymore, right?
```suggestion

```

eldruin: The `-alpha.1` version would still need 1.35 but after merging this PR we could decrease the MSRV again for the next release.

eldruin:
As highlighted by @david-sawatzke, with this PR we do not need Rust 1.35. anymore.
Should we decrease the MSRV back to 1.31?
What do you think @rust-embedded/hal?

therealprof:
I'd rather not. We're already very conservative with our MSRV and that seems like a rather pointless change, in that it will create work which at best has a temporary use (i.e. until we have to bump the MSRV for something else) and no one so far has complained about 1.35 being a problem to begin with.

eldruin:
Alright, I was also thinking along those lines.

ryankurte (approved):
lgtm, @therealprof anything outstanding?

therealprof (approved):
LGTM, too.

bors r+

PR #245, opened 2020-08-29
Add ManagedCS marker trait

ryankurte:
Per #180 this PR introduces a `ManagedCS` marker trait to indicate that the `CS` line on an SPI peripheral is managed by the underlying driver.  This is required to address unsoundness due to bus-sharing without explicit management of the CS pin per https://github.com/Rahix/shared-bus/issues/8, and paired with #191 this will significantly improve the ergonomics of writing SPI-based drivers by no longer requiring every implementation to re-implement CS managed transactions and error handling (or to use a third party crate to do this).

I have been using this approach for some time in the [embedded-spi](https://github.com/ryankurte/rust-embedded-spi) crate, for the purposes of evaluation the intent is to refactor both `embedded-spi` and `shared-bus` to utilise this marker, as well as some dependent drivers (though this is for me difficult due to a cross dependency on #191).

## Questions

- Adding this marker introduces a need for a wrapper type that consumes a SPI object and a pin to provide a [`ManagedCS` interface](https://github.com/ryankurte/rust-embedded-spi/blob/master/src/wrapper.rs#L94), should this live here (in the `embedded-hal`) or in a third party crate?
- Initially I proposed this be called `ChipSelect`, another option is `ManagedChipSelect`, and other suggestions are welcome, however, please avoid adding noise to the discussion of this issue / implementation with minor naming concerns where possible.

## Demonstration

- [x] add expectations for `blocking::spi::*` trait use
- [x] add `ManagedCS` impl to [shared-bus](https://github.com/Rahix/shared-bus)
- [ ] ~alias `embedded-spi::ManagedChipSelect` to `embedded_hal::blocking::spi::ManagedCs` to update all `embedded` dependents~
- [x] implement basic `SpiWithCS` wrapper in whatever crate is appropriate (and update embedded-hal docs)
- [ ] update a couple of drivers to demonstrate the use of the marker traits

~I am assigning this to the v1.0 milestone though it is technically _non breaking_ as it does effect all SPI drivers, and is somewhat critical to support use of `shared-bus` with SPI peripherals (which is _no longer possible_ due to unsoundness with the previous impl)~

~update: no longer part of 1.0 to avoid blocking, though hopefully to be landed shortly after this~

cc. @rahix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   implement these `Error` traits, which implies providing a conversion to a common
   set of error kinds. Generic drivers using these interfaces can then convert the errors
   to this common set to act upon them.
+- `ManagedCS` trait for SPI, signals that CS is managed by the driver to allow
+  shared use of an SPI bus, as well as an `SpiWithCs` wrapper implementing this
+  over an Spi and Pin implementation.
 
 ### Removed
 - Removed `DelayMs` in favor of `DelayUs` with `u32` as type for clarity.
--- a/src/spi/blocking.rs
+++ b/src/spi/blocking.rs
@@ -1,9 +1,4 @@
 //! Blocking SPI API
-//!
-//! In some cases it's possible to implement these blocking traits on top of one of the core HAL
-//! traits. To save boilerplate when that's the case a `Default` marker trait may be provided.
-//! Implementing that marker trait will opt in your type into a blanket implementation.
-
 /// Blocking transfer
 pub trait Transfer<W> {
     /// Error type
@@ -88,5 +83,149 @@ impl<T: Transactional<W>, W: 'static> Transactional<W> for &mut T {
 
     fn exec<'a>(&mut self, operations: &mut [Operation<'a, W>]) -> Result<(), Self::Error> {
         T::exec(self, operations)
+    }
+}
+
+/// Provides SpiWithCS wrapper using for spi::* types combined with an OutputPin
+pub use spi_with_cs::{SpiWithCs, SpiWithCsError};
+mod spi_with_cs {
+
+    use core::fmt::Debug;
+
+    use super::{Transfer, Write, WriteIter};
+    use crate::digital::blocking::OutputPin;
+    use crate::spi::{ErrorKind, ManagedChipSelect};
+
+    /// [`SpiWithCs`] wraps an SPI implementation with Chip Select (CS)
+    /// pin management for exclusive (non-shared) use.
+    /// For sharing SPI between peripherals, see [shared-bus](https://crates.io/crates/shared-bus)
+    pub struct SpiWithCs<Spi, Pin> {
+        spi: Spi,
+        cs: Pin,
+    }
+
+    /// Wrapper for errors returned by [`SpiWithCs`]
+    #[derive(Clone, Debug, PartialEq)]
+    pub enum SpiWithCsError<SpiError, PinError> {
+        /// Underlying SPI communication error
+        Spi(SpiError),
+        /// Underlying chip-select pin state setting error
+        Pin(PinError),
+    }
+
+    /// Implement [`crate::spi::Error`] for wrapped types
+    impl<SpiError, PinError> crate::spi::Error for SpiWithCsError<SpiError, PinError>
+    where
+        SpiError: crate::spi::Error + Debug,
+        PinError: Debug,
+    {
+        fn kind(&self) -> ErrorKind {
+            match self {
+                SpiWithCsError::Spi(e) => e.kind(),
+                SpiWithCsError::Pin(_e) => ErrorKind::Other,
+            }
+        }
+    }
+
+    /// [`ManagedChipSelect`] marker trait indicates Chip Select (CS) pin management is automatic
+    impl<Spi, Pin> ManagedChipSelect for SpiWithCs<Spi, Pin> {}
+
+    impl<Spi, Pin> SpiWithCs<Spi, Pin>
+    where
+        Pin: OutputPin,
+    {
+        /// Create a new SpiWithCS wrapper with the provided Spi and Pin
+        pub fn new(spi: Spi, cs: Pin) -> Self {
+            Self { spi, cs }
+        }
+
+        /// Fetch references to the inner Spi and Pin types.
+        /// Note that using these directly will violate the `ManagedChipSelect` constraint.
+        pub fn inner(&mut self) -> (&mut Spi, &mut Pin) {
+            (&mut self.spi, &mut self.cs)
+        }
+
+        /// Destroy the SpiWithCs wrapper, returning the bus and pin objects
+        pub fn destroy(self) -> (Spi, Pin) {
+            (self.spi, self.cs)
+        }
+    }
+
+    impl<Spi, Pin, W> Transfer<W> for SpiWithCs<Spi, Pin>
+    where
+        Spi: Transfer<W>,
+        Pin: OutputPin,
+        <Spi as Transfer<W>>::Error: Debug,
+        <Pin as OutputPin>::Error: Debug,
+    {
+        type Error = SpiWithCsError<<Spi as Transfer<W>>::Error, <Pin as OutputPin>::Error>;
+
+        /// Attempt an SPI transfer with automated CS assert/deassert
+        fn transfer<'w>(&mut self, data: &'w mut [W]) -> Result<(), Self::Error> {
+            // First assert CS
+            self.cs.set_low().map_err(SpiWithCsError::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_result = self.spi.transfer(data).map_err(SpiWithCsError::Spi);
+
+            // Deassert CS
+            self.cs.set_high().map_err(SpiWithCsError::Pin)?;
+
+            // Return failures
+            spi_result
+        }
+    }
+
+    impl<Spi, Pin, W> Write<W> for SpiWithCs<Spi, Pin>
+    where
+        Spi: Write<W>,
+        Pin: OutputPin,
+        <Spi as Write<W>>::Error: Debug,
+        <Pin as OutputPin>::Error: Debug,
+    {
+        type Error = SpiWithCsError<<Spi as Write<W>>::Error, <Pin as OutputPin>::Error>;
+
+        /// Attempt an SPI write with automated CS assert/deassert
+        fn write<'w>(&mut self, data: &'w [W]) -> Result<(), Self::Error> {
+            // First assert CS
+            self.cs.set_low().map_err(SpiWithCsError::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_result = self.spi.write(data).map_err(SpiWithCsError::Spi);
+
+            // Deassert CS
+            self.cs.set_high().map_err(SpiWithCsError::Pin)?;
+
+            // Return failures
+            spi_result
+        }
+    }
+
+    impl<Spi, Pin, W> WriteIter<W> for SpiWithCs<Spi, Pin>
+    where
+        Spi: WriteIter<W>,
+        Pin: OutputPin,
+        <Spi as WriteIter<W>>::Error: Debug,
+        <Pin as OutputPin>::Error: Debug,
+    {
+        type Error = SpiWithCsError<<Spi as WriteIter<W>>::Error, <Pin as OutputPin>::Error>;
+
+        /// Attempt an SPI write_iter with automated CS assert/deassert
+        fn write_iter<WI>(&mut self, words: WI) -> Result<(), Self::Error>
+        where
+            WI: IntoIterator<Item = W>,
+        {
+            // First assert CS
+            self.cs.set_low().map_err(SpiWithCsError::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_result = self.spi.write_iter(words).map_err(SpiWithCsError::Spi);
+
+            // Deassert CS
+            self.cs.set_high().map_err(SpiWithCsError::Pin)?;
+
+            // Return failures
+            spi_result
+        }
     }
 }
--- a/src/spi/mod.rs
+++ b/src/spi/mod.rs
@@ -107,3 +107,15 @@ impl core::fmt::Display for ErrorKind {
         }
     }
 }
+
+/// ManagedChipSelect marker trait indicates the CS pin is managed by the underlying driver.
+///
+/// This specifies that `spi` operations will be grouped:
+/// preceded by asserting the CS pin, and followed by
+/// de-asserting the CS pin, prior to returning from the method.
+///
+/// This is important for shared bus access to ensure that only one CS can be asserted at a given time.
+/// To support shared use, drivers should require this (and not manage their own CS pins).
+///
+/// For usage examples see: [`crate::spi::blocking::SpiWithCs`].
+pub trait ManagedChipSelect {}


eldruin:
A managed CS addition sounds good to me in principle.
- Whether `ManagedChipSelect` or `ManagedCS` I have no strong opinions. `ManagedChipSelect` is self-descriptive, which is a good thing. I think an argument would be looking at how long types get when using it.
- I would say `SpiWithCS` would be central enough to be included directly here.
- I understand the motivation to release this in 1.0 but I would rather not block that with anything else and just get it out the door as soon as possible. **Shortly** after that we can do a 1.1 release with this PR, transactional SPI/I2C, I2S and so on.
- How would you address device timing requirements between setting the CS pin low and starting SPI communication?

I can properly review once there is more in this PR.

therealprof:
@eldruin Thanks for your write-up, so I can save my time saying pretty much the same thing. 😅

ryankurte:
agreed on pretty much all points, to be clear i only meant the `CS` part from `embedded-spi`.

> I understand the motivation to release this in 1.0 but I would rather not block that with anything else and just get it out the door as soon as possible. Shortly after that we can do a 1.1 release with this PR, transactional SPI/I2C, I2S or whatever is ready.

my concern with this is that since `shared-bus` no longer contains an SPI proxy, bumping the `embedded-hal` and related `shared-bus` impl in use is going to break any proxied SPI with no useful recourse, which is not going to be a great experience on top of the breaking HAL API changes. maybe re-adding the SPI proxy with some big soundness warnings would be viable to mitigate this? (cc. @rahix)

> How would you address device timing requirements between setting the CS pin low and starting SPI communication? This might be a good time to think about this.

good call. in my experience (where the driver manages CS) this is _usually_ this is implemented by the underlying driver, in this case `SpiWithCS` will be a lightweight wrapper with no specific timing consideration, if manual timing is required it is reasonably trivial to implement a third party `SpiTimedWithCS` or equivalent, and where the drivers handle this it's configuration-time which i think is outside the scope for the hal.

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+        PinError: Debug,
+    {
+        /// Create a new SpiWithCS wrapper with the provided Spi and Pin
+        pub fn new(spi: Spi, cs: Pin) -> Self {
+            Self{spi, cs, _spi_err: PhantomData, _pin_err: PhantomData}
+        }
+
+        /// Fetch references to the inner Spi and Pin types.
+        /// Note that using these directly will violate the `ManagedCs` constraint.
+        pub fn inner(&mut self) -> (&mut Spi, &mut Pin) {
```
eldruin, line 162: Wouldn't this also need a `destroy` or similar where the actual instances are returned?

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+            (&mut self.spi, &mut self.cs)
+        }
+    }
+
+    impl <Spi, SpiError, Pin, PinError> Transfer<u8> for SpiWithCs<Spi, SpiError, Pin, PinError>
```
eldruin: This default implementation should be in a submodule so that it is explicitly imported. Otherwise it would not be possible for a third-party to offer their own implementation of the traits as it would result in conflicting implementations. e.g. to handle `CS` assertion timing requirements.

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+    {
+        type Error = SpiWithCsErr<SpiError, PinError>;
+
+        /// Attempt an SPI transfer with automated CS assert/deassert
+        fn try_transfer<'w>(&mut self, data: &'w mut [u8]) -> Result<&'w [u8], Self::Error> {
+            // First assert CS
+            self.cs.try_set_low().map_err(SpiWithCsErr::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_res = self.spi.try_transfer(data).map_err(SpiWithCsErr::Spi);
```
eldruin, line 182: Can we call this just `result`?
```suggestion
            let result = self.spi.try_transfer(data).map_err(SpiWithCsErr::Spi);
```

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+        spi: Spi,
+        cs: Pin,
+
+        _spi_err: PhantomData<SpiError>,
+        _pin_err: PhantomData<PinError>,
+    }
+
+    /// SpiWithCsErr provies an error numberation over generic Spi and Pin variants
+    #[derive(Clone, Debug, PartialEq)]
+    pub enum SpiWithCsErr<SpiError, PinError> {
```
eldruin, line 134: Some bikeshedding: How about `SpiWithCsError`?

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+    /// For sharing SPI between peripherals, see [shared-bus]()
+    pub struct SpiWithCs<Spi, SpiError, Pin, PinError> {
+        spi: Spi,
+        cs: Pin,
+
+        _spi_err: PhantomData<SpiError>,
+        _pin_err: PhantomData<PinError>,
+    }
+
+    /// SpiWithCsErr provies an error numberation over generic Spi and Pin variants
```
eldruin, line 132:
```suggestion
    /// Underlying causes for errors. Either SPI communication or CS pin assertion error
```

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+    #[derive(Clone, Debug, PartialEq)]
+    pub enum SpiWithCsErr<SpiError, PinError> {
+        /// Underlying SPI error
```
eldruin, line 135:
```suggestion
        /// Underlying SPI communication error
```

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+        Spi(SpiError),
+        /// Underlying Pin error
```
eldruin, line 137:
```suggestion
        /// Underlying chip-select pin state setting error
```

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+
+    use core::marker::PhantomData;
+    use core::fmt::Debug;
+
+    use crate::digital::OutputPin;
+    use super::*;
+
+    /// SpiWithCS wraps an blocking::spi* implementation with Chip Select (CS)
+    /// pin management.
+    /// For sharing SPI between peripherals, see [shared-bus]()
```
eldruin, line 123:
```suggestion
    /// For sharing SPI between peripherals, see [shared-bus](https://crates.io/crates/shared-bus)
```

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+    pub struct SpiWithCs<Spi, SpiError, Pin, PinError> {
+        spi: Spi,
+        cs: Pin,
+
+        _spi_err: PhantomData<SpiError>,
+        _pin_err: PhantomData<PinError>,
```
eldruin, line 129: Is this necessary?

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
+
+            // Deassert CS
+            self.cs.try_set_high().map_err(SpiWithCsErr::Pin)?;
+
+            // Return failures
+            spi_res
+        }
+    }
+
+    impl <Spi, SpiError, Pin, PinError> Write<u8> for SpiWithCs<Spi, SpiError, Pin, PinError>
```
eldruin: Could all these implementations be made generic over the word type?

src/blocking/spi.rs
```diff
@@ -108,3 +108,134 @@ pub mod write_iter {
...
 }
+
+/// Provides SpiWithCS wrapper around an spi::* and OutputPin impl
+pub mod spi_with_cs {
+
+    use core::marker::PhantomData;
+    use core::fmt::Debug;
+
+    use crate::digital::OutputPin;
+    use super::*;
```
eldruin, line 118: Can we import them explicitly? it is probably just `Transfer`, `Write` and `WriteIter`, right?

eldruin (changes requested):
Great work, thank you!
- Another alternative would be skipping e-h 1.0 support for `shared-bus` and updating only on 1.1 (which would give us more pressure to release 1.1 :), although I understand it does not look good.
- Some more bikeshedding: It seems `ManagedCs` will not need to be written often, so I think `ManagedChipSelect` would be preferable.

Another (albeit far-fetched) argument for delaying stuff to 1.1 would be that we _could_ yank version 1.1 if something is terribly wrong with it but yanking 1.0 would be more problematic.

Rahix:
> my concern with this is that since shared-bus no longer contains an SPI proxy, bumping the embedded-hal and related shared-bus impl in use is going to break any proxied SPI with no useful recourse, which is not going to be a great experience on top of the breaking HAL API changes. maybe re-adding the SPI proxy with some big soundness warnings would be viable to mitigate this? (cc. @Rahix)

@ryankurte: Adding the unsound SPI proxy is definitely off the table.  We've seen it bite people before so it's safe to assume that it would happen again.  We need a different solution here ...

For the time being, we have the single-threaded SPI proxy which should be plenty for many use-cases already.  For anything more complex, I'm unsure ... I think the work from @ryan-summers for shared-bus RTIC support could be a viable workaround for the time being.  Essentially, it ensures safety by stuffing all bus-users into a struct and wrapping that struct in a mutex:

```rust
struct SpiDevices {
    pub device_a: DeviceA<SpiProxy<...>>,
    pub device_b: DeviceB<SpiProxy<...>>,
}

static SPI_DEVICES: Mutex<Option<SpiDevices>> = Mutex::new(None);

fn task_a() -> Result<..., ...> {
    ...

    SPI_DEVICES.lock(|s| {
        let device_a = &mut s.ok_or(Error)?.device_a;

        device_a.do_things();
    })?;
}

fn task_b() -> Result<..., ...> {
    ...

    SPI_DEVICES.lock(|s| {
        let device_b = &mut s.ok_or(Error)?.device_b;

        device_b.do_other_things();
    })?;
}
```

---

But of course, `ManagedChipSelect` or later a full transactional SPI interface would be the cleaner solution.

src/blocking/spi.rs
ryankurte: the associated errors? not _strictly_ but ime it makes the use of the objects a bit clearer having the types consistently bound to the object rather than able to vary by impl block.

src/blocking/spi.rs
ryankurte: is `Err` unclear given it's already part of rust nomenclature?

src/blocking/spi.rs
ryankurte: may be useful to have, not sure what to call it / destroy doesn't seem quite right to me?

src/blocking/spi.rs
ryankurte: is result more clear?

eldruin: I prefer whole words, yes, but feel free to dismiss this.

src/blocking/spi.rs
eldruin: It is clear, but all other error types we have here are called `Error` in full which is just 2 more letters. Anyway it is not very important.

src/blocking/spi.rs
ryankurte: i am not sure what you mean by this? it is not an opt-in default like the other trait defaults but and entirely separate container with CS, so, reuse is not expected?

For additional timing constraints one would need create a new container type (like `SpiWithCs` but `SpiWithTimedCs` or whatever and implement the traits on the new object, afaik trying to re-use both the container object and traits from this module would run afoul of the orphan rules (both type and trait would be external).

src/blocking/spi.rs
ryankurte: unfortunately i believe not, this runs afoul of another set of rust trait constraints. if there is demonstrated need in the future we can however implement these traits over different word sizes.

src/blocking/spi.rs
ryankurte: yep totally

src/blocking/spi.rs
eldruin: Ah, I see now that `ManagedCs` is not part of the `spi_with_cs` module. That is all I need. Sorry for the confusion.

src/blocking/spi.rs
ryankurte: fair and, renamed!

therealprof:
We discussed this in todays meeting due to the nominated flag but it's not quite clear what needs doing here. Seems like a good idea to have this to me. Can anyone summarize what we would need to discuss on the next meeting? Thanks.

ryankurte:
updated this PR with documentation around expectations and added an implementation to https://github.com/Rahix/shared-bus/pull/21, just need to update a driver to take advantage of this.

the updated expectation comes down to, if you expect operations to occur in a known order you need to use `Transactional`, and if you expect `CS` assertions to be exclusive you need to use `ManagedCs`.

in practice a bunch of drivers don't do this, primarily because `Transactional` didn't used to exist, but, it's pretty trivial to update (and i think the only way we get to safe shared SPI access).

last thing on the list is to update a driver / example to talk to a couple of SPI devices but, the behaviour is not something the compiler can check, so, this probably should be run but still doesn't serve as a demonstration of the negative...

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
 }
 
 /// Transactional trait allows multiple actions to be executed
-/// as part of a single SPI transaction
+/// as part of a single SPI transaction.
+///
+/// This API guarantees ordering, ensuring operations from
+/// different sources will not be interleaved on a shared bus.
+/// [ManagedCs]
```
eldruin, line 71: Clarify the reference to `ManagedCs` here.

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+    /// Underlying causes for errors. Either SPI communication or CS pin state setting error
+    #[derive(Clone, Debug, PartialEq)]
+    pub enum SpiWithCsError<SpiError, PinError> {
+        /// Underlying SPI communication error
+        Spi(SpiError),
+        /// Underlying chip-select pin state setting error
+        Pin(PinError),
+    }
+
+    /// ManagedCS marker trait indicates Chip Select management is automatic
```
eldruin, line 109:
```suggestion
    /// `ManagedCs` marker trait indicates chip-select management is automatic
```

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+    impl<Spi, SpiError, Pin, PinError> ManagedCs for SpiWithCs<Spi, SpiError, Pin, PinError> {}
+
+    impl<Spi, SpiError, Pin, PinError> SpiWithCs<Spi, SpiError, Pin, PinError>
+    where
+        Pin: crate::blocking::digital::OutputPin<Error = PinError>,
+        SpiError: Debug,
+        PinError: Debug,
+    {
+        /// Create a new SpiWithCS wrapper with the provided Spi and Pin
```
eldruin, line 118:
```suggestion
        /// Create a new `SpiWithCs` wrapper with the provided SPI bus and Pin
```

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+        pub fn new(spi: Spi, cs: Pin) -> Self {
+            Self {
+                spi,
+                cs,
+                _spi_err: PhantomData,
+                _pin_err: PhantomData,
+            }
+        }
+
+        /// Fetch references to the inner Spi and Pin types.
```
eldruin, line 128:
```suggestion
        /// Fetch references to the inner SPI bus and Pin types
        ///
```

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+        /// Note that using these directly will violate the `ManagedCs` constraint.
+        pub fn inner(&mut self) -> (&mut Spi, &mut Pin) {
+            (&mut self.spi, &mut self.cs)
+        }
+
+        /// Destroy the SpiWithCs wrapper, returning the bus and pin objects
```
eldruin, line 134:
```suggestion
        /// Destroy the `SpiWithCs` wrapper, returning the SPI bus and Pin objects
```

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+            // First assert CS
+            self.cs.set_low().map_err(SpiWithCsError::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_result = self.spi.transfer(data).map_err(SpiWithCsError::Spi);
+
+            // Deassert CS
+            self.cs.set_high().map_err(SpiWithCsError::Pin)?;
+
+            // Return failures
```
eldruin, line 160:
```suggestion
            // Return transfer result or failure
```

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+            // First assert CS
+            self.cs.set_low().map_err(SpiWithCsError::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_result = self.spi.write(data).map_err(SpiWithCsError::Spi);
+
+            // Deassert CS
+            self.cs.set_high().map_err(SpiWithCsError::Pin)?;
+
+            // Return failures
```
eldruin, line 185:
```suggestion
            // Return transfer result or failure
```

src/blocking/spi.rs
```diff
@@ -41,11 +64,154 @@ pub enum Operation<'a, W: 'static> {
...
+            // First assert CS
+            self.cs.set_low().map_err(SpiWithCsError::Pin)?;
+
+            // Attempt the transfer, storing the result for later
+            let spi_result = self.spi.write_iter(words).map_err(SpiWithCsError::Spi);
+
+            // Deassert CS
+            self.cs.set_high().map_err(SpiWithCsError::Pin)?;
+
+            // Return failures
```
eldruin, line 213:
```suggestion
            // Return transfer result or failure
```

eldruin (commented):
Great. I added some comments but it is looking good. Thanks!

eldruin:
> > How would you address device timing requirements between setting the CS pin low and starting SPI communication? This might be a good time to think about this.
> 
> good call. in my experience (where the driver manages CS) this is _usually_ this is implemented by the underlying driver, in this case `SpiWithCS` will be a lightweight wrapper with no specific timing consideration, if manual timing is required it is reasonably trivial to implement a third party `SpiTimedWithCS` or equivalent, and where the drivers handle this it's configuration-time which i think is outside the scope for the hal.

@therealprof: I had forgotten about it but I already asked about timing considerations here.
Do you have any further concerns with leaving this to the implementation?

I guess a crucial point would be that HALs provide both managed and non-managed CS SPI so that something like `SpiTimedWithCs`, or something more complex like cycling CS between operations if necessary can be implemented.

ryankurte:
> I guess a crucial point would be that HALs provide both managed and non-managed CS SPI so that something like SpiTimedWithCs, or something more complex like cycling CS between operations if necessary can be implemented.

i think like most of the HAL this addresses the core issue but, cannot solve everything. if you have to configure timing between transactions or after CS it's going to require either device configuration or wrapping and, as you suggest, you can always implement this either in hardware or as another layer.

given the generic `SpiWithCS` implementation i would only expect HALs to implement `ManagedCS` where the hardware (or OS) actually does manage the chip select?

almindor:
> > > How would you address device timing requirements between setting the CS pin low and starting SPI communication? This might be a good time to think about this.
> > 
> > 
> > good call. in my experience (where the driver manages CS) this is _usually_ this is implemented by the underlying driver, in this case `SpiWithCS` will be a lightweight wrapper with no specific timing consideration, if manual timing is required it is reasonably trivial to implement a third party `SpiTimedWithCS` or equivalent, and where the drivers handle this it's configuration-time which i think is outside the scope for the hal.
> 
> @therealprof: I had forgotten about it but I already asked about timing considerations here. Do you have any further concerns with leaving this to the implementation?
> 
> I guess a crucial point would be that HALs provide both managed and non-managed CS SPI so that something like `SpiTimedWithCs`, or something more complex like cycling CS between operations if necessary can be implemented.

These things are usually very specific when using HW. See for example [how it's done in the e310x-hal](https://github.com/riscv-rust/e310x-hal/blob/master/src/spi/config.rs)

We decided to expose this so users can configure the SPI in all possible ways. I don't think timings and delays should be considered in the trait abstraction level.

ryankurte:
tried to update this against master, i am not sure why all those other commits have been pulled in :-/

ryankurte:
closing in favour of #351

PR #248, opened 2020-09-18
Add storage ReadWrite trait, and helper structs, traits and iterators

### LINKED ISSUE #28: Thoughts on Flash API?

MathiasKoch:
I would like to propose an alternative solution to #28 .

There is currently one solution in #241, this is an alternative and simpler approach, highly inspired by the blog series by Cuervo in https://www.ecorax.net/as-above-so-below-1/ and https://www.ecorax.net/as-above-so-below-2/

The traits are implemented here (still working on the documentation): https://github.com/MathiasKoch/is25xp 
and here: https://github.com/MathiasKoch/stm32l4xx-hal/blob/factbird-mini-1.0/src/flash.rs

I think this makes it MUCH easier for driver authors to make use of such a trait, compared to the solution given in #241 with a total of 5 traits. Also the helper iterators makes it much easier to handle optimizing for NOR flashes where writing `1`s is not an option, without just defaulting to writing everything, as explained in above blog series. 

**EDIT:**
After having used this for a while, there are a couple of notes and pain points to be addressed/discussed.
1. I think the `try_erase` function is close to useless as it is currently, as it does only allow erasing the full memory range, as opposed to sections/regions/range. Any suggestions for a better approach? One possible could be to take a range (`fn try_erase(&mut self, range: (Address, Address))` or `fn try_erase(&mut self, from: Address, to: Address)`)?
2. It might make sense to split the trait into a `trait Read` and a `trait ReadWrite: Read`, to support read-only memory? I have never encountered write-only, so i guess `ReadWrite` would make perfectly sense?

Any feedback is welcommed!

Closes #28

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -697,6 +697,7 @@ pub mod pwm;
 pub mod qei;
 pub mod rng;
 pub mod serial;
+pub mod storage;
 pub mod spi;
 pub mod timer;
 pub mod watchdog;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -697,8 +697,8 @@ pub mod pwm;
 pub mod qei;
 pub mod rng;
 pub mod serial;
-pub mod storage;
 pub mod spi;
+pub mod storage;
 pub mod timer;
 pub mod watchdog;
 
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -141,6 +141,9 @@ pub trait ReadWrite {
     /// (start_addr, end_addr)
     fn range(&self) -> (Address, Address);
 
-    /// Erase the full storage range, clearing all data within `self.range()`.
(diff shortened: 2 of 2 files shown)


MathiasKoch:
I am closing this in favor of https://github.com/rust-embedded-community/embedded-storage that we started to get things right and mature.

PR #250, opened 2020-10-02
Make use of nb explicit

eldruin:


therealprof (approved):
Makes sense, should we backport this?

bors r+

eldruin:
Backporting would not be possible as v0.2.x is still in Rust 2015.

PR #251, opened 2020-10-02
Update reference implementation implementation to stm32f1xx-hal and update examples

### LINKED ISSUE #156: Change reference implementation to something more maintained

eldruin:
I think the `stm32f1xx-hal` is a good candidate for a reference implementation.
I have also updated the examples accordingly and modernized them a bit.

An additional possibility would be to remove the `futures` examples as well, since async handling will be different.

Closes #156


therealprof (approved):
LGTM, thanks.

bors r+

eldruin:
What do you think about removing the `futures`-based examples? Those are even stuck on `futures` `0.1`.

therealprof:
Sounds good to me.

PR #252, opened 2020-10-02
[v0.2.x] force nb to version compatible with 1.0

eldruin:
With `nb` < `0.1.3` other crates cannot use `nb` `1.0` as the types will not match.
A new release of 0.2.x including this would be good.


eldruin:
I opened this against the wrong branch. See: #253

PR #253, opened 2020-10-02
[0.2.x] Use nb version 0.1.3 for compatibility with nb 1.0

eldruin:
With `nb` < `0.1.3` other crates cannot use `nb` `1.0` as the types will not match.
A new release of 0.2.x including this would be good.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ default-features = false
 version = "1.0.2"
 
 [dependencies.nb]
-version = "0.1.1"
+version = "0.1.3"
 
 [dev-dependencies]
 stm32f30x = "0.8.0"


ryankurte:
i'm not opposed to this change but, shouldn't "0.1.1"  already match "0.1.3" as it is interpreted as a [caret requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)? seems like you should be able to `cargo-update` to this (tho sometimes ime you do have to delete `Cargo.lock` and start again).

eldruin:
That is correct, it would work fine after a cargo-update/`Cargo.lock` removal.
This is just to force it so that unmatching type errors are not possible. It recently happened to me and I figured it would be good to avoid it.

eldruin:
Somebody else hit this at https://github.com/nrf-rs/nrf-hal/issues/264

ryankurte (approved):
seems like a decrease in compatibility but, also fair enough.

eldruin:
bors r=ryankurte

ryankurte:
oooh, `Pin` API change in nightly?

```
---- src/lib.rs - (line 385) stdout ----
error[E0061]: this function takes 1 argument but 0 arguments were supplied
  --> src/lib.rs:442:31
   |
58 |         Pin::new(&mut blinky).resume();
   |                               ^^^^^^- supplied 0 arguments
   |

help: expected the unit value `()`; create it with empty parentheses
   |
58 |         Pin::new(&mut blinky).resume(());
   |              
```

eldruin:
Indeed. Should be fixed now.

Sh3Rm4n:
ping :)

eldruin:
bors try

ryankurte (approved):

ryankurte:
lgtm, maybe we should backport the gh-actions to this too one day...

bors r+

eldruin:
I did so in #272 I will rebase once that is merged

ryankurte (approved):
:+1: 

bors r+

PR #260, opened 2020-11-06
Transactional I2C address modes

eldruin:
The transactional I2C traits did not support multiple address modes.
Thanks to @ryankurte for noticing this at https://github.com/rust-embedded/linux-embedded-hal/pull/44


ryankurte (approved):
lgtm

bors r+

PR #261, opened 2020-11-11
Embedded-hal 1.0.0-alpha.4 release

eldruin:
Once we have a new alpha release we can move forward with merging https://github.com/rust-embedded/linux-embedded-hal/pull/44 
I also added the `html_root_url`, which is annoying but necessary according to the Rust API guidelines.


ryankurte (approved):
:+1:

eldruin:
I have now tagged this and published to crates.io.

PR #266, opened 2021-03-11
Fix code formatting + check it on bors

eldruin:
This was missed in #246.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -6,4 +6,5 @@ status = [
   "ci-linux (stable, thumbv6m-none-eabi)",
   "ci-linux (stable, thumbv7m-none-eabi)",
   "ci-linux (1.35.0, x86_64-unknown-linux-gnu)",
+  "fmt",
 ]
--- a/.github/workflows/rustfmt.yml
+++ b/.github/workflows/rustfmt.yml
@@ -7,7 +7,6 @@ name: Code formatting check
 
 jobs:
   fmt:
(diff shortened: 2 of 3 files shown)


eldruin:
bors try

therealprof (approved):
LGTM, thanks.

bors r+

eldruin:
bors r=therealprof

eldruin:
bors r=therealprof

eldruin:
bors try

eldruin:
bors try

eldruin:
bors try

eldruin:
Sorry for the back and forth. The bors requirements were not right.
Could you review this again @therealprof?

therealprof (approved):
Sure.

bors r+

therealprof:
Just for reference: I'm not super keen on the idea of making formatting checks mandatory. We had problems in the past when formatting rules changed (similar to clippy rules) and we should actually run CI on a schedule and not just on PRs (which should be fine since they should be up-to-date re formatting).

Here's to hope that Rust is not going to make a lot of tweaks to the formatting to not get back at us. 😅 Let's see how it goes.

eldruin:
True. clippy bit me all over the place with [a rule I believe is not good](https://github.com/rust-lang/rust-clippy/issues/6455) and I am still not even done updating all my drivers to disregard it. fmt has been quite stable so far, but yeah, good point.

PR #268, opened 2021-04-20
Remove try_ prefix.

Dirbaio:
This removes the `try_` prefix from all the trait method names, after discussion on issue #177 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Swap PWM channel arguments to references
+- All trait methods have been renamed to remove the `try_` prefix (i.e. `try_send` -> `send`) for consistency.
 
 ## [v1.0.0-alpha.4] - 2020-11-11
 
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -73,7 +73,7 @@ pub trait Channel<ADC> {
 /// {
 ///    type Error = ();
 ///
-///    fn try_read(&mut self, pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
+///    fn read(&mut self, pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
 ///        let chan = 1 << pin.channel();
 ///        self.power_up();
 ///        let result = self.do_conversion(chan);
@@ -90,5 +90,5 @@ pub trait OneShot<ADC, Word, Pin: Channel<ADC>> {
     ///
     /// This method takes a `Pin` reference, as it is expected that the ADC will be able to sample
     /// whatever channel underlies the pin.
-    fn try_read(&mut self, pin: &mut Pin) -> nb::Result<Word, Self::Error>;
+    fn read(&mut self, pin: &mut Pin) -> nb::Result<Word, Self::Error>;
 }
--- a/src/blocking/delay.rs
+++ b/src/blocking/delay.rs
@@ -16,7 +16,7 @@ pub trait DelayMs<UXX> {
     type Error;
 
     /// Pauses execution for `ms` milliseconds
-    fn try_delay_ms(&mut self, ms: UXX) -> Result<(), Self::Error>;
+    fn delay_ms(&mut self, ms: UXX) -> Result<(), Self::Error>;
 }
 
 /// Microsecond delay
@@ -28,5 +28,5 @@ pub trait DelayUs<UXX> {
     type Error;
 
     /// Pauses execution for `us` microseconds
-    fn try_delay_us(&mut self, us: UXX) -> Result<(), Self::Error>;
+    fn delay_us(&mut self, us: UXX) -> Result<(), Self::Error>;
 }
--- a/src/blocking/i2c.rs
+++ b/src/blocking/i2c.rs
@@ -30,7 +30,7 @@
 //! {
 //! #   type Error = ();
 //! #
-//!     fn try_write(&mut self, addr: u8, output: &[u8]) -> Result<(), Self::Error> {
+//!     fn write(&mut self, addr: u8, output: &[u8]) -> Result<(), Self::Error> {
 //!         // ...
 //! #       Ok(())
 //!     }
@@ -40,7 +40,7 @@
 //! {
 //! #   type Error = ();
 //! #
-//!     fn try_write(&mut self, addr: u16, output: &[u8]) -> Result<(), Self::Error> {
+//!     fn write(&mut self, addr: u16, output: &[u8]) -> Result<(), Self::Error> {
 //!         // ...
 //! #       Ok(())
 //!     }
@@ -66,7 +66,7 @@
 //!     pub fn read_temperature(&mut self) -> Result<u8, E> {
 //!         let mut temp = [0];
 //!         self.i2c
-//!             .try_write_read(ADDR, &[TEMP_REGISTER], &mut temp)
+//!             .write_read(ADDR, &[TEMP_REGISTER], &mut temp)
 //!             .and(Ok(temp[0]))
 //!     }
 //! }
@@ -89,7 +89,7 @@
 //!     pub fn read_temperature(&mut self) -> Result<u8, E> {
 //!         let mut temp = [0];
 //!         self.i2c
-//!             .try_write_read(ADDR, &[TEMP_REGISTER], &mut temp)
+//!             .write_read(ADDR, &[TEMP_REGISTER], &mut temp)
 //!             .and(Ok(temp[0]))
 //!     }
 //! }
@@ -135,7 +135,7 @@ pub trait Read<A: AddressMode = SevenBitAddress> {
     /// - `MAK` = master acknowledge
     /// - `NMAK` = master no acknowledge
     /// - `SP` = stop condition
-    fn try_read(&mut self, address: A, buffer: &mut [u8]) -> Result<(), Self::Error>;
+    fn read(&mut self, address: A, buffer: &mut [u8]) -> Result<(), Self::Error>;
 }
 
 /// Blocking write
@@ -159,7 +159,7 @@ pub trait Write<A: AddressMode = SevenBitAddress> {
     /// - `SAK` = slave acknowledge
     /// - `Bi` = ith byte of data
     /// - `SP` = stop condition
-    fn try_write(&mut self, address: A, bytes: &[u8]) -> Result<(), Self::Error>;
+    fn write(&mut self, address: A, bytes: &[u8]) -> Result<(), Self::Error>;
 }
 
 /// Blocking write (iterator version)
@@ -172,7 +172,7 @@ pub trait WriteIter<A: AddressMode = SevenBitAddress> {
     /// # I2C Events (contract)
     ///
     /// Same as `Write`
-    fn try_write_iter<B>(&mut self, address: A, bytes: B) -> Result<(), Self::Error>
+    fn write_iter<B>(&mut self, address: A, bytes: B) -> Result<(), Self::Error>
     where
         B: IntoIterator<Item = u8>;
 }
@@ -204,7 +204,7 @@ pub trait WriteRead<A: AddressMode = SevenBitAddress> {
     /// - `MAK` = master acknowledge
     /// - `NMAK` = master no acknowledge
     /// - `SP` = stop condition
-    fn try_write_read(
+    fn write_read(
         &mut self,
         address: A,
         bytes: &[u8],
@@ -223,7 +223,7 @@ pub trait WriteIterRead<A: AddressMode = SevenBitAddress> {
     /// # I2C Events (contract)
     ///
     /// Same as the `WriteRead` trait
-    fn try_write_iter_read<B>(
+    fn write_iter_read<B>(
         &mut self,
         address: A,
         bytes: B,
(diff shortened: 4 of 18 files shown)


therealprof:
This PR is related to a proposal in #177 and was discussed in todays meeting.

In the meeting there was a consensus that infallible versions of the methods/traits are not really required since the main proposition of `e-h` is to provide interfacing between drivers and HALs and not necessarily users of the HALs for which the HALs might as well provide additional infallible versions for "ease of use" purposes.

This PR is open for discussion and ultimately a team vote to give members not on the meeting a chance to way in.

ryankurte:
it seems to me this change is predominantly aesthetic, and i don't have strong opinions on this aspect. i would note that this will break compatibility for all `-alpha` users. pre-`1.x` is an appropriate time to do this but, it's also always annoying to have to go and update all the newly-incompatible things.

i would pose a weak opposition on that basis, and though it's not a requirement with the alpha series would be chuffed to see some mitigation for the incompatibility. i've been experimenting with [backwards-compatibility](https://github.com/ryankurte/embedded-hal-compat) via third party repo (as an alternative to the semver trick we used to maintain here), and i think an extension (or an alternative) to this that paves over this change could reduce this complaint.

Disasm:
If we have both fallible and infallible traits -- what naming convention should we use for infallible methods then? `do_`? `sudo_`?

therealprof:
@ryankurte OTOH it would mean that for a lot of API it would basically stay the same as in 0.2.x and wouldn't even have to change to for 1.0, no?

Sounds like a bonus benefit to me that I hadn't even considered before.

Dirbaio:
@Disasm the reasoning for removing `try_` prefixes is based on the idea that we're not going to add infallible traits. There seems to be consensus on not adding infallible traits.

Disasm:
@Dirbaio I don't see a consensus: I only see your intention. Infallible traits are still useful for GPIO, ADC, delays and RNGs because a lot of chips have peripherals that are infallible by their nature. Why not to implement both fallible and infallible traits in this case?
After all, why would anyone want to use fallible delay in their code?
To sum up: I do want to be able to use infallible traits for peripherals which are infallible by their nature. I don't see any point in this "fallible-only" limitation.

Dirbaio:
> I don't see a consensus: I only see your intention.

In the discussions in #177 and in today's meeting people have expressed neutral or against opinions, no one has voiced an in-favor opinion until now. 

Anyway, now that you have, lets discuss! :)

## Pros

1. Improved usability for the cases when it is actually infallible. You write `pin.set_high()` instead of `pin.set_high().unwrap()`.

## Cons

1. Having both fallible and infallible traits opens the door for drivers to use the infallible ones. This fragments the ecosystem. For example, a driver may use InfallibleOutputPin because the author's pins are infallible, ruling out using it with a I2C gpio expander.
2. Usability improvement is not realized in drivers due to con 1. This weakens pro 1.
3. Usability improvement can be achieved without infallible traits by having HALs define infallible methods. See [comment](https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-821083026) by @Rahix. The embedded-hal API doesn't have to be the main API for users using a HAL directly. For many HALs this is already the case, they define methods with extra functionality or usability not covered by the e-h traits. This weakens pro 1.
4. It requires ugly naming for the fallible traits: `TryFoo` traits, `try_foo` methods.
5. You can still write bounds requiring infallibility without infallible traits: Instead of `T: InfallibleOutputPin` , do `T: OutputPin<Error = !>`.
6. There's a (still unstable) [`Result::into_ok()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.into_ok) API to unwrap infallible Results without using `unwrap()`. This removes the "but my code is full of unwraps that can't panic, making the unwraps that really panic harder to find" concern.

6. Many traits' infallible versions will be useless in practice. I2C can always NAK, UART can always have framing errors, etc.

note 1: The Never type `!` is not stable, but can be replaced with core::convert::Infallible.

eldruin:
I do not have strong opinions about this if we do not want to add infallible traits. I also agree with @Dirbaio's pro/con summary, although I see that point 1 applies to a lot of traits (delay, RNG,...) so the HALs will add a lot of infallible types.

MathiasKoch:
I think the most important point of this discussion is the fragmentation introduced by having both fallible and infallible traits doing essentially the same thing.

From a driver authors point of view, there is a huge risk that only one of them would be supported (fallible or infallible), with the consequence of getting two camps of drivers that aren't compatible with each other, because one camp would always bound on the fallible, and the other camp would bound on the infallible versions.

In my eyes the main motivation for E-H should be to give some interfaces that would allow for as much code re-use as possible, while maintaining a drop-in replaceability.

I totally agree with @Dirbaio pro/cons list, but i think the success of fallible traits in general boils down to the outcome of the revamped error types, if driver authors are to ever be able to handle any of them generically, and not just propagate them to the user (who can't handle them either)

Disasm:
@Dirbaio let's discuss!

### Cons
1. This is the only concern I have about infallible traits. However, this is what we already have and we have a good solution for compatibility! One can convert between fallible and infallible with a wrapper that panics on error and opposite conversion is simple.
2. I'm not sure I get your point here.
3. The problem is that HALs almost certainly will define methods with different names/signatures which beats the purpose. Traits solve this problem.
4. `This could be a helper crate` is called `embedded-hal`, yeah.
5. Let's discuss the alternatives! Anyway, `_try` contains less characters than `.unwrap()`.
6. Fix: Instead of `T: InfallibleOutputPin` , do `T: OutputPin<Error = Infallible>`. Now it's longer, right? As you mentioned, Never type is not stabilized yet.
7. `into_ok()` is even uglier than `unwrap()`, also one character longer.
8. Yes, but we don't have to add infallible versions for every single fallible trait! I mentioned those entities where infallible traits are useful in practice.

-----
Here is my version:

Proposal: don't remove the `try_` prefix everywhere, at least don't remove it for traits where infallible counterparts make sense. Add infallible traits after v1.0 (as decided in https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary).
## Pros
1. Appropriate types are used to represent results. If function result is `usize`, the function return type is `usize` instead of `Result<usize, Infallible>`.
2. Improved usability for the cases when it is actually infallible. You write `pin.set_high()` instead of `pin.set_high().unwrap()`.

## Cons
1. Having both fallible and infallible traits opens the door for drivers to use the infallible ones. This fragments the ecosystem. For example, a driver may use `InfallibleOutputPin` because the author's pins are infallible, requiring an additional shim when using it with an I2C GPIO expander.

adamgreig:
> Yes, but we don't have to add infallible versions for every single fallible trait! I mentioned those entities where infallible traits are useful in practice.

The main new argument I've seen against "useful" infallible traits was that end users don't need the infallible versions to be in traits, because they're already using a specific HAL and can use its normal methods, so only people writing generic drivers need to use the e-h traits, and we'd like them to always use the fallible versions for maximum compatibility.

Basically: the extra effort of dealing with fallible methods only falls to generic driver authors, and end-user application code would normally use the infallible methods provided by their HAL directly.

However, maybe you're thinking of end users who are writing firmware like hs-probe where you actually want to easily port the same application firmware between different HALs, so you'd like to only/mostly use embedded-hal interfaces? In which case having the fallible methods would be more convenient for you, but to me it doesn't seem like it outweighs the advantage of only having fallible methods in the traits.

caemor:
The pins are able to fall on "embedded" linux systems for example so at least for my display driver (epd_waveshare) it is easier to just use the fallible version to easily switch between the different boards and supporting both would just double the code size and effort which is not worth it in my case.

Disasm:
@adamgreig so you don't think that infallible delay is useful? I think it's the only option right now and it's used almost everywhere.

adamgreig:
For GPIO, ADCs, and RNGs fallible all makes sense to me - GPIOs via expanders or on Linux etc, RNGs usually have error detection bits, ADCs are quite likely to be accessed over SPI or I2C which can be fallible - but I'm not sure about delays... I guess the main idea here is to not have both fallible and infallible versions of the same thing, not to only have fallible methods, but it seems we can think of an example that might need to be fallible for basically every type of thing.

Is there a use case for fallible delays?

Disasm:
@adamgreig Fallible traits make sense for literally everything. In Linux `usleep(3)` is fallible as it can be interrupted by a signal.

Dirbaio:
replying inline to @Disasm 

> For example, a driver may use InfallibleOutputPin because the author's pins are infallible, requiring an additional shim when using it with an I2C GPIO expander.

one of `embedded-hal`'s goals (the main one IMO) is to enable writing portable drivers. It would be a spectacular failure of that goal if such shims become widespread in the ecosystem. Also, how would that shim handle errors? It can't propagate them up because it has to implement the infallible trait, all I can think of is either unwrap, or store the error in `self` or a global static, all really bad options.

> 2. I'm not sure I get your point here.

The argument is that since driver crates must use the fallible traits for maximum compatibility, the usability improvement doesn't apply to them. It applies only to end-user code where the user knows their HAL is infallible.

> 3. The problem is that HALs almost certainly will define methods with different names/signatures which beats the purpose. Traits solve this problem.

It seems to me you want infallible traits to be able to write *somewhat* portable code that works with multiple HALs but only as long as they're infallible?

As I see it, most people are either writing fully-portable code (drivers) which should use the infallible traits, or writing code against a particular HAL for a particular project. The middle-ground "somewhat portable only for infallible HALs" use case seems quite niche to me.

Aditionally, HALs defining methods with different names/signatures/contracts is not necessarily a bad thing:

- Non-trait methods can provide extra functionality not covered by the traits. For example, [nrf52's SPI](https://docs.rs/nrf52840-hal/0.12.1/nrf52840_hal/spim/struct.Spim.html) can do different buffers for Tx/Rx, different buffer sizes, hardware CS control.
- There's often hardware limitations that require workarounds when implementing the traits. For example, nrf52's DMA can only transfer from RAM (not Flash) and transfer max 255/65535 bytes. There's no such limitation in the trait's contract, so the trait impl must [copy to the stack in chunks](https://github.com/nrf-rs/nrf-hal/blob/04e3fcb03c171c1a68f29e989a0021780992bbc0/nrf-hal-common/src/uarte.rs#L355) to support buffers in flash and buffers bigger than 255/65535 bytes. The non-trait methods document this restriction, and can therefore skip this inefficiency.

> 4. This could be a helper crate is called embedded-hal, yeah.

I think embedded-hal's goal should be to collect fully-portable traits, enabling writing fully-portable drivers. Adding infallible traits in embedded-hal undermines that goal.

> 5. Let's discuss the alternatives! Anyway, _try contains less characters than .unwrap().
> 7. into_ok() is even uglier than unwrap(), also one character longer.

Not sure how these are arguments in favor of infallible traits.

>6. Fix: Instead of T: InfallibleOutputPin , do T: OutputPin<Error = Infallible>. Now it's longer, right? As you mentioned, Never type is not stabilized yet.

Not sure I follow. My argument was that infallible traits are not required for writing bounds requiring something is infallible. You write `T: OutputPin<Error = core::convert::Infallible>`. It being short or long is not the point, the point is it's *possible* to do it without infallible traits. This makes "I want to bound something to be infallible" no longer be an argument in favor of infallible traits.

8. Yes, but we don't have to add infallible versions for every single fallible trait! I mentioned those entities where infallible traits are useful in practice.

Fair! It would be inconsistent to add infallible versions of only some traits though.

> Add infallible traits after v1.0 (as decided in https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary).

No decision to add infallible traits has been made. The decision made there is to *not do it for 1.0, maybe do it later if necessary*. 

I'm raising the discussion now because IMO we don't want infallible traits, and deciding that now (vs post 1.0) allows releasing 1.0 with better naming (no `try_`). It'd be a shame to release 1.0 with `try_` "just in case", to later decide that we're not adding infallible traits.

jounathaen:
A slightly different _con_ came to my mind:

I hate the faillable traits for GPIO and stuff and for "user" code they are usually unnecessary and lead to a lot of unnecessary `unwrap()`. *But* the embedded-hal is not about "user" code but rather about providing an abstraction for drivers. Therefore I can see that only providing fallible traits makes much sense.

To make both sides happy, an implementation of the embedded-hal could then implement the faillible `try-` traits to implement the embedded-hal but could also implement the infailible versions where it makes sense. The infailible ones are then just hal specific and not standardized. However, if the embedded-hal blocks these names for the infallible functions, then a hal implementation would have to work around this by some other name schema which wouldn't be that natural.

**TLDR: embedded-hal could block sensible names for custom infallible implementations**

Dirbaio:
> if the embedded-hal blocks these names for the infallible functions, then a hal implementation would have to work around this by some other name schema which wouldn't be that natural.

You can implement an inherent method `foo`, and a trait requiring a method `foo` at the same time, no problem. If the user does `x.foo()`, it'll call the inherent method (if the trait is not imported, which is the most likely).

If you're working with a particular HAL that implements infallible set_high and set_low, they do *exactly* the same thing as the trait methods, so there's no reason whatsoever you'll want to call the trait methods.

If you're writing portable driver code, you'll be working with a generic `T: OutputPin`, in which case you only can call the trait methods, so there's no ambiguity there either.

jounathaen:
Ah, thanks for explaining it. Then of course this argument is invalid. :+1:

therealprof:
As discussed on today meeting all arguments have been exchanged and this is very much in line with earlier agreements to make all traits fallible. If later deemed as needed we can always put infallible traits into a separate module to avoid naming clashes.

Thus putting this up to a @rust-embedded/hal team vote:

- [ ] @therealprof
- [ ] @ryankurte 
- [ ] @eldruin

eldruin (approved):
Alright, we got consensus :tada: 
Thanks everybody!

bors r+

ryankurte:
y'all might find something like `find . -type f -name "*.rs" | xargs sed -i 's/try_//g'` helpful when updating, everything.

PR #269, opened 2021-04-25
Add digital::IoPin to resolve issue 29

### LINKED ISSUE #29: [RFC] `digital::IoPin`, pins that can switch between input and output modes at runtime

MorganR:
Closes #29.

This implementation is inspired by [chrismoos's comment](https://github.com/rust-embedded/embedded-hal/issues/29#issuecomment-717590055) in #29.

I've demonstrated using this trait for a [no-std DHT11 driver](https://github.com/MorganR/rust-simple-sensors/blob/main/src/dht11.rs), and verified this on a Raspberry Pi 3B+ using an implementation in linux-embedded-hal (PR to follow). 

I'll also be sending a PR to implement this in at least one additional HAL implementation library (recommendations welcome).

One question I have is how copyright/authorship is tracked in this repo? I believe the copyright owner for my code would technically be Google LLC. When patching repos, I'm meant to add that to the appropriate copyright/authors list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Added `IoPin` trait for pins that can change between being inputs or outputs
+  dynamically.
 
 ### Changed
 - Swap PWM channel arguments to references
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -174,3 +174,49 @@ pub trait InputPin {
     /// Is the input pin low?
     fn try_is_low(&self) -> Result<bool, Self::Error>;
 }
+
+/// Single pin that can switch from input to output mode, and vice-versa.
+///
+/// Example use (assumes the `Error` type is the same for the `IoPin`,
+/// `InputPin`, and `OutputPin`):
+///
+/// ```
+/// use core::time::Duration;
+/// use embedded_hal::digital::{IoPin, InputPin, OutputPin};
+///
+/// pub fn ping_and_read<TInputPin, TOutputPin, TError>(
+///     mut pin: TOutputPin, delay_fn: &dyn Fn(Duration) -> ()) -> Result<bool, TError>
+/// where
+///     TInputPin : InputPin<Error = TError> + IoPin<TInputPin, TOutputPin, Error = TError>,
+///     TOutputPin : OutputPin<Error = TError> + IoPin<TInputPin, TOutputPin, Error = TError>,
+/// {
+///     // Ping
+///     pin.try_set_low()?;
+///     delay_fn(Duration::from_millis(10));
+///     pin.try_set_high()?;
+///
+///     // Read
+///     let pin = pin.try_into_input_pin()?;
+///     delay_fn(Duration::from_millis(10));
+///     pin.try_is_high()
+/// }
+/// ```
+pub trait IoPin<TInput, TOutput>
+where
+    TInput: InputPin + IoPin<TInput, TOutput>,
+    TOutput: OutputPin + IoPin<TInput, TOutput>,
+{
+    /// Error type.
+    type Error;
+
+    /// Tries to convert this pin to input mode.
+    ///
+    /// If the pin is already in input mode, this method should succeed.
+    fn try_into_input_pin(self) -> Result<TInput, Self::Error>;
+
+    /// Tries to convert this pin to output mode with the given initial state.
+    ///
+    /// If the pin is already in the requested state, this method should
+    /// succeed.
+    fn try_into_output_pin(self, state: PinState) -> Result<TOutput, Self::Error>;
+}


MorganR:
PR on linux-embedded-hal: https://github.com/rust-embedded/linux-embedded-hal/pull/60

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- Added IoPin for swappable input/output pins.
```
eldruin, line 11:
```suggestion
- Added `IoPin` trait for pins that can change between being inputs or outputs dynamically.
```
Or something similar.

src/digital.rs
```diff
@@ -174,3 +174,22 @@ pub trait InputPin {
...
+    /// 
+    /// If the pin is already in input mode, this method should succeed.
+    fn try_switch_to_input_pin(self) -> Result<TInput, Self::Error>;
+
+    /// Tries to convert this pin to output mode with the given initial state.
+    /// 
+    /// If the pin is already in the requested state, this method should
+    /// succeed.
+    fn try_switch_to_output_pin(self, state: PinState) -> Result<TOutput, Self::Error>;
+}
```
eldruin, line 195:
```suggestion
}

```
newline

src/digital.rs
```diff
@@ -174,3 +174,22 @@ pub trait InputPin {
     /// Is the input pin low?
     fn try_is_low(&self) -> Result<bool, Self::Error>;
 }
+
+/// Single input pin that can switch from input to output mode, and vice-versa.
```
eldruin, line 178:
```suggestion
/// Single pin that can switch from input to output mode, and vice-versa.
```

src/digital.rs
```diff
@@ -174,3 +174,22 @@ pub trait InputPin {
...
+pub trait IoPin<TInput, TOutput> 
+where TInput : InputPin + IoPin<TInput, TOutput>,
+      TOutput : OutputPin + IoPin<TInput, TOutput> {
+    /// Error type.
+    type Error;
+
+    /// Tries to convert this pin to input mode.
+    /// 
+    /// If the pin is already in input mode, this method should succeed.
+    fn try_switch_to_input_pin(self) -> Result<TInput, Self::Error>;
```
eldruin, line 188: Since this consumes `self`, the method should be called something like `into_`...

src/digital.rs
```diff
@@ -174,3 +174,22 @@ pub trait InputPin {
...
+
+    /// Tries to convert this pin to output mode with the given initial state.
+    /// 
+    /// If the pin is already in the requested state, this method should
+    /// succeed.
+    fn try_switch_to_output_pin(self, state: PinState) -> Result<TOutput, Self::Error>;
```
eldruin, line 194: Would _not_ setting an initial state also make sense?

eldruin (commented):
Thanks for this proposal!
Could you add some examples on how to use this to the docs?
I do not know about copyright / authorship attribution matters. The current crate author list is missing many people.

src/digital.rs
MorganR: If the user didn't specify the state, then the result would be a pin in output mode that's either high or low, but the user doesn't know which it is. I think it's better to make the user specify which state they want rather than to leave it ambiguous.

That being, said I'm no expert in the embedded space. Do you see a valid use case?

src/digital.rs
eldruin: Yes, thanks!

src/digital.rs
eldruin: I was just wondering. From a software perspective it makes sense to always change to a defined state. However, from a hardware perspective, it may not _always_ make sense to set a pin, since it has electrical implications.
Does anybody else here have an use case for changing the pin type to output but leaving it unset?

src/digital.rs
```diff
@@ -174,3 +174,47 @@ pub trait InputPin {
     /// Is the input pin low?
     fn try_is_low(&self) -> Result<bool, Self::Error>;
 }
+
+/// Single pin that can switch from input to output mode, and vice-versa.
+/// 
+/// Example use (assumes the Error type is the same for IoPin, InputPin, and
```
eldruin, line 180:
```suggestion
/// Example use (assumes the `Error` type is the same for `IoPin`, `InputPin`, and
```

src/digital.rs
```diff
@@ -174,3 +174,47 @@ pub trait InputPin {
...
+/// OutputPin):
```
eldruin, line 181:
```suggestion
/// `OutputPin`):
```

eldruin (commented):
Looks good to me, although I do not have the whole discussion about this in #29 in mind at the moment.

therealprof:
@MorganR This is looking great.

I'm wondering if we can make the type stuff a little bit easier on the eyes and the mind but other than that I like it.

src/digital.rs
MorganR: IIUC, an example use-case might be an output pin in open drain mode (I could be wrong; like I said, I'm not an embedded expert). However, the current digital traits don't seem to provide any support around these states either. Maybe it makes sense to look at that separately, eg. as an addition to PinState? Definitely interested in what other folks think here, as @eldruin said.

MorganR:
Thanks @therealprof! I agree the type system becomes a bit verbose when using the abstract trait. However, it's actually fairly clean when implementing the trait, and in user code that gives a concrete implementation to something using the abstraction.

To simplify things a little more, I've removed the `TIoPin` parameter from the example function. This is superfluous because the function can instead accept a pin that's already in either TInputPin or TOutputPin format.

therealprof:
If you could squash the commits and maybe add a "Closes #29" to your initial post, that would be great.

MorganR:
> If you could squash the commits and maybe add a "Closes #29" to your initial post, that would be great.

Done

therealprof (dismissed):
Thanks a lot for your efforts!

bors r+

MorganR:
Sorry, fixed that remaining fmt error. Should be good now.

therealprof:
@MorganR I guess rustfmt doesn't like your formatting. 😅 It's a pity the new GH protection mechanisms seem to silently prevent actions from running...

therealprof (approved):
Let's try this again.

bors r+

burrbull:
Which combinations of output and input modes this should be implemented?
PushPull, OpenDrain
PullUp, PullDown, Floating
Alternate<MODE>

PR #270, opened 2021-04-27
Remove RNG traits in favor of rand_core.

### LINKED ISSUE #128: [Discussion] rand::RngCore for blocking RNG trait

jounathaen:
As proposed in #128 here is a PR that completely removes the RNG traits from the embedded-hal.
This closes  #128.

# Reasoning:
The [rand_core](https://crates.io/crates/rand_core) crate already implements relevant traits for random number generation. A HAL which provides access to an RNG peripheral should implement these traits instead ([Example: stm32f4xx-hal](https://docs.rs/stm32f4xx-hal/0.9.0/stm32f4xx_hal/rng/struct.Rng.html)) and a driver should rely on these.

# Impact:

This is a breaking change. Should be done before a 1.0.0 release.

# Alternatives

- Don't drop this trait. See #128 for suggestions on this
- Instead of instantly dropping it, put in a deprecation warning and drop in a later point release. See #271


jounathaen:
In the discussion, it was mentioned that some pointers and documentation on this change are required. I'm not sure on where to do this. The release notes seem to be the best place as it is hard to document stuff that is not existing anymore.

jounathaen:
It's probably wiser to first mark the traits as deprecated as in #271 before removing them.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Swap PWM channel arguments to references
+- Dropped RNG traits in favor of [rand_core](https://crates.io/crates/rand_core)
```
eldruin, line 14:
```suggestion
### Removed
- Removed RNG traits in favor of [rand_core](https://crates.io/crates/rand_core)
```

eldruin (commented):
Looks good to me. Thanks! I would go ahead with the removal without deprecation step.
Any objections @rust-embedded/hal?

ryankurte:
i'm okay with it, maybe it'd be nice to have a note in the docs somewhere to go looking at `rand_core`?

jounathaen:
Regarding the documentation: One could re-add the modules but without the traits and only with a note to `rand_core`.
I've [already prepared the commit](https://github.com/jounathaen/embedded-hal/commit/1f17c03365e4a42fab61935b5323bf6d3c188c07) (but I've not added it to this PR yet to avoid commit spam).

eldruin:
I would add the documentation note but not the `rng` module, since removing it would technically be a breaking change.

jounathaen:
> I would add the documentation note but not the rng module, since removing it would technically be a breaking change.

Sorry, but I didn't fully understand where you think the note should go.

eldruin:
I think I would put it somewhere in `lib.rs`.

PR #271, opened 2021-04-27
Mark RNG traits as deprecated

### LINKED ISSUE #128: [Discussion] rand::RngCore for blocking RNG trait

jounathaen:
This is the softer approach than the nuclear option presented in #270 or at least the first step. Would also close #128 .

I'm not super sure about the removal of the traits from the prelude. This might break stuff. Alternatively an `#[allow(deprecated)]` for these imports would work.


therealprof:
Since we're in agreement to go with #270 I'm closing this.

PR #273, opened 2021-04-28
Some minor improvements

eldruin:
- Update codeowners
- Remove unnecessary imports
- Replace `into_iter()` with `iter()` when used on reference
- Deny warnings on cargo check and cargo test on CI (but allow them on clippy since that is a fast-moving target)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @rust-embedded/hal @ilya-epifanov @thejpster
+* @rust-embedded/hal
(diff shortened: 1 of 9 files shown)


eldruin:
This needs to be rebased again after #274 is merged

eldruin:
Alright, rebased. This should be good to go now.

burrbull:
https://doc.rust-lang.org/std/primitive.str.html#method.bytes

eldruin:
Thanks @burrbull! That simplified the code a bit further.

ryankurte (approved):
:+1:

PR #276, opened 2021-04-29
Add documentation on how to support an alpha version side-by-side in a HAL impl

eldruin:
Credit goes to @hannobraun. Thank you!
See [here](https://github.com/flott-motion/stepper/blob/main/documentation/platform-support.md#embedded-hal-alpha)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ as well as the latest `1.0.0-alpha` version side by side. This has several big a
 - Allows for a faster upgrade to `1.0` once it comes out.
 - Provides more oportunities to test the new `embedded-hal` version.
 
-This approach has been exemplary implemented in [LPC8xx HAL]. Here are the steps:
+This approach has been implemented in [LPC8xx HAL](https://github.com/lpc-rs/lpc8xx-hal). Here are the steps:
 
 1. Add a dependency to the latest `embedded-hal` version to `Cargo.toml`.
    Use the `package` attribute to refer to it by another name, to prevent name collision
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +57,29 @@ list.
...
 
+### Adding support for an `embedded-hal` `-alpha` version in a HAL implementation
+
+It is possible for HAL implementations to support both the latest `0.2.x` version of `embedded-hal`
+as well as the latest `1.0.0-alpha` version side by side. This has several big advantadges:
+- Allows for a more gradual upgrade process within the community.
+- Allows for a faster upgrade to `1.0` once it comes out.
+- Provides more oportunities to test the new `embedded-hal` version.
+
+This approach has been exemplary implemented in [LPC8xx HAL]. Here are the steps:
```
hannobraun, line 68:
```suggestion
This approach has been implemented in [LPC8xx HAL](https://github.com/lpc-rs/lpc8xx-hal). Here are the steps:
```
Link was missing.

Also, "exemplary" reads a bit weird to me, as it has [other meanings](https://en.wiktionary.org/wiki/exemplary) besides "providing an example". Also, wouldn't it need to be an adverb here (exemplarily)? I suggest to just remove the word, but it doesn't matter much. I don't think I would have even thought much about it, if there hadn't been the broken link in the same sentence :grin:

eldruin: Thanks! and good point.

therealprof (approved):
LGTM, thanks!

bors r+

Issue #186, opened 2020-02-04
Discussion:  Add way to read the current timer value in CountDown trait


conorpp:
I'd like to use a timer to measure how long something takes.  I can't use the timer trait currently since there's no way to read the current timer value.  What do we think about adding a method to `CountDown` that returns the current timer value?

For example, see `lap` method below.

```rust
pub trait CountDown {
    /// The unit of time used by this timer
    type Time;

    /// Starts a new count down
    fn start<T>(&mut self, count: T)
    where
        T: Into<Self::Time>;

    /// Returns the current value of the timer as how much time has elapsed since starting.
    fn lap<T>(&self,) -> T
    where
        T: Into<Self::Time>;

    /// Non-blockingly "waits" until the count down finishes
    ///
    /// # Contract
    ///
    /// - If `Self: Periodic`, the timer will start a new count down right after the last one
    /// finishes.
    /// - Otherwise the behavior of calling `wait` after the last call returned `Ok` is UNSPECIFIED.
    /// Implementers are suggested to panic on this scenario to signal a programmer error.
    fn wait(&mut self) -> nb::Result<(), Infallible>;
}
```

dbrgn:
I also need a way to get the current counter value of a timer. However, in my case I *only* need that, no need for starting / resetting / waiting.

In the case of an STM32 timer, an implementation is actually really simple, you just start the timer and then read the CNT (counter) register of the timer.

However, I'm not sure what the best API design is around timers... Seems that they can be used in a *lot* of different ways, so it's hard to build an abstraction that works everywhere. Maybe the focus should be on use cases, not on the capabilities.

One use case is "waiting until a certain time has elapsed". That is best solved by an interrupt based timer like `CountDown`.

Another use case is "measuring relative time that has elapsed" between two events. That can be done using a periodic up- or downcounting timer where the counter register can be read. Depending on timer resolution, an implementation might need to deal with overflows. I'm not sure whether this is something that should be in `embedded_hal` or whether this should be up to the implementation.

An example of the second use case is the `Monotonic` trait of RTIC:

https://github.com/rtic-rs/cortex-m-rtic/blob/8a4f9c6b8ae91bebeea0791680f89375a78bffc6/src/lib.rs#L129-L162

An example implementation can be found [here](https://github.com/rtic-rs/rtic-examples/blob/master/rtfm_v5/monotonic_nrf52/src/monotonic_nrf52.rs).

PTaylor-us:
[...]

PR #277, opened 2021-05-02
Add trait for count down times that allow reading out elapsed time

nickray:
Concrete suggestion to resolve https://github.com/rust-embedded/embedded-hal/issues/186.

I'd prefer `Elapsed::elapsed` to be infallible, but that is probably against the tide of the times :)

We use pretty much this implementation in https://github.com/lpc55/lpc55-hal/blob/5c70be5f73da2a6e1330aefc82a53f6cd3ddc7d6/src/drivers/timer.rs#L43-L48

Of course this is infinitely bike-sheddable (remaining? now? etc...), it would be a pity not to have a standardized way of reading out a timer if embedded HAL is headed for a 1.0.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -79,6 +79,12 @@ pub trait CountDown {
     fn wait(&mut self) -> nb::Result<(), Self::Error>;
 }
 
+/// A count down timer that allows reading out elapsed time since start.
+pub trait Elapsed: CountDown {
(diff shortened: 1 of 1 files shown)


eldruin:
Looks good to me. I can imagine `elapsed()` implementations returning an error if the timer was canceled in the mean time, for example.
Maybe some more documentation about this could be added.

PR #278, opened 2021-05-07
[Reorganization Proposal 1] Move traits into blocking/nonblocking modules

eldruin:
As discussed in our weekly meeting.
I chose `nonblocking` as module name because `nb` could be misunderstood as a re-exported `nb` crate.
An alternative is #279


burrbull (dismissed):

ryankurte:
is it worth considering this in the context of https://github.com/rust-embedded/embedded-hal/issues/172 (and the possible provision of future-based non-blocking types?)

eldruin:
As we discussed in the meeting, `async` traits would live in a parallel `asynchronous`/`futures`/`...` module besides `blocking` and `nonblocking`/`nb`, since it seems it will not be possible to use `nb` (the crate) as intended in its version `0.1` to handle all execution models.

eldruin:
Any objections or can we merge this @rust-embedded/hal ?
This PR is a bit of a gatekeeper for any other changes because of the file shuffling.

ryankurte:
parallel `futures` module seems reasonable and i don't have strong opinions for / against, lgtm

eldruin:
Alright, can I get an actual approval here @rust-embedded/hal?

therealprof (approved):
Seems we're all in agreement here, so let's do this.

bors r+

PR #279, opened 2021-05-07
[Reorganization Proposal 2] Nonblocking traits to module, blocking traits to root

eldruin:
As discussed in our weekly meeting.
I chose `nonblocking` as module name because `nb` could be misunderstood as a re-exported `nb` crate.
An alternative is #278


eldruin:
Closing in favor of #278

PR #280, opened 2021-05-28
[RFC] blocking/serial: make method naming consistent

Dirbaio:
`blocking::serial::Write`'s methods are named `bwrite_all` and `bflush`.

- the `b` prefix indicates blocking (I guess?). However no other blocking trait method has this prefix.
- the `_all` suffix indicates it writes a slice instead of a single word (I guess?). However, no other trait method that writes a slice has this suffix.

To make them consistent, this PR renames them to `write` and `flush`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Moved all traits into two modules depending on the execution model: `blocking` and `nb` (non-blocking).
 - Re-export `nb::{block!, Error, Result}` to avoid version mismatches. These should be used instead of
   importing the `nb` crate directly in dependendent crates.
+- `blocking::Serial`: renamed `bwrite_all` to `write`, `bflush` to `flush.
 
 ## [v1.0.0-alpha.4] - 2020-11-11
 
--- a/src/blocking/serial.rs
+++ b/src/blocking/serial.rs
@@ -10,13 +10,13 @@ pub trait Write<Word> {
     /// An implementation can choose to buffer the write, returning `Ok(())`
     /// after the complete slice has been written to a buffer, but before all
     /// words have been sent via the serial interface. To make sure that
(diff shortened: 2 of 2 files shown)


eldruin:
For completeness, as I highlighted in #281, different method names ensure unequivocal selection between different traits when they are all pulled in by the prelude. 
- `dev.read()` -> `nb::serial::Read::read()`
- `dev.write()` -> `nb::serial::Write::write()`
- `dev.flush()` -> `nb::serial::Write::flush()`
- `dev.bwrite_all()` -> `blocking::serial::Write::bwrite_all()`
- `dev.bflush()` -> `blocking::serial::Write::bflush()`

The method names can certainly be improved, though. Removing the prelude is also an option.
With the same method names users need to either stop using the prelude or turbofish calls.

ryankurte:
the blocking/non-blocking delineation makes more sense than the send/write difference to me, i suppose we could also split the prelude for blocking/non-blocking/futures?

Dirbaio:
> the blocking/non-blocking delineation makes more sense than the send/write difference to me,

Agreed, but then all blocking methods should have the `b` prefix, not just UART. Or all the nb methods should have the `nb_` prefix.

> i suppose we could also split the prelude for blocking/non-blocking/futures?

That won't work nicely because people end up using both blocking and nb traits in the same mod. Especially since some traits are only available in blocking flavor and others only in nonblocking flavor.

Nobody uses the *same* trait in both blocking and nonblocking flavors at the same time though. IMO naming blocking and nonblocking the same is not that bad.

eldruin (dismissed):
Looks good to me, thanks!

Dirbaio:
ping @ryankurte @therealprof is there anything missing to get this in? also for #282

therealprof (dismissed):
LGTM, bors r+

therealprof:
Should have a CHANGELOG.md entry.

therealprof (approved):
LGTM now, thanks.

bors r+

PR #282, opened 2021-06-15
Remove prelude.

Dirbaio:
**DEPENDS on #280**

Following discussions in #280 #281, we might want to use the same method names for the `blocking` and `nb` (and in the future, `futures`) trait flavors.

In that case, importing the prelude guarantees naming conflicts, so it's better to remove it.

An alternative would be splitting the prelude in `blocking` and `nb`. However, users often use blocking trait A and nonblocking trait B at the same time, so they'd end up importing both preludes and having name conflicts anyway.


eldruin (approved):
Looks good to me, thanks!

therealprof:
This is missing a CHANGELOG.md entry

Dirbaio:
Added changelog.

to avoid conflicts **this PR now depends on #280**

therealprof (approved):
LGTM, thanks!

bors r+

PR #284, opened 2021-06-16
Add initial version of migration guide

### LINKED ISSUE #283: Migration page
As we have a lot of changes (especially renames) in coming 1.0, it is good to have migration manual, how to port hal and driver crates on it.

eldruin:
This PR will be updated again once we settle on the latest changes. See #177.

[Rendered](https://github.com/eldruin/embedded-hal/blob/migration-guide-1.0/MIGRATING-0.2-1.0.md)

TODO: 
- [x] inside-out modules
- [ ] embedded-hal-compat
- [ ] Dedicated use-case help
- [ ] Interoperability via implementation of both 0.2.x and 1.0.0

Fixes #283

### DIFF
--- a/MIGRATING-0.2-1.0.md
+++ b/MIGRATING-0.2-1.0.md
@@ -9,6 +9,7 @@
   - [Method renaming](#method-renaming)
   - [`nb` dependency](#nb-dependency)
   - [Prelude](#prelude)
+  - [`rng` module](#rng-module)
   - [Features](#features)
   - [Use-case-specific help](#use-case-specific-help)
     - [For driver authors](#for-driver-authors)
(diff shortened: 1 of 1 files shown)


MIGRATING-0.2-1.0.md
```diff
@@ -0,0 +1,96 @@
...
+```diff
+- use nb::{Result, block};
++ use embedded_hal::nb::{Result, block};
+```
+
+## Prelude
+
+The prelude has been removed because it could make method calls ambiguous, since the method names are now
+the same across execution models.
+To overcome this, simply import the traits you wish to use individually.
```
ryankurte, line 62: v specific but, prefer not to use `simply` as it lessens the impact of actually having to do this

```suggestion
To overcome this, please import the traits you wish to use individually.
```

eldruin: Thanks!

eldruin:
@ryankurte Maybe you can think of a section introducing `embedded-hal-compat`

Dirbaio:
done in #554

Issue #149, opened 2019-08-03
Plan for core::future


tustvold:
Currently the traits within embedded-hal make use of the nb crate's flavour of async IO, however, `core::future` and `core::task` have recently been standardised and I believe are slated for inclusion in Rust 1.38. As such I was wondering whether migration towards using them would be something the maintainers of this project would be open to? 

In particular use of  `core::task::Waker` opens up the possibility for HAL crates to use interrupts to signal when they are ready to make progress, avoiding the busy waiting that appears to be largely unavoidable with the current interface.

I'm aware that I'm not the first person to wonder this, see [here](https://josh.robsonchase.com/embedded-executor/), but I couldn't find an issue discussing it. The closest I could find was on the nb crate itself [here](https://github.com/japaric/nb/issues/14).

If there is interest in this course of action I'm happy to look into writing a PoC, but I'm acutely aware there is, understandably, some caution around making breaking changes to the traits - see [here](https://github.com/rust-embedded/embedded-hal/issues/100).

tustvold:
So I'll caveat all the below with the fact I'm relatively new to rust and so what I say may be incorrect, but following further experimentation I do not believe it is yet possible for embedded_hal to make use of `core::future` and related types.

This is not to say that these types can't be used in embedded, far from it - it would be clunky but it is perfectly possible to use them in an embedded context, however, the specific constraints imposed by the design of embedded_hal, namely its use of traits and non static lifetimes, make this at least beyond my capabilities.

The easiest way to justify this, is to explain what I did, the problems I faced and how I bodged around them.

My first step was to build an "executor" that can run futures in an embedded environment. I don't believe it is possible to implement something like the executors in the standard library as the spawn interface seems to require dynamic memory allocation. However, it is relatively straightforward to write methods that can drive a `core::future` or an array of `&mut core::future` to completion, using a combination of the stateless wakers I describe in the linked issue, and `core::future::LocalFutureObj`. Implementations of the familiar `block!`, etc... macros can then be written using these methods.

Next I extended the traits within embedded_hal to use associated types to represent the future return types. I then tried to implement these traits within stm32f4xx_hal, however, immediately ran into an issue when the future needs to borrow a reference to the underlying peripheral for the lifetime of the future. Currently generic associated types are not close to stable, and so I bodged around this by making each of the embedded_hal traits generic over a lifetime parameter, and using this to parameterise the future associated type. The default blocking implementations required some HRTB magic to work, but I managed to make it work.

[...]

PR #285, opened 2021-06-21
Add asynchronous versions of most embedded-hal traits using GATs

lachlansneff:
GATs (generic associated types), among other things let you return futures from trait methods.

The lifetimes can be a bit tricky but it results in a pretty clean trait.

For example:

```rust
/// Async transfer in place.
pub trait TransferInPlace<Word: 'static> {
    /// Error type
    type Error;

    /// Associated future for the `transfer` method.
    type TransferInPlaceFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
    where
        Self: 'a;

    /// Writes `words` to the slave from the `readwrite` buffer and reads words into the same buffer.
    /// This method uses a single `readwrite` buffer.
    fn transfer_inplace<'a>(&'a mut self, readwrite: &'a mut [Word]) -> Self::TransferInPlaceFuture<'a>;
}
```

It's important to note that GATs are only available on nightly at the moment, so this uses feature attributes. I'd like to discuss how this should be exposed (likely through a feature until GATs are stabilized).

Also, there are some traits that don't use futures that are duplicated in the `nb`, `blocking`, and `futures` modules. They should probably be merged.


src/futures/spi.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// method.
+    fn read<'a>(&'a mut self) -> Self::ReadFuture<'a>;
+
+    /// Writes a word to the slave
+    fn write<'a>(&'a mut self, word: Word) -> Self::WriteFuture<'a>;
+}
+
+/// Clock polarity
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum Polarity {
```
Dirbaio, line 47: These should probably not be copypasted.

src/futures/spi.rs
```diff
@@ -0,0 +1,295 @@
...
+    type Error;
+
+    /// Associated future for the `transfer` method.
+    type TransferFuture<'a>: Future<Output = Result<&'a [Word], Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Writes `words` to the slave. Returns the `words` received from the slave
+    fn transfer<'a>(&'a mut self, words: &'a mut [Word]) -> Self::TransferFuture<'a>
+    where
```
Dirbaio, line 108: Why `Self::Error: 'a`?

Dirbaio:
Nice to see more interest on async! :) 

Some general feedback:

## 1. Can't use DMA

I see the proposed traits are modeled after the existing `nb` traits, which read/write one word at a time. This rules out implementing them with DMA. CPU needs to get involved for every single word, waking the async task and having it start the next word write. Waking an async task is not free (can take a few us), for fast IO (eg multi-mhz SPI) it's likely that the time to wake the task will be similar to the time to send one word. In this case, the behavior will degrade to "blocking, but crappier": the CPU will have no time to run other tasks anyway, and it'll be slower and clunkier than real blocking.

Note that `nb` traits suffer from this same problem. However this limitation was "okay-ish", since it's impossible to impl `nb` traits with DMA anyway (because they can't borrow a buffer for the entire duration of the operation, only the instant when polled). With futures it's possible to allow DMA impls, so it'd be a mistake not to.

`embassy-traits` reads/writes entire slices ([example](https://github.com/embassy-rs/embassy/blob/9e5406f761a5a71e8a634796a60f1148b7d96ec1/embassy-traits/src/uart.rs#L15)) so that impls can set up DMA to read/write the entire slice in one go and only wake the task when it's fully done.

*However* doing DMA with "naked" slices is unsound *but only if the futures are leaked*. Since DMA is reading/writing directly from the slice, DMA has to be stopped if the future is dropped mid-operation. This is easy, embassy does that no problem. However *in Rust you can leak things (not run Drop) without unsafe*. If you do that, DMA will not be stopped and can keep writing to the buffer after it's deallocated. Detailed explanation of the problem [here](https://docs.rust-embedded.org/embedonomicon/dma.html).

To fully solve this, the traits would have to use `embedded-dma`, but then the ergonomics heavily suffer, especially without alloc.

Embassy "solves" it by forbidding the user to leak futures. Dropping them is fine, leaking them is not. This means embassy's API is not fully safe, but in practice it's really really easy to use it correctly, since there's only a handful of ways to leak things (`mem::forget`, `ManuallyDrop`, `Box::leak`, `Rc/Arc` cycles), and they're all really easy to avoid. IMO this is the right tradeoff: super efficient DMA use, almost-safe API, super ergonomic. I expect this to be *very* controversial for adoption in embedded-hal though, as many people are of the "must be fully safe at any cost" opinion.

The third solution would be to fix the Leak Problem in Rust itself. IMO the Leak Problem is a Rust design issue. There's been some discussions, it might be fixable with a `Leak` auto trait that works similar to `Sized`. However it seems no one is feeling the pain of this as we (embedded) do, as we don't want alloc. For example io-uring has the same problem (io-uring is essentially DMA done by the kernel instead of the hardware) and they solve it by owning the buffers with Box/Vec. 

And no, Pin doesn't prevent the leak problem. Pinning the buffer doesn't. Pinning the future doesn't.

## 2. Maybe not add all traits

I'd wait until there's clear demand of traits before adding them. 

- Async Uart, Spi and I2c are definitely wanted. 
- Async ADC is likely to have no benefit, since usually taking an ADC sample is fast enough that the core won't have time to do useful work on other tasks during it.
- I'm not sure about Capture, its usage is quite niche.

## 3. Lifetimes

Lifetimes are indeed complicated. `embassy-traits` has `where Self: 'a` (I guess this is where you took it from?) because embassy drivers can borrow pins/peripherals instead of owning them, so eg an Uart driver is `Uart<'a, UART1>` instead of the more common `Uart<UART1>`. 

If we drop this requirement then it can be simplified (I think that bound can be simply removed), but I would like to keep it. It's nice that drivers can be non-`'static`. 

(I just wanted to point out the reason behind these bounds, since it's not very obvious).

## 4. Iterator traits

I think iterator-based traits like `WriteIter` shouldn't exist in async flavor because 1. it forces CPU involvement for each word and 2. utility is arguably limited since the iter itself is not async (should be a Stream, which is a whole other can of worms).

Cargo.toml
```diff
@@ -15,6 +15,10 @@ readme = "README.md"
 repository = "https://github.com/rust-embedded/embedded-hal"
 version = "1.0.0-alpha.4" # remember to update html_root_url
 
+[features]
+# Enabling this feature enables the `futures` module using generic associated types (GATs), which are still unstable.
+unstable-gats = []
```
burrbull, line 20: Why not just `futures = []`?

src/futures/spi.rs
lachlansneff: Oops, that was left over from some stuff before.

src/futures/spi.rs
lachlansneff: It seems to me like there should be an agnostic place that has these kinds of items. Should the user just import them from blocking?

lachlansneff:
@Dirbaio 

> With futures it's possible to allow DMA impls, so it'd be a mistake not to.

I agree with this. And I'm personally fine with an unsafe API to improve the ergonomics, as long as it's easy to use safely.

I'm unsure if `FullDuplex` should be copied over to the `futures` module because reading and writing single words to SPI is probably going to be nearly instant.

Cargo.toml
```diff
@@ -17,7 +17,9 @@ version = "1.0.0-alpha.4" # remember to update html_root_url
 
 [features]
 # Enabling this feature enables the `futures` module using generic associated types (GATs), which are still unstable.
-unstable-gats = []
+# Therefore, this feature requires compiling on nightly.
+futures = []
```
Dirbaio, line 21: I think we do need the word "unstable" somewhere, signaling that "it might get breaking changes even in minor releases"

src/futures/spi.rs
ryankurte: given the blocking/nb/async split perhaps we could bubble them up to somewhere shared?

ryankurte:
hey thanks for the PR! looks generally good, and async is going to be great fun when we get it worked out ^_^

i'm in agreement with @eldruin here (https://github.com/rust-embedded/embedded-hal/issues/172#issuecomment-865650608), the GAT approach seems workable to me, definitely need to demonstrate / prove each of them in a couple of use cases ([linux-embedded-hal](https://github.com/rust-embedded/linux-embedded-hal) may be one good HAL candidate as there's already some async in the sub-projects).

@Dirbaio also makes some great points, perhaps we could focus on `SPI` and `I2C` (and `UART`?) to reduce the scope of the initial implementation?

> embassy-traits reads/writes entire slices (example) so that impls can set up DMA to read/write the entire slice in one go and only wake the task when it's fully done.

this approach is what i would expect / seems reasonable to me in that it should be implementable _with or without_ DMA, i don't think it'd be useful to expose `embedded-dma` or other strange constraints in this interface.

the one caveat would seem to be UART, to me the UART read/write would ideally be represented as a stream/sink, though i understand these aren't _yet_ in core. i would be okay with a `futures` dependency for now _if_ there's a `no_std` and `alloc` free way of getting it in. 

i think UART differs from the other peripherals because the transactions are unbounded, it seems like circular buffering / DMA in this case is a configuration / implementation detail that shouldn't need to be exposed to HAL consumers. i can see an argument for a slice based write to avoid the byte-wise copy/flush, but it doesn't seem like there's an unsurprising way to implement a serial read via slice without pushing complexity up to users.

it also doesn't seem like there's an obvious/ergonomic way to address wakers at the moment (https://github.com/rust-embedded/embedded-hal/issues/172#issuecomment-772671241), but the proposed wake-every-poll approach would seem to do the job for now.

For reference, couple of prior works on the topic (that i think y'all have already mentioned, just useful to have a list):
- https://github.com/rust-embedded/embedded-hal/issues/149 - discussion about core::futures
- https://github.com/rust-embedded/embedded-hal/issues/172 - proposal to add core::futures based traits
- https://github.com/rust-embedded/embedded-hal/pull/206 - future/poll based impl
- [Disasm/async-embedded-traits](https://github.com/Disasm/async-embedded-traits) - broadly same GAT approach as proposed here

Dirbaio:
> UART read/write would ideally be represented as a stream/sink

These have the same problem as per-word traits: you need CPU involvement for each word, you can eg write a big slice with DMA. Maybe it could be done with a stream/sink of slices, but that gets ugly fast.

> it doesn't seem like there's an unsurprising way to implement a serial read via slice without pushing complexity up to users.

"Read a slice of words" is the lowest common denominator of what most hardware out there can do, while not sacrificing much:

It allows efficient DMA use if you know how much data you're going to receive upfront.
It still allows stuff like "read until newline" by reading 1-word silces in a loop.

ryankurte:
> These have the same problem as per-word traits: you need CPU involvement for each word, you can eg write a big slice with DMA. Maybe it could be done with a stream/sink of slices, but that gets ugly fast.

yeah, though they'd work well implemented _on top of_ IRQs or DMA and a ring-buffer, i think the ideals differ a bit depending on whether you look at this as _an abstraction for physical UART peripherals_ vs. _a UART abstraction for drivers to use_.

> "Read a slice of words" is the lowest common denominator of what most hardware out there can do, while not sacrificing much:

seems alright so long as we're clear that any call might return a partially filled slice (say you had a `[u8; 1024]` buffer you'd expect to RX data roughly as it came in not to fill the buffer first i think. the opposite case would mean you'd almost always have to use byte-wise reads for responsiveness.)

eldruin:
I would like to note that if we only fill the received array partially, we need to return the number of words read. Alternatively, return a slice to the input array, although this would result in an interface like #286.

Dirbaio:
> seems alright so long as we're clear that any call might return a partially filled slice (say you had a [u8; 1024] buffer you'd expect to RX data roughly as it came in not to fill the buffer first i think. the opposite case would mean you'd almost always have to use byte-wise reads for responsiveness.)

In practice that would degenerate to byte-by-byte read anyway. The future would have to complete as soon as 1 byte has been received, and while the CPU is processing that byte no more bytes would be read. The hardware FIFO for UARTs is typically tiny, usually just 1..4 bytes.

The "allow short writes" solution makes a lot of sense when the IO is backed by a big buffer, and read calls can copy big chunks off it. That's what `std::io::Read` does. However I don't think forcing implementations to buffer is a good idea. With DMA hardware it's nice that the hardware can read/write directly to the slices provided by the user.

`embassy-nrf` has two uart drivers: `Uarte` implementing `uart::Read` and `uart::Write`, which wait for the entire slice to be read/written and can therefore do zerocopy IO, and `BufferedUarte` which has ringbuffers hardware reads/writes to, and impls `io::AsyncRead`, `io::AsyncWrite`. These copy to/from the ringbuffers, so it's not zerocopy IO, but can be more performant since the hardware can be constantly reading/writing to the ringbuffer even when the CPU is busy with other things.

Embassy also has a `ReadUntilIdle` trait, although I think that one is maybe too specific: https://github.com/embassy-rs/embassy/blob/d49adc98beebe5b7982204150ebdc0a485fefab2/embassy-traits/src/uart.rs#L23-L25

It really depends on the use case. I don't think there's One Uart Trait To Rule Them All.

- Terminal for a human to type in? use `uart::Read` with 1-byte slices. Humans are slow, you won't miss keypresses.
- Custom protocol where you reasonably control the timing and frame sizes? maybe you can get away with `uart::Read` or `uart::ReadUntilIdle`, which is nice because you get zerocopy IO.
- Complex protocol, hard to predict timings/sizes, want max performance? use `BufferedUarte` with `AsyncRead/AsyncWrite`.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Added `IoPin` trait for pins that can change between being inputs or outputs
   dynamically.
+- Added `futures` module that contains async traits (using unstable GATs) for I2C, RNG, Serial, SPI, and Timers.
```
eldruin, line 13: Could you mention here that it is behind the `unstable-futures` flag?

src/futures/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Asynchronous APIs
+//!
+//! This traits use `core::future::Future` and generic associated types.
```
eldruin, line 3: The whole `futures` approach should be documented here. Including examples, executors, waking, etc.

CHANGELOG.md
eldruin, line 13: Could you mention here that it is behind the `unstable-futures` flag?

eldruin:
```suggestion
- Added `futures` module that contains asynchronous traits (using currently unstable GATs) for I2C, RNG, Serial, SPI, and Timers. These traits are behind the feature flag `unstable-futures`. The `futures` module currently needs Rust nightly and it is not included in `embedded-hal`'s SemVer guarantees. We may release breaking changes in any patch release. If you use this module, please use an `=1.x.x` crate version specification.
```

src/futures/spi.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// Associated future for the `transfer_inplace` method.
+    type TransferInPlaceFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Writes `words` to the slave from the `readwrite` buffer and reads words into the same buffer.
+    /// This method uses a single `readwrite` buffer.
+    ///
+    /// The returned buffer is the initialized `readwrite` buffer.
+    fn transfer_inplace<'a>(&'a mut self, readwrite: &'a mut [W]) -> Self::TransferInPlaceFuture<'a>;
```
eldruin, line 34: The `readwrite` buffer should probably be named `words` or similar, so that no operation order is implied.

src/futures/timer.rs
```diff
@@ -0,0 +1,101 @@
+//! Timers
+
+use core::future::Future;
+
+/// A count down timer
+///
+/// # Contract
+///
+/// - `self.start(count); block!(self.wait());` MUST block for AT LEAST the time specified by
```
eldruin, line 9: The whole documentation here needs to be adapted to using futures, including examples.

eldruin (commented):
As described [here](https://github.com/rust-embedded/embedded-hal#how-to-add-a-new-trait), in order to prove that the traits work in practice, for each trait, two implementations and one generic driver using them are necessary.
I may be able to help with I2C and SPI drivers once there is a HAL implementation I can use somewhere for something like an STM32F1/F3/F4/H7, nRF51/52 or an RPi.
We can skip `rng`, though, as we will remove it soon (#291)

src/futures/mod.rs
lachlansneff: I agree with those — that being said I don't think it's quite sorted out how HALs are going to support this. I think they definitely can (and will), but it's going to require some thought as to how HALs should install interrupt handlers and whatnot for async operations to work.

CHANGELOG.md
lachlansneff: This looks great

src/futures/mod.rs
eldruin: Sure. There is no "_one true answer_" as for how all operations should concert in all situations. However, the different alternatives and considerations about how to _implement_ and _consume_ these traits as well as what the expectations are about how the operations will _run_ should be explained here.

src/futures/digital.rs
```diff
@@ -0,0 +1,50 @@
...
+/// ```
+pub trait AsyncInputPin: InputPin {
+    /// The future returned by the `until_high` function.
+    type UntilHighFuture<'a>: Future<Output=()> + 'a;
+
+    /// The future returned by the `until_low` function.
+    type UntilLowFuture<'a>: Future<Output=()> + 'a;
+
+    /// Returns a future that resolves when this pin becomes high.
+    fn until_high<'a>(&self) -> Self::UntilHighFuture<'a>;
```
Dirbaio, line 46: Requiring `&self` instead of `&mut self` means multiple tasks can wait on the same pin concurrently. This will require implementations to support storing an arbitrary amount of wakers, which is generally not feasible.

src/futures/digital.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// Returns a future that resolves when this pin becomes high.
```
Dirbaio, line 48: becomes low

src/futures/spi.rs
```diff
@@ -0,0 +1,67 @@
...
+pub trait Transfer<W: 'static> {
+    /// Error type
+    type Error;
+
+    /// Associated future for the `transfer` method.
+    type TransferFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Writes `words` to the slave from the `write` buffer. Puts the words returned in the `read` buffer.
```
Dirbaio, line 15: this should document what happens when the buffers are different size, same as the blocking trait.

src/futures/serial.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Read error
+    type Error;
+
+    /// The future associated with the `read` method.
+    type ReadFuture<'a>: Future<Output=Result<Word, Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Reads a single word from the serial interface
+    fn read<'a>(&'a mut self) -> Self::ReadFuture<'a>;
```
Dirbaio, line 19: I still think uart read/write should take slices like all other traits. Writing/reading word-by-word is super slow, especially in nRF chips which can *only* do DMA.

src/futures/digital.rs
lachlansneff: Good point, hadn't thought of that.

What do you think of the naming? Is there a better name than `until_high|low`?

src/futures/digital.rs
```diff
@@ -0,0 +1,57 @@
...
+///     futures::select_biased! {
+///         x => ready_pin.until_high() => {
+///             x.expect("failed to await input pin");
+///             true
+///         },
+///         _ => delay.delay(Duration::from_millis(1)) => false, // ignore the error
+///     }
+/// }
+/// ```
+pub trait AsyncInputPin: InputPin {
```
Dirbaio, line 45: - What's the benefit of having the `InputPin` bound? Users can bound on both if they want both, it seems a bit unnecessary restriction to me.
- Name shouldn't have "async", all traits in the futures module are async
- Name being "InputPin" suggests this is the async equivalent of InputPin, namely having is_high/is_low functions but async. That's not what the trait does. I like the naming in `embassy_traits` much more: `WaitForHigh`, `WaitForLow`. It makes it clear what it does, and it makes method names to be verbs like all the other traits.

src/futures/digital.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// The future returned by the `until_high` function.
+    type UntilHighFuture<'a>: Future<Output=Result<(), Self::Error>> + 'a;
+
+    /// The future returned by the `until_low` function.
+    type UntilLowFuture<'a>: Future<Output=Result<(), Self::Error>> + 'a;
+
+    /// Returns a future that resolves when this pin becomes high.
```
Dirbaio, line 52: Perhaps clarify that it waits for the pin to *be* high, not to undergo a low-to-high *transition*. ie if the pin is already high it returns immediately.

src/futures/serial.rs
lachlansneff: How does DMA work when the slices would be different sizes? Is that something DMA on MCUs supports?

src/futures/digital.rs
Dirbaio: just commented about the names in a new thread

src/futures/digital.rs
lachlansneff: These are all good points. I'm not sure it should be two different traits (e.g. `WaitForHigh`, `WaitForLow`) though. Maybe `WaitForPin`?

src/futures/digital.rs
lachlansneff: Good call.

src/futures/serial.rs
Dirbaio: I think you meant to ask that on SPI?

lachlansneff: Oops, yep. As for this, I'm fine with that, slice it is.

src/futures/spi.rs
lachlansneff: How does DMA work when the slices would be different sizes? Is that something DMA on MCUs supports?

src/futures/digital.rs
lachlansneff: One worry I have here is that the future might need to have some state here. On the first poll call, it'll need to check, but it shouldn't after that. It sounds a little error-prone.

src/futures/digital.rs
Dirbaio: No idea! there might be some cursed hardware out there that assumes external interrupts are active-low or something, so I made it 2 separate traits just in case in embassy. All the other traits are one trait per function too.

lachlansneff: One other thing: if this trait doesn't have an `InputPin` bound, then it _could_ have a different error type, which would result in people using the `P: InputPin + WaitFor<Error = <P as InputPin>::Error>` (if that even works) bound. I'm not a fan of there being an implicit agreement that these have the same error type.

src/futures/spi.rs
Dirbaio: Some hardware supports this out of the box. For example nrf will pad write with zeros or ignore extra read data. Others like stm32 might need impls to emulate it. For example if read is longer than write, start the buf write and when it's done start DMA again writing zeros.

Either way docs should specify whether different sizes are allowed or not, and what the behavior is if they are. #287 is still stuck on discussing this, so I'd just mirror that.

src/futures/digital.rs
Dirbaio: Some hardware allows you to directly wait for high/low (nrf's GPIOTE). Some doesn't (stm32's EXTI) so the impl would have to do extra check yup, and doing them without race conditions is tricky.

However I think it's a good thing to force HALs to deal with that, otherwise the burden shifts to the end user. External chip interrupts are often "active low" and you want `wait_for_low` to return immediately if it's already low. If it waits for a falling edge it's super easy to get stuck in a case where the line is low and you're stuck waiting forever for a falling edge that never comes.

src/futures/digital.rs
Dirbaio: True, but this is already the case with SPI Read/Write/Transfer/TransferInplace. I don't think this is a problem...

lachlansneff: > there might be some cursed hardware out there that assumes external interrupts are active-low or something, so I made it 2 separate traits just in case in embassy.

Seems reasonable. Will do two traits then.

src/futures/digital.rs
lachlansneff: Thanks for the explanation.

src/futures/digital.rs
lachlansneff: > One other thing: if this trait doesn't have an `InputPin` bound, then it _could_ have a different error type, which would result in people using the `P: InputPin + WaitFor<Error = <P as InputPin>::Error>` (if that even works) bound. I'm not a fan of there being an implicit agreement that these have the same error type.

Maybe this trait shouldn't have an error type?

src/futures/digital.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use core::future::Future;
+
+/// Asynchronously wait for a pin to be high.
+pub trait WaitForHigh {
+    /// Enumeration of errors.
+    type Error;
+
+    /// The future returned by the `wait_for_high` function.
+    type Fut<'a>: Future<Output = Result<(), Self::Error>> + 'a
```
Dirbaio, line 48: Naming the associated future type `Fut` will cause naming conflicts if a type `Pin` impls both `WaitForHigh` and `WaitForLow` and the user tries to do `Pin::Fut`.

It's a very niche edge case, since users will usually `.await` the futures directly instead of storing them somewhere, but still.

I think the best convention is: for a method named `foo_bar`, name the associated future `FooBarFuture`. All other traits follow this convention, imo these here should too

Dirbaio (changes requested):

src/futures/digital.rs
lachlansneff: Yeah, I guess this makes sense. It seems so verbose to list out the entire method in the associated type, but it avoids naming conflicts as you said.

lachlansneff:
Now that [GATs are no longer an incomplete feature](https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html), it's looking likely that they'll land before the end of the year! Maybe it's time to get this in as an `unproven` feature?

eldruin:
We have decided against an `unproven` feature. See [here](https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary).
The best way would be to demonstrate that these traits work well by creating HAL implementations for them and also demonstrating their use in a driver.

burrbull:
I propose to hide them under `async` feature anyway and add doc that this is unstable.
Before stabilization it will prevent us to use unstable feature.
After stabilization this will enable "additional" functionality.
Unlike `unproven` feature we don't need to delete it ever.

lulf:
@lachlansneff I see there is one unresolved comment from @Dirbaio. I volunteer continuing this work if you don't have time for it.  

@Dirbaio From embassy POV, do you think these traits look sufficiently complete that we can give a go at making the embassy implementations support them? (I can make a PR for it)

Dirbaio:
@lulf Yes, the traits here look great and I don't think there will be any issue switching embassy to them. It'd be great to do it, it'll help test the traits :) 

I have one major concern about merging this PR: [async fn in traits](https://rust-lang.github.io/async-fundamentals-initiative/roadmap/mvp.html) is coming. It'll allow using actual `async fn`s, which will desugar to GAT + regular fn like the traits in this PR. It is very likely that `async fn` traits will become the idiomatic way, and switching from GATs to `async fn` will be a breaking change.

IMO we should release this in a separate crate, for example `async-embedded-hal`. This crate can get breaking changes as needed, releasing v0.1, v0.2, v0.3... When we consider it "proven", instead of releasing a v1.0 we can merge it into the main `embedded-hal` as originally planned.

Gating the async traits behind an `async` feature won't work if we plan to make breaking changes to them, even if we declare "if you enable the `async` feature you're not covered by semver". There will be cases where e.g. async traits got broken in v1.1 -> v1.2 so you have to pin `embedded-hal = "=1.1"`, but v1.2 added some unrelated blocking trait that your HAL wants to implement so your HAL requires v1.2 and now you can't build at all.

lachlansneff:
@lulf 
> @lachlansneff I see there is one unresolved comment from @Dirbaio. I volunteer continuing this work if you don't have time for it.

I actually already made his suggested changes, just forgot to resolve the comment. I'm with @Dirbaio on this: a separate crate `async-embedded-hal` or `embedded-hal-async` would be the right move until async trait is in nightly. I'll go ahead and make the crate in the next day or so.

If anyone has an opinion about the name, I'm fine with either.

burrbull:
+ on `embedded-hal-async`. Easier to find on crates.io

derekdreery:
> I have one major concern about merging this PR: async fn in traits is coming. It'll allow using actual async fns, which will desugar to GAT + regular fn like the traits in this PR. It is very likely that async fn traits will become the idiomatic way, and switching from GATs to async fn will be a breaking change.

I've had the following issue with SPI using embassy: I want to write out bits to a display as quickly as possible (since there are a lot to write and my SPI is not that fast). To do this, I want to use 2 buffers for my DMA. I want one buffer to be DMAd out to the peripheral while I fill the other.

Now a difference between async fn and futures with GATs is that the futures are a concrete type. This type could have a method like `start_dma` or similar that starts the operation immediately. For the async fn version, remember *futures do nothing until polled*. So we have to poll the future (`await` the async fn). This complicates how you implement the parallel write mentioned above.

So basically what I'm saying is that I'd like a concrete type for the future, and then a method like `start_dma` that will eagerly start writing to the peripheral without waiting for the future to be polled. Maybe there is an elegant solution I've missed, if so please tell me, but otherwise, I feel like we're losing ergonomics if we use `async fn`, at least for my use case.

lulf:
@lachlansneff I think starting with the embedded-hal-async is nice. What is the preferred way to go about this? Merge this PR and reorganize the e-h into two crates then or do we want to see that reorg happening in this PR?

lachlansneff:
> @lachlansneff I think starting with the embedded-hal-async is nice. What is the preferred way to go about this? Merge this PR and reorganize the e-h into two crates then or do we want to see that reorg happening in this PR?

Sorry, the semester just finished, so I didn't have time to set this up. I have it half set up on my laptop. Was thinking to just create the crate and drop this PR.

ryankurte:
given the fundamentals (GATs for now, async trait one day) are pretty clear now i wonder whether it might be useful for us (@rust-embedded/hal) to create the `embedded-hal-async` crate here with the usual tooling and a bunch of warnings about it's alpha-ness? then we could take a similar approach to `e-h` to collaboratively build it up component by component.

Dirbaio:
I've opened a PR adding `embedded-hal-async`.  https://github.com/rust-embedded/embedded-hal/pull/334

Dirbaio:
This can probably be closed, now that we have `embedded-hal-async`.

eldruin:
Indeed. Thank you everybody!

PR #286, opened 2021-06-22
blocking/spi: Don't return the same buffer back.

Dirbaio:
This is a rather minor improvement to the blocking SPI Transfer trait, but that IMO is still worth it.

Old: `fn transfer<'w>(&mut self, words: &'w mut [W]) -> Result<&'w [W], Self::Error>`
New: `fn transfer(&mut self, words: &mut [W]) -> Result<(), Self::Error>`

It is redundant (and confusing IMO) to return back the same buffer the user has passed in. It allows for 2 ways of writing the same thing: use the returned buffer, or ignore it and use the buffer that was passed in. It's not obvious which is better, and from the docs it's not obvious that both are really the same either.

Also, with the old signature, it is possible for implementors to return a subslice of the buffer instead of the whole buffer. This would be arguably wrong and would break user code that assumes the entire buffer is returned. Not re-returning the buffer makes this mistake impossible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `blocking::Serial`: renamed `bwrite_all` to `write`, `bflush` to `flush.
 - Removed `prelude` to avoid method name conflicts between different flavors (blocking, nb) of the same trait. Traits must now be manually imported.
 - Removed the various `Default` marker traits.
+- Removed `&[W]` returned slice in `spi::blocking::Transfer`.
 
 ### Removed
 - Removed random number generation (`rng`) traits in favor of [rand_core](https://crates.io/crates/rand_core).
--- a/src/blocking/spi.rs
+++ b/src/blocking/spi.rs
@@ -5,8 +5,10 @@ pub trait Transfer<W> {
     /// Error type
     type Error;
 
-    /// Writes `words` to the slave. Returns the `words` received from the slave
-    fn transfer<'w>(&mut self, words: &'w mut [W]) -> Result<&'w [W], Self::Error>;
+    /// Writes and reads simultaneously. The contents of `words` are
+    /// written to the slave, and the received words are stored into the same
+    /// `words` buffer, overwriting it.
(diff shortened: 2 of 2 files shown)


lachlansneff:
One thing that I just thought of here. If the read buffers get switched over to `MaybeUninit`, then returning the buffer as a slice of words lets people use the api without unsafe. There is the issue of returning a subslice though.

Dirbaio:
True!

The propsed change is what I feel makes most sense *given the current convention* of not using MaybeUninit or similar.

Maybe you could open an issue on "add a way to read into uninitialized buffers" and we can discuss there :)

David-OConnor:
Agreed. I noticed this recently as well, and was surprised.

Dirbaio:
ping @eldruin  @ryankurte @therealprof 

I believe this should be fairly uncontroversial. The question of "what about `MaybeUninit`" is better debated in its own issue, and the verdict from that applied to all traits. I don't think it justifies keeping the returned slice here "just in case": no other trait does this, and we can always add it back if we do `MaybeUninit`.

eldruin (approved):
Looks good to me, thanks!

therealprof:
@japaric Since you added the API way back when, did you have any intentions with regards to this particular interface? Any objections to removing it?

ryankurte:
i guess without the returned subslice (or an `n`) you can't easily signal partial-success (eg, tried to transfer four bytes, only achieved three), but, it'd probably be less surprising to signal that as an error (especially with the new error types) than to require people to -also- check `n` on success. pending any thoughts by @japaric, sgtm.

therealprof:
As discussed last week we're going to merge this pending objections. @Dirbaio Would you mind fixing the conflicts?

Dirbaio:
@therealprof rebased

eldruin:
uh, CI is broken due to the `stm32f1` dev dependency. Updating it does not help: #304

eldruin:
Alright, #305 should fix CI, then we can merge this.

therealprof (commented):
bors try

therealprof (approved):
LGTM, thanks!

bors r+

PR #287, opened 2021-06-22
spi: add Read and separate-buffers Transfer

Dirbaio:
~~Depends on #286~~

- Added spi::Read for only reading
- Renamed Transfer to TransferInplace
- Added a new Transfer, for reading+writing simultaneously with different buffers.

Open question
- ~~Should Transactional gain Read, Transfer (not-inplace) support?~~ yes, done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   dynamically.
 - Added `Debug` to all spi mode types.
 - Add impls of all traits for references (`&T` or `&mut T` depending on the trait) when `T` implements the trait.
+- SPI: Added blocking `Read` trait and `Read` transactional operation
+- SPI: Added blocking `Transfer` trait with separate buffers (single-buffer `Transfer` has been renamed `TransferInplace`)
 
 ### Changed
 - Swap PWM channel arguments to references
--- a/src/spi/blocking.rs
+++ b/src/spi/blocking.rs
@@ -1,21 +1,65 @@
 //! Blocking SPI API
 
-/// Blocking transfer
+/// Blocking transfer with separate buffers
 pub trait Transfer<W = u8> {
+    /// Error type
+    type Error: crate::spi::Error;
+
+    /// Writes and reads simultaneously. `write` is written to the slave on MOSI and
+    /// words received on MISO are stored in `read`.
+    ///
+    /// It is allowed for `read` and `write` to have different lengths, even zero length.
+    /// The transfer runs for `max(read.len(), write.len())` words. If `read` is shorter,
+    /// incoming words after `read` has been filled will be discarded. If `write` is shorter,
+    /// the value of words sent in MOSI after all `write` has been sent is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
+    fn transfer(&mut self, read: &mut [W], write: &[W]) -> Result<(), Self::Error>;
+}
+
+impl<T: Transfer<W>, W> Transfer<W> for &mut T {
+    type Error = T::Error;
+
+    fn transfer(&mut self, read: &mut [W], write: &[W]) -> Result<(), Self::Error> {
+        T::transfer(self, read, write)
+    }
+}
+
+/// Blocking transfer with single buffer (in-place)
+pub trait TransferInplace<W = u8> {
     /// Error type
     type Error: crate::spi::Error;
 
     /// Writes and reads simultaneously. The contents of `words` are
     /// written to the slave, and the received words are stored into the same
     /// `words` buffer, overwriting it.
-    fn transfer(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
+    fn transfer_inplace(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
 }
 
-impl<T: Transfer<W>, W> Transfer<W> for &mut T {
+impl<T: TransferInplace<W>, W> TransferInplace<W> for &mut T {
     type Error = T::Error;
 
-    fn transfer(&mut self, words: &mut [W]) -> Result<(), Self::Error> {
-        T::transfer(self, words)
+    fn transfer_inplace(&mut self, words: &mut [W]) -> Result<(), Self::Error> {
+        T::transfer_inplace(self, words)
+    }
+}
+
+/// Blocking read
+pub trait Read<W = u8> {
+    /// Error type
+    type Error: crate::spi::Error;
+
+    /// Reads `words` from the slave.
+    ///
+    /// The word value sent on MOSI during reading is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
+    fn read(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
+}
+
+impl<T: Read<W>, W> Read<W> for &mut T {
+    type Error = T::Error;
+
+    fn read(&mut self, words: &mut [W]) -> Result<(), Self::Error> {
+        T::read(self, words)
     }
 }
 
@@ -63,10 +107,14 @@ impl<T: WriteIter<W>, W> WriteIter<W> for &mut T {
 /// This allows composition of SPI operations into a single bus transaction
 #[derive(Debug, PartialEq)]
 pub enum Operation<'a, W: 'static = u8> {
+    /// Read data into the provided buffer.
+    Read(&'a mut [W]),
     /// Write data from the provided buffer, discarding read data
     Write(&'a [W]),
     /// Write data out while reading data into the provided buffer
-    Transfer(&'a mut [W]),
+    Transfer(&'a mut [W], &'a [W]),
+    /// Write data out while reading data into the provided buffer
+    TransferInplace(&'a mut [W]),
 }
 
 /// Transactional trait allows multiple actions to be executed


src/blocking/spi.rs
```diff
@@ -1,12 +1,30 @@
 //! Blocking SPI API
 
-/// Blocking transfer
-pub trait Transfer<W> {
+/// Blocking simultaneous read+write with separate buffers
+pub trait ReadWrite<W> {
```
lachlansneff, line 4: To be consistent with the async traits, can we rename `W` to `Word`?

Dirbaio: existing code uses a mix of `W` and `Word`. I do agree it should all be consistent, I like `W` somewhat more. It makes it more obvious that it's a generic param vs an actual type, since generic params are usually single uppercase letters.

src/blocking/spi.rs
```diff
@@ -18,6 +36,16 @@ pub trait Write<W> {
...
 }
 
+/// Blocking read
+pub trait Read<W> {
+    /// Error type
+    type Error;
+
+    /// Reads `words` to the slave. The word value sent on MOSI during
+    /// reading is implementation defined, typically `0x00`, `0xFF`, or configurable.
+    fn read(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
```
lachlansneff, line 46: Also, I'd suggest that this take `&mut [MaybeUninit<Word>]`.

src/blocking/spi.rs
lachlansneff: I'm fine with either, I just want consistency.

lachlansneff: My one issue with `W` is that `W` is often used for type parameters that implement `std::io::Write`.

src/blocking/spi.rs
Dirbaio: Agreed it'd be nice! However, it should be on all "read" traits, and on its own RFC/PR so it can be discussed properly independently.

lachlansneff: Ah, gotcha. Makes sense.

src/blocking/spi.rs
```diff
@@ -29,26 +57,59 @@ pub trait WriteIter<W> {
...
+        W: Clone + core::default::Default,
+    {
+        type Error = S::Error;
+
+        fn read_write(&mut self, read: &mut [W], write: &[W]) -> Result<(), S::Error> {
+            for i in 0..core::cmp::max(read.len(), write.len()) {
+                let word_out = if i < write.len() {
+                    write[i].clone()
+                } else {
+                    W::default()
```
eldruin, line 78: The value sent here should be documented.

src/blocking/spi.rs
```diff
@@ -76,6 +137,29 @@ pub mod write {
...
+    impl<W, S> crate::blocking::spi::Read<W> for S
+    where
+        S: Default<W>,
+        W: core::default::Default,
+    {
+        type Error = S::Error;
+
+        fn read(&mut self, words: &mut [W]) -> Result<(), S::Error> {
+            for word in words.iter_mut() {
+                nb::block!(self.write(W::default()))?;
```
eldruin, line 155: The value written here should be documented.

eldruin (commented):
Other than my comments below, looks good to me, thanks!
Closes #286

src/blocking/spi.rs
Dirbaio: Done! I've added to the `Default` docs

src/blocking/spi.rs
Dirbaio: done

src/blocking/spi.rs
Dirbaio: discussion in #288

eldruin (dismissed):
Looks good to me, thanks!
Any objections @rust-embedded/hal ?

src/blocking/spi.rs
```diff
@@ -29,26 +57,62 @@ pub trait WriteIter<W> {
...
+    pub trait Default<W>: crate::nb::spi::FullDuplex<W> {}
+
+    impl<W, S> crate::blocking::spi::ReadWrite<W> for S
+    where
+        S: Default<W>,
+        W: Clone + core::default::Default,
+    {
+        type Error = S::Error;
+
+        fn read_write(&mut self, read: &mut [W], write: &[W]) -> Result<(), S::Error> {
```
ryankurte, line 76: the reason for `transfer` is _i believe_ because the operation is simultaneous (duplex), rather than sequential (simplex). as an example in the context of the hal, `i2c::write_read` means a write _then_ a read, i believe renaming this for SPI actually makes the operation less clear.

src/blocking/spi.rs
```diff
@@ -18,6 +36,16 @@ pub trait Write<W> {
     fn write(&mut self, words: &[W]) -> Result<(), Self::Error>;
 }
 
+/// Blocking read
+pub trait Read<W> {
```
ryankurte, line 40: previously we elected to not offer a `read` type as the behavior of the data output would be undefined (or dependent on your hardware) which could lead to surprising behavior on real peripherals, where a `transfer` with zero's (or other value) in the output buffer is more explicit.

is there a new / improved argument for adding this?

https://github.com/rust-embedded/embedded-hal/issues/52

ryankurte (commented):
thanks for the PR! i'm not convinced all of these are improvements. have left a couple of comments, though it's possible i missed some discussion of this already?

src/blocking/spi.rs
Dirbaio: True. IMO the "transfer" name is a bit confusing since arguably all the SPI traits do different kinds of transfers: read transfer, write transfer, readwrite transfer... Having one of them named "transfer" feels strange.

No strong opinion though, we can name them `Transfer` and `TransferInplace` too. The main change I wanted to propose was adding the "separate buffers" transfer trait and the Read trait.

src/blocking/spi.rs
Dirbaio: :sweat_smile: I hadn't seen that issue.

It's true that leaving it implementation-defined limits the usefulness for hal-independent drivers. Maybe it could be specified to always send 0?

The usefulness I see is:
- No need to zero big buffers if we just want to read big data (can use an buffer with whatever data without leaking the data on MOSI)
- If we do add a way to read to uninit buffers (#288) then Read would be needed
- When using an SPI with only MISO pin, no MOSI pin.

Maybe it's better to leave it for later.

src/blocking/spi.rs
ryankurte: > No strong opinion though, we can name them Transfer and TransferInplace too. The main change I wanted to propose was adding the "separate buffers" transfer trait and the Read trait.

imo at the hardware there is really only transfer, everything else is convenience above that. regardless it's different to other APIs so i'd prefer this / to keep the names different.

src/blocking/spi.rs
ryankurte: > It's true that leaving it implementation-defined limits the usefulness for hal-independent drivers. Maybe it could be specified to always send 0?

so long as implementations support it, [linux](https://github.com/spotify/linux/blob/master/include/linux/spi/spidev.h#L54) appears to be okay with shifting zeros.

> No need to zero big buffers if we just want to read big data

yep, i suppose there's a use case for this with SPI flash.

> If we do add a way to read to uninit buffers (Methods to read into a buffer should take an uninitialized buffer #288) then Read would be needed

won't `transfer` also take a `MaybeUninit` read buffer in this case?

> When using an SPI with only MISO pin, no MOSI pin

is this an thing one would actually do? :-P


i'm alright with adding a `Read` (trait and `Transational` variant) if it's clearly documented that the write-data will be zeroed, @rust-embedded/hal any other opinions?

Dirbaio: > won't transfer also take a MaybeUninit read buffer in this case?

Single-buf Transfer would take a `&mut [u8]`. Multi-buf Transfer would take `read: &mut [MaybeUninit<u8>], write: &[u8]`. Still not useful to read a big buf without having to zero-out a big write buf, only `Read` can do that.

> is this an thing one would actually do? :-P

People do weird hacks with readonly SPI. For example a fancy logic analyzer that bit-packs the samples for you... however in that case the user will be OK with using the HAL directly or even the PAC, so this is a weak argument for adding it to embedded-hal, yup.

adamgreig: Could we just say the write data will be undefined, and implementations might use 0, 1, or whatever? If you need the write data to be something particular you can use the transfer method and specify it; if you're just reading and don't care what's written then it should be OK to let the implementation decide whatever's best. I guess we can also provide a default impl of read in terms of transfer.

If you were in a MISO-only situation then the guarantee wouldn't even make sense since MOSI wouldn't have a defined level, and in any event I think it's typical that if you're reading you really don't care what's being written. The same is true of half-duplex buses using one wire for combined MISO and MOSI, where the master won't drive the output at all while reading - so again there's no defined level.

Still, I guess implementations have to pick something and I can't think of any practical situation where it's easier for the implementation to use 1 instead of 0 but in such a way that the actual application doesn't care. So maybe it's less confusing to just require writing all 0s.

src/blocking/spi.rs
adamgreig: The other verb I've seen used for this sort of method name is `exchange` (for example, that's what ChibiOS uses for this).

FreeRTOS uses `transfer`, STM32CubeMX uses `TransmitReceive`, libopencm3 uses `xfer`, Contiki uses `transfer`, Zephyr calls it `spi_transceive`.

eldruin: Good point with the comparison to I2C's method name @ryankurte.
I would also prefer `transfer`/`transfer_inplace` or `exchange`, `exchange_inplace`. My gut feeling is that `exchange` sounds more of a two-way transaction than `transfer`, which sounds more one-way-y, like a (monetary) bank transfer.
But anyway I am no native speaker and there seems to be a bit more usage of the `transfer` verb (thanks for the research @adamgreig!). I would be fine with either.

Dirbaio: Renamed to `Transfer`, `TransferInplace`. (I think "transfer" and "exchange" are equally good, but Transfer makes more sense, to minimize the name changes in 1.0)

src/blocking/spi.rs
Dirbaio: on nRF it's configurable. I'd prefer specifying it as "implementation defined" so it leaves room for HALs to add a method to configure it.

burrbull:
Merge conflicts

lachlansneff:
Any chance of this getting into 1.0?

Dirbaio:
Rebased. Now that Default is gone, the diff is much smaller :D

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `IoPin` trait for pins that can change between being inputs or outputs
   dynamically.
 - Added `Debug` to all spi mode types.
+- spi: added `Read`
+- spi: added `Transfer` with separate buffers (single-buffer `Transfer` is now `TransferInplace`)
```
eldruin, line 20:
```suggestion
- SPI: Added blocking `Read` trait and `Read` transactional operation
- SPI: Added blocking `Transfer` trait with separate buffers (single-buffer `Transfer` has been renamed `TransferInplace`)
```

src/spi/blocking.rs
```diff
@@ -4,15 +4,41 @@
...
     /// Error type
     type Error: core::fmt::Debug;
 
+    /// Writes and reads simultaneously. `write` is written to the slave on MOSI and
+    /// words received on MISO are stored in `read`.
+    ///
+    /// It is allowed for `read` and `write` to have different lengths, even zero length.
+    /// The transfer runs for `max(read.len(), write.len())` words. If `read` is shorter,
+    /// incoming words after `read` has been filled will be discarded. If `write` is shorter,
+    /// the value of words sent in MOSI after all `write` has been sent is implementation defined,
```
eldruin, line 18:
```suggestion
    /// the value of words sent in MOSI after all `write` has been sent is implementation-defined,
```

src/spi/blocking.rs
```diff
@@ -4,15 +4,41 @@
...
+    fn transfer_inplace(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
+}
+
+/// Blocking read
+pub trait Read<W> {
+    /// Error type
+    type Error;
+
+    /// Reads `words` to the slave. The word value sent on MOSI during
+    /// reading is implementation defined, typically `0x00`, `0xFF`, or configurable.
```
eldruin, line 40:
```suggestion
    /// reading is implementation-defined, typically `0x00`, `0xFF`, or configurable.
```

eldruin (commented):
Other than my nitpicks, looks good to me, thanks!
@ryankurte do you have any further concerns about implementation-defined fill values?

src/spi/blocking.rs
```diff
@@ -4,15 +4,41 @@
...
     type Error: core::fmt::Debug;
 
+    /// Writes and reads simultaneously. `write` is written to the slave on MOSI and
+    /// words received on MISO are stored in `read`.
+    ///
+    /// It is allowed for `read` and `write` to have different lengths, even zero length.
+    /// The transfer runs for `max(read.len(), write.len())` words. If `read` is shorter,
+    /// incoming words after `read` has been filled will be discarded. If `write` is shorter,
+    /// the value of words sent in MOSI after all `write` has been sent is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
```
ryankurte, line 19: has this always been the case or have we implicitly been assuming `read.len() == write.len()`? and is this sensible to represent for more complex implementors (ie. using DMA, linux)?

we have `Transactional` for building sequences, if you need to do varying length things, it would seem to me to be simpler (or maybe, [less surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) to demand the same lengths here?

src/spi/blocking.rs
```diff
@@ -4,15 +4,41 @@
...
+/// Blocking transfer with single buffer (in-place)
+pub trait TransferInplace<W> {
+    /// Error type
+    type Error: core::fmt::Debug;
+
     /// Writes and reads simultaneously. The contents of `words` are
     /// written to the slave, and the received words are stored into the same
     /// `words` buffer, overwriting it.
-    fn transfer(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
+    fn transfer_inplace(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
```
ryankurte, line 31: :+1:

src/spi/blocking.rs
Dirbaio: This was [discussed](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$ONZJiK_kqPu2beGIGMicS4e1XWNVhOrxE1P_C5gZQEs?via=matrix.org&via=psion.agg.io&via=mozilla.org) in today's REWG meeting. TLDR: seems people are OK with differing lengths.

Also the possibility of allowing impls to support it or not was discussed, but IMO it defeats the point since drivers wouldn't be able to use it.

eldruin:
@Dirbaio This is good to go! Could you resolve the conflict?

Dirbaio:
@eldruin rebased

eldruin:
Ach, a new conflict appeared due to the default `u8` PR, could you rebase again @Dirbaio ?

Dirbaio:
nooooo :rofl:

Dirbaio:
@eldruin rebased

eldruin (approved):
Alright, thanks!
bors merge

PR #290, opened 2021-06-30
Add changelog checker for PRs

eldruin:
A bit ironic that adding the checker requires deactivating it, haha

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,20 @@
+on:
+  pull_request_target:
+
+name: Changelog check
+
+jobs:
+  changelog:
+    name: Changelog check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Changelog updated
+        uses: Zomzog/changelog-checker@v1.2.0
+        with:
+          fileName: CHANGELOG.md
+        env:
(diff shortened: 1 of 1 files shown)


eldruin:
bors try

eldruin:
Alright, now it works.
bors try

eldruin:
Alright, now it works
bors try

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,23 @@
+on:
+  push:
+    branches: [ staging, trying, master ]
+  pull_request:
```
Disasm, line 4: Note that you need this to make it work for PRs from forks: https://github.com/stm32-rs/stm32f4xx-hal/blob/master/.github/workflows/changelog.yml#L1-L2

eldruin: Ah, I see. Then I can enable the detailed check version instead of the simple one, thanks!

therealprof (approved):
bors r+

eldruin:
Ah, it seems the changelog checker does not work well with bors since deactivating the changelog check for a PR does not generate any output at all for it in the status and thus bors waits forever for it.
Closing.

adamgreig:
I think you just don't make bors require it, but it can still run for PRs to remind people to add a changelog.

eldruin:
@adamgreig That would be possible, yes. Not great but I guess better than nothing.
Opinions @rust-embedded/hal?

therealprof (approved):
bors r+

PR #291, opened 2021-06-30
Remove RNG traits

### LINKED ISSUE #128: [Discussion] rand::RngCore for blocking RNG trait

eldruin:
Duplicate of #270 adapted to the current status.
The documentation about what happened to the traits will go into the migration guide. See: #284.

Closes #270
Closes #128


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `blocking::Serial`: renamed `bwrite_all` to `write`, `bflush` to `flush.
 - Removed `prelude` to avoid method name conflicts between different flavors (blocking, nb) of the same trait. Traits must now be manually imported.
 
+### Removed
+- Removed random number generation (`rng`) traits in favor of [rand_core](https://crates.io/crates/rand_core).
```
ryankurte, line 24: probably out of scope but (and plausible we already have one) but, it could be neat to have a using rand section in one of our books we could direct people to?

eldruin: That is a good idea. We can then also add a reference to that in the migration guide.
I filed https://github.com/rust-embedded/rust-embedded.github.io/issues/35

ryankurte (approved):
i don't really mind either way, seems fine to me

eldruin:
Ok let's get this merged then.
bors r=ryankurte

PR #293, opened 2021-07-05
Added purpose and requirements to readme

David-OConnor:
These sections towards the top of the readme clarify the purpose and scope of `embedded-hal`. It's intended to show users new to rust embedded how this project fits into the ecosystem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ to access functionality not included here should interact with the intended plat
 3. Provide generic traits for other peripherals that are suitably common, and easy to abstract over
 4. Provide both blocking, and nonblocking traits for all supported communications protocols
 5. Use fallible trait methods when applicable
+6. Document each trait and method with a description, and example code
 
 ## Releases
 
--- a/README.md
+++ b/README.md
@@ -18,9 +18,10 @@ traits, therefore being compatible with any platform that implements them.
 This project supports the most commonly-used features of supported peipherals. Drivers that need
 to access functionality not included here should interact with the intended platform directly.
 
-## Requirements
+## Project goals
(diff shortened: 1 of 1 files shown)


eldruin (commented):
Thank you for this proposal.
However, not to be too harsh but the purpose of `embedded-hal` is already described in the documentation (it can surely be improved but I think this is more eloquently written than in this proposal, in my opinion):

    - Serve as a foundation for building an ecosystem of platform agnostic drivers. Here driver
    means a library crate that lets a target platform interface an external device like a digital
    sensor or a wireless transceiver. The advantage of this system is that by writing the driver as
    a generic library on top of `embedded-hal` driver authors can support any number of target
    platforms (e.g. Cortex-M microcontrollers, AVR microcontrollers, embedded Linux, etc.). The
    advantage for application developers is that by adopting `embedded-hal` they can unlock all
    these drivers for their platform.

Is the problem that this is too high level? or that it is rather buried in the documentation?

This proposal lists which peripherals are modeled right now in `embedded-hal`, however, the goals go further than what is proposed here. See: [design goals](https://github.com/rust-embedded/embedded-hal/blob/1b2ae70618cf9bff9a6f46ba0c60f2ffb6f99737/src/lib.rs#L10)

It may be useful to add these design goals to the readme and put the how to add a trait and similar things further down or in a separate file, though.
Documenting that the methods are to be fallible and that for further functionality the specific platform should be used the  should be are also good points.

David-OConnor:
I agree - didn't read the Rust docs.

eldruin:
Ok, I'll prepare a proposal about this, then.

PR #296, opened 2021-07-13
Add error traits for communication interfaces

### LINKED ISSUE #229: [RFC] Revamp `Error` types
As demonstrated by #181 and other issues there's a real world usability issue with the `Error` type of the `Result` of operations being an associated type. An associated type means that a driver or user of an HAL impl is required to adapt to the custom Error type; while possible for applications this makes universal drivers impossible.

In my estimation the main reason for this was probably caused by `embedded-hal` predating the improvements to Rust type and specifically `Error` handling. Now we do have better possibilities and with the upcoming big 1.0 breakage it should be possible to also revamp the Error types without too much grief.

Currently our trait definition e.g. for I2C looks as follows:
```
pub trait Read {
    /// Error type
    type Error;
    ...
    fn try_read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Self::Error>;
}
```

My proposal would be to instead provide a specific `Error` type such as:
```
#[non_exhaustive]
pub enum Error {
    /// An unspecific bus error occured
    BusError,
    /// The arbitration was lost, e.g. electrical problems with the clock signal
    ArbitrationLoss,
    /// A bus operation received a NACK, e.g. due to the addressed device not being online
    NoAcknowledge,
    /// The peripheral buffer was overrun
    Overrun,
    /// The peripheral buffer ran out of data
    Underrun,
}
```

and then redefine all traits as:

```
pub trait Read {
    fn try_read(&mut self, address: u8, buffer: &mut [u8]) -> Result<(), Error>;
}
```

The `#[non_exhaustive]` flag (available from Rust 1.40.0, cf. https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html) allows us to easily extend the reasons in the feature without being a breaking change because it requires implementations which care about the actual error to also have a wildcard match branch, which is a good idea anyway.

By having the `Error` types defined in `embeded-hal` we do some advantages:
* Drivers can now universally check the error kind reported by a HAL implementation (cf. #181 where a chip might return a NACK due to currently not being able to handle more requests, yet the driver cannot make use of this information and retry later) and act accordingly
* More uniform and exhaustive documentation of available error conditions next to the API itself
* Better guidelines for HAL authors which cases might be work distinguishing
* Improved portability of embedded applications between different hardware

Dowsides:
* More breakage right now
* MRSV 1.40

eldruin:
There has been a lengthy discussion about this in #229.
This implements proposal 3 by @TeXitoi which was the latest for I2C, SPI and Serial interfaces.
See [here for the analysis](https://github.com/rust-embedded/embedded-hal/issues/229#issuecomment-701258313)
I included a minimal number of variants for each protocol, mostly borrowed from `embedded-error`. 

This raises the MSRV to Rust 1.40.0 as well in order to mark the `ErrorKind`s as `#[non_exhaustive]` so that adding a new variant is not a breaking change.
I am also fine with _not_ marking the `ErrorKind`s as `#[non_exhaustive]`.

Closes #229

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -123,9 +123,10 @@ pub enum ErrorKind {
     Bus,
     /// The arbitration was lost, e.g. electrical problems with the clock signal
     ArbitrationLoss,
-    /// A bus operation was not acknowledged, e.g. due to the addressed device not being available on
-    /// the bus or the device not being ready to process requests at the moment
-    NoAcknowledge,
+    /// The device did not acknowledge its address. The device may be missing.
+    NoAcknowledgeAddress,
+    /// The device did not acknowled the data. It may not be ready to process requests at the moment.
+    NoAcknowledgeData,
     /// The peripheral receive buffer was overrun
     Overrun,
     /// A different error occurred. The original error may contain more information.
@@ -143,7 +144,8 @@ impl core::fmt::Display for ErrorKind {
         match self {
             Self::Bus => write!(f, "Bus error occurred"),
             Self::ArbitrationLoss => write!(f, "The arbitration was lost"),
-            Self::NoAcknowledge => write!(f, "A bus operation was not acknowledged"),
+            Self::NoAcknowledgeAddress => write!(f, "The device did not acknowledge its address"),
+            Self::NoAcknowledgeData => write!(f, "The device did not acknowledge the data"),
             Self::Overrun => write!(f, "The peripheral receive buffer was overrun"),
             Self::Other => write!(
                 f,
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -26,13 +26,13 @@
 //! Here is an example of an embedded-hal implementation of the `Write` trait
 //! for both modes:
 //! ```
-//! # use embedded_hal::i2c::{SevenBitAddress, TenBitAddress, blocking::Write};
+//! # use embedded_hal::i2c::{ErrorKind, SevenBitAddress, TenBitAddress, blocking::Write};
 //! /// I2C0 hardware peripheral which supports both 7-bit and 10-bit addressing.
 //! pub struct I2c0;
 //!
 //! impl Write<SevenBitAddress> for I2c0
 //! {
-//! #   type Error = ();
+//! #   type Error = ErrorKind;
 //! #
 //!     fn write(&mut self, addr: u8, output: &[u8]) -> Result<(), Self::Error> {
 //!         // ...
@@ -42,7 +42,7 @@
 //!
 //! impl Write<TenBitAddress> for I2c0
 //! {
-//! #   type Error = ();
+//! #   type Error = ErrorKind;
 //! #
 //!     fn write(&mut self, addr: u16, output: &[u8]) -> Result<(), Self::Error> {
 //!         // ...
@@ -56,14 +56,14 @@
 //! For demonstration purposes the address mode parameter has been omitted in this example.
 //!
 //! ```
-//! # use embedded_hal::i2c::blocking::WriteRead;
+//! # use embedded_hal::i2c::{blocking::WriteRead, Error};
 //! const ADDR: u8  = 0x15;
 //! # const TEMP_REGISTER: u8 = 0x1;
 //! pub struct TemperatureSensorDriver<I2C> {
 //!     i2c: I2C,
 //! }
 //!
-//! impl<I2C, E: core::fmt::Debug> TemperatureSensorDriver<I2C>
+//! impl<I2C, E: Error> TemperatureSensorDriver<I2C>
 //! where
 //!     I2C: WriteRead<Error = E>,
 //! {
@@ -79,14 +79,14 @@
 //! ### Device driver compatible only with 10-bit addresses
 //!
 //! ```
-//! # use embedded_hal::i2c::{TenBitAddress, blocking::WriteRead};
+//! # use embedded_hal::i2c::{Error, TenBitAddress, blocking::WriteRead};
 //! const ADDR: u16  = 0x158;
 //! # const TEMP_REGISTER: u8 = 0x1;
 //! pub struct TemperatureSensorDriver<I2C> {
 //!     i2c: I2C,
 //! }
 //!
-//! impl<I2C, E: core::fmt::Debug> TemperatureSensorDriver<I2C>
+//! impl<I2C, E: Error> TemperatureSensorDriver<I2C>
 //! where
 //!     I2C: WriteRead<TenBitAddress, Error = E>,
 //! {
@@ -100,6 +100,58 @@
 //! ```
 
 use crate::private;
+
+/// I2C error
+pub trait Error: core::fmt::Debug {
+    /// Convert error to a generic I2C error kind
+    ///
+    /// By using this method, I2C errors freely defined by HAL implementations
+    /// can be converted to a set of generic I2C errors upon which generic
+    /// code can act.
+    fn kind(&self) -> ErrorKind;
+}
+
+/// I2C error kind
+///
+/// This represents a common set of I2C operation errors. HAL implementations are
+/// free to define more specific or additional error types. However, by providing
+/// a mapping to these common I2C errors, generic code can still react to them.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[non_exhaustive]
+pub enum ErrorKind {
+    /// An unspecific bus error occurred
+    Bus,
+    /// The arbitration was lost, e.g. electrical problems with the clock signal
+    ArbitrationLoss,
+    /// A bus operation was not acknowledged, e.g. due to the addressed device not being available on
+    /// the bus or the device not being ready to process requests at the moment
+    NoAcknowledge,
+    /// The peripheral receive buffer was overrun
+    Overrun,
+    /// A different error occurred. The original error may contain more information.
+    Other,
+}
+
+impl Error for ErrorKind {
+    fn kind(&self) -> ErrorKind {
+        *self
+    }
+}
+
+impl core::fmt::Display for ErrorKind {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::Bus => write!(f, "An unspecific bus error occurred"),
+            Self::ArbitrationLoss => write!(f, "The arbitration was lost"),
+            Self::NoAcknowledge => write!(f, "A bus operation was not acknowledged"),
+            Self::Overrun => write!(f, "The peripheral receive buffer was overrun"),
+            Self::Other => write!(
+                f,
+                "A different error occurred. The original error may contain more information"
+            ),
+        }
+    }
+}
 
 /// Address mode (7-bit / 10-bit)
 ///
@@ -119,12 +171,12 @@ impl AddressMode for TenBitAddress {}
 /// Blocking I2C traits
 pub mod blocking {
 
-    use super::{AddressMode, SevenBitAddress};
+    use super::{AddressMode, Error, SevenBitAddress};
 
     /// Blocking read
     pub trait Read<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Reads enough bytes from slave with `address` to fill `buffer`
         ///
@@ -158,7 +210,7 @@ pub mod blocking {
     /// Blocking write
     pub trait Write<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Writes bytes to slave with address `address`
         ///
@@ -190,7 +242,7 @@ pub mod blocking {
     /// Blocking write (iterator version)
     pub trait WriteIter<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Writes bytes to slave with address `address`
         ///
@@ -216,7 +268,7 @@ pub mod blocking {
     /// Blocking write + read
     pub trait WriteRead<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Writes bytes to slave with address `address` and then reads enough bytes to fill `buffer` *in a
         /// single transaction*
@@ -264,7 +316,7 @@ pub mod blocking {
     /// Blocking write (iterator version) + read
     pub trait WriteIterRead<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Writes bytes to slave with address `address` and then reads enough bytes to fill `buffer` *in a
         /// single transaction*
@@ -314,7 +366,7 @@ pub mod blocking {
     /// This allows combining operations within an I2C transaction.
     pub trait Transactional<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Execute the provided operations on the I2C bus.
         ///
@@ -353,7 +405,7 @@ pub mod blocking {
     /// This allows combining operation within an I2C transaction.
     pub trait TransactionalIter<A: AddressMode = SevenBitAddress> {
         /// Error type
-        type Error: core::fmt::Debug;
+        type Error: Error;
 
         /// Execute the provided operations on the I2C bus (iterator version).
         ///
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -119,7 +119,7 @@ pub trait Error: core::fmt::Debug {
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 #[non_exhaustive]
 pub enum ErrorKind {
-    /// An unspecific bus error occurred
+    /// Bus error occurred. e.g. A START or a STOP condition is detected and is not located after a multiple of 9 SCL clock pulses.
     Bus,
     /// The arbitration was lost, e.g. electrical problems with the clock signal
     ArbitrationLoss,
@@ -141,7 +141,7 @@ impl Error for ErrorKind {
 impl core::fmt::Display for ErrorKind {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
-            Self::Bus => write!(f, "An unspecific bus error occurred"),
+            Self::Bus => write!(f, "Bus error occurred"),
             Self::ArbitrationLoss => write!(f, "The arbitration was lost"),
             Self::NoAcknowledge => write!(f, "A bus operation was not acknowledged"),
             Self::Overrun => write!(f, "The peripheral receive buffer was overrun"),
--- a/src/spi/mod.rs
+++ b/src/spi/mod.rs
@@ -53,3 +53,63 @@ pub const MODE_3: Mode = Mode {
     polarity: Polarity::IdleHigh,
     phase: Phase::CaptureOnSecondTransition,
 };
+
+/// SPI error
+pub trait Error: core::fmt::Debug {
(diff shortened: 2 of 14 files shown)


TeXitoi (approved):

therealprof:
@eldruin Is this in line with the plans of the Rust error handling project group? Haven't had time to check, just making sure that we're not working contrary to what's coming up in upstream Rust...

eldruin:
Good point.
This proposal is simpler than that and only includes the already standard `Error::kind()` mechanism, without any kind of chaining, which is what the error WG described [here](https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html).
If we were to offer a mechanism somewhat similar to [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#examples), I think we would need to make the errors contain a generic parameter for the source error, which we have already discussed.
Nevertheless, with this proposal users and HALs can continue to create an error chaining mechanism however they like.

The only thing this proposal does is requiring that whatever error type the HAL implementation returns, a mapping to a set of common error variants exists, upon which generic drivers can act.

ryankurte:
this seems to me like a good synthesis of the work in the related issue! as with @therealprof i wonder how this aligns with the error-handling WG and existing `std:error::Error` / `failure::Fail` infrastructure.

it doesn't seem like it'd be useful to enforce those bounds through our API, but, it might be helpful to add (feature-gated) implementations for either `std:error::Error` or `failure::Fail` to our ErrorKinds for if/when they're used directly?

eldruin:
Hmm, `ErrorKind`s would not be part of the error chain themselves as I do not understand them as a `std::error::Error::source()`.
Rather, the original HAL-defined error types are the sources and those would form the error chain.
What we would add here is requiring that those errors could be translated into a predefined subset: `ErrorKind`s, similar to [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html#method.kind).

If those HAL-defined error types implement `std::error::Error`, they get `failure::Error` for free IIUC.

I have implemented `core::fmt::Display` for the `ErrorKind`s, though. Which I guess could be useful.

yodaldevoid:
> Hmm, `ErrorKind`s would not be part of the error chain themselves as I do not understand them as a `std::error::Error::source()`.
> Rather, the original HAL-defined error types are the sources and those would form the error chain.
> What we would add here is requiring that those errors could be translated into a predefined subset: `ErrorKind`s, similar to [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html#method.kind).

I believe there was talk about allowing `ErrorKind` to directly be as the HAL error type where it fits for a specific HAL. It's fine to explicitly disallow or discourage this use, but if we would still like to support this then I would ask that `std::error::Error` be conditionally implemented for `ErrorKind`.

eldruin:
HALs can indeed use `ErrorKind` as their error type if that is enough for them. I have implemented the `embedded-hal::<protocol>::Error` traits for those in this PR.

A different topic would be implementing `std::error::Error` for `ErrorKind` as well. That would be an independent addition but only really useful for `linux-embedded-hal` or so, which I highly doubt would adopt `ErrorKind` as its protocol error type.
The main point against it being that `std::error::Error` requires memory allocation so it is uninteresting for `no_std` HALs.
Anyway if there would really be interest, we could add this in a future release via an opt-in feature, but I would not hold the merge of this since that would be a hypothetical non-breaking addition.

yodaldevoid:
This PR is just waiting for a final review and a quick update to update the change-log, correct?

src/errors.rs
```diff
@@ -0,0 +1,180 @@
+//! Error definitions
+
+pub mod i2c {
+    /// I2C error
+    pub trait Error: core::fmt::Debug {
+        /// Convert error to a generic I2C error kind
+        ///
+        /// By using this method I2C errors freely defined by HAL implementations
```
dbrgn, line 8: Nitpick: Comma missing.

```suggestion
        /// By using this method, I2C errors freely defined by HAL implementations
```

(Same thing applies to other `Error` definitions below.)

src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+        Bus,
+        /// The arbitration was lost, e.g. electrical problems with the clock signal
+        ArbitrationLoss,
+        /// A bus operation was not acknowledged, e.g. due to the addressed device not being available on
+        /// the bus or the device not being ready to process requests at the moment
+        NoAcknowledge,
+        /// The peripheral receive buffer was overrun
+        Overrun,
+        /// The peripheral send buffer ran out of data
+        Underrun,
```
dbrgn, line 32: I think no current HAL uses this error kind so far: https://github.com/rust-embedded/embedded-hal/issues/229#issuecomment-647428967 But I assume there's still value if the error is standardized, in case it could be used in the future.

(However, in that case, would SPI and serial need an `Underrun` error as well?)

src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+        /// A different error occurred. The original error may contain more information.
+        Other,
+    }
+
+    impl Error for ErrorKind {
+        fn kind(&self) -> ErrorKind {
+            *self
+        }
+    }
```
dbrgn, line 41: This is elegant 👍🏻 Quick-and-dirty HAL implementations can use the `ErrorKind` directly as error, while a more feature-rich HAL would implement a custom error type with appropriate conversions.

src/errors.rs
```diff
@@ -0,0 +1,180 @@
...
+                    f,
+                    "The arbitration was lost, e.g. electrical problems with the clock signal"
+                ),
+                Self::NoAcknowledge => write!(f, "A bus operation was not acknowledged"),
+                Self::Overrun => write!(f, "The peripheral receive buffer was overrun"),
+                Self::Underrun => write!(f, "The peripheral send buffer ran out of data"),
+                Self::Other => write!(
+                    f,
+                    "A different error occurred. The original error may contain more information"
+                ),
```
dbrgn, line 57: I wonder if those descriptions are a bit _too_ detailed, and should instead go into a `description` method (returning static strings), or something similar.

dbrgn (commented):
Thanks for the proposal, I like it. It's a lightweight approach that provides flexibility for HALs to define their custom error types, while at the same time offering a possibility to map custom errors to a standardized set of error types.

src/errors.rs
eldruin: Good point. As I said I mostly got the variants from `embedded-error`. I am fine with removing the `Underrun` variant since it seems unclear it can be detected at all.
The `ErrorKind`s are marked as `#[non_exahustive]` so adding it in the future is not a breaking change.

src/errors.rs
eldruin: The `description` method was [deprecated](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.description) in the `std` in favor of the `Display` impl so it would not be good re-adding it here.
An additional drawback would be that with `description` methods, all those strings end up added to the binaries, which causes unnecessary bloat. With the `Display` impl, if you do not use it, it will not be added.
About wording changes, please feel free to suggest any.

dbrgn: Ah, thanks, that makes sense. Regarding wording, I'd simply make the descriptions more compact, so instead of "The arbitration was lost, e.g. electrical problems with the clock signal" you could use "Arbitration lost". The more detailed variant could then live in the docs. But maybe other people see value in more detailed error descriptions? I don't know.

eldruin: There is still a difference between `Debug` and `Display`. The `Display` impl is for user-facing output, so I figured a more elaborate formulation would be due, but yeah, "everything is bloat" I guess. I will remove that "e.g." in the arbitration lost message but what do you think about the others?

Dirbaio:
+1 to having official "error kinds" that generic drivers can react to. This is sorely needed!

Maybe this PR should enforce that the `Error` associated types impl the `Error` trait? The original proposal does so.

yodaldevoid:
> Maybe this PR should enforce that the `Error` associated types impl the `Error` trait? The original proposal does so.

The `std::error::Error` trait is only in `std`, not `core`.

eldruin:
@Dirbaio @yodaldevoid I missed re-adding the `Error` trait requirement on the associated types today by mistake as I re-did this PR completely on top of the current master branch. Sorry for the confusion.

src/i2c.rs
```diff
@@ -101,6 +101,58 @@
...
+/// I2C error kind
+///
+/// This represents a common set of I2C operation errors. HAL implementations are
+/// free to define more specific or additional error types. However, by providing
+/// a mapping to these common I2C errors, generic code can still react to them.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[non_exhaustive]
+pub enum ErrorKind {
+    /// An unspecific bus error occurred
+    Bus,
```
Dirbaio, line 123: IMO this error is the same as `Other` in practice. Any ErrorKind can be an "i2c bus error", because they're all about i2c. I'd remove it.

src/i2c.rs
```diff
@@ -101,6 +101,58 @@
...
+    /// The arbitration was lost, e.g. electrical problems with the clock signal
+    ArbitrationLoss,
+    /// A bus operation was not acknowledged, e.g. due to the addressed device not being available on
+    /// the bus or the device not being ready to process requests at the moment
+    NoAcknowledge,
```
Dirbaio, line 128: It would be useful to distinguish between "NAK on address" and "NAK on data". First may indicate the slave is not there (or busy), the second indicates the slave is deinitely there, but isn't accepting (more) data. 

nRFs and STM32s can distinguish them, probably others too.

src/spi/mod.rs
```diff
@@ -53,3 +53,63 @@ pub const MODE_3: Mode = Mode {
...
+#[non_exhaustive]
+pub enum ErrorKind {
+    /// An unspecific bus error occurred
+    Bus,
+    /// The peripheral receive buffer was overrun
+    Overrun,
+    /// Multiple devices on the SPI bus are trying across each other, e.g. in a multi-master setup
+    ModeFault,
+    /// CRC does not match the received data
+    Crc,
```
Dirbaio, line 82: If nothing on the traits deals with CRCs, do we want a CRC error?

src/spi/mod.rs
```diff
@@ -53,3 +53,63 @@ pub const MODE_3: Mode = Mode {
...
+/// SPI error kind
+///
+/// This represents a common set of SPI operation errors. HAL implementations are
+/// free to define more specific or additional error types. However, by providing
+/// a mapping to these common SPI errors, generic code can still react to them.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[non_exhaustive]
+pub enum ErrorKind {
+    /// An unspecific bus error occurred
+    Bus,
```
Dirbaio, line 76: I'd remove this, see comment on i2c

Dirbaio (commented):
I've commented some bikeshedding on the error enums :)

src/i2c.rs
TeXitoi: you can have other errors, as permission errots on linux for example. Now, it's better to pit the minimal amount of error as we can't remove errors, bit we can add errors layer if needed.

Dirbaio: I don't think distinguishing "unknown error somewhat related to the bus" and "unknown error probably not related to the bus" is valuable. In practice, platform-independent drivers can do nothing about them, they'll just pass them up.

(Also, I'd say platform-specific errors like "permission denied" or "usb-to-i2c converter unplugged" shouldn't be listed in ErrorKind, they should always be `Other`, because again, they're not useful. Drivers will just pass them up)

TeXitoi: I agree

therealprof: > IMO this error is the same as `Other` in practice. Any ErrorKind can be an "i2c bus error", because they're all about i2c. I'd remove it.

No, this is not the same. This is taken straight from the STM32 implementation and defined in an RM as:
```
A bus error is detected when a START or a STOP condition is detected and is not located after a multiple of 9 SCL clock pulses. A START or a STOP condition is detected when a SDA edge occurs while SCL is high.
The bus error flag is set only if the I2C is involved in the transfer as master or addressed slave (i.e not during the address phase in slave mode).
```

IMNSHO the the ErrorKinds should be a superset of all error native peripherals can distinguish.

src/i2c.rs
therealprof: I'm only aware of a "NACK on data" for SMBus and haven't seen that supported for I2C yet. Which STM32 supports that if I may ask?

therealprof:
Still not too much of a fan of the `Other` kind and would have preferred an `ImplError` instead as I've done in `embedded-error`. Anyway, this PR is fine with me.

src/i2c.rs
yodaldevoid: I agree with @therealprof, these are not the same as Bus errors are a defined type of error for I2C.

src/i2c.rs
yodaldevoid: I have seen a few micro controllers support this (though those that come to mind don't have publicly available datasheets). I'm not sure how useful this information will be in practice as I cannot think of any time anything but the I2C peripheral driver itself would care about this distinction.

src/i2c.rs
Dirbaio: Ah, I see. "bus error" is a thing with a particular meaning. I'm fine with keeping it then, but the docs for it should be more specific. "Unspecific bus error" sounds like it's a "catch-all" for all i2c errors that are not explicitly categorized...

src/spi/mod.rs
eldruin: At least STM32s support hardware CRC on SPI. See for example: [AN2576 chapter 3](https://application-notes.digchip.com/005/5-10375.pdf#G1065017)

Dirbaio: Yeah, but the traits have no way to enable/configure CRC checking in a HAL-independent way. If some driver wants CRC checking, it'll presumably use the HAL-specific stm32xxx-hal APIs to enable it, so it can check for CRC errors using the concrete `Error` type from the HAL instead of through ErrorKind.

IMO the HAL-independent ErrorKind should only have errors relating to the functionalities in HAL-independent traits.

dbrgn:
Should this PR (or #229) be added to the "1.0" milestone, now that it's listed in #177?

src/i2c.rs
eldruin: I have updated the documentation of the bus error with the information provided here.

src/spi/mod.rs
eldruin: Good point, the I2C condition does not apply here. Done.

src/spi/mod.rs
eldruin: If we follow that logic rigorously we can basically remove everything but `Overrun` and `Other` (same thing for serial).
I think there is value in providing variants for most common error conditions, though.

yodaldevoid:
Is this still awaiting further reviews?

src/spi/mod.rs
```diff
@@ -53,3 +53,60 @@ pub const MODE_3: Mode = Mode {
...
+    }
+}
+
+impl core::fmt::Display for ErrorKind {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::Overrun => write!(f, "The peripheral receive buffer was overrun"),
+            Self::ModeFault => write!(
+                f,
+                "Multiple devices on the SPI bus are trying across each other"
```
threema-danilo, line 99: This wording could probably be improved, "trying across each other" sounds a bit wonky.

src/errors.rs
threema-danilo: I think the others are fine.

dbrgn: I think the others are fine.

src/spi/mod.rs
dbrgn, line 99: This wording could probably be improved, "trying across each other" sounds a bit wonky.

eldruin: Yes, sounds weird. Any suggestions?

yodaldevoid: `"Multiple devices on the SPI bus are trying to drive the peripheral select pin"`?

src/i2c.rs
dbrgn: If a microcontroller doesn't support distinguishing the two, mapping a NACK to one of those two variants would be difficult.

src/spi/mod.rs
eldruin: Done, thanks for the suggestion!

src/i2c.rs
yodaldevoid: That is a good point. The two solutions I could see here would either to just keep one `NoAcknowledge` variant or to have `NoAcknowledgeAddress` and `NoAcknowledgeData` variants and specify that when a driver can't differentiate between the two to default to a specific one.

I think I would prefer the second option, but I would not be against the first.

src/spi/mod.rs
yodaldevoid: I am with @Dirbaio on this one. I say we remove everything but `Overrun` and `Other`, and if a CRC trait or some such is added to the HAL then these sorts of variants can be added at that time.

src/i2c.rs
Dirbaio: I propose doing this:

- NoAcknowledge: NAK after addr or data byte, the hardware is incapable of distinguishing them.
- NoAcknowledgeAddress: NAK after address byte
- NoAcknowledgeData: NAK after data byte

yodaldevoid: That seems fine by me. There would just need to be documentation detailing this, of course.

src/spi/mod.rs
eldruin: I have now removed `ErrorKind::Crc` as discussed.

src/i2c.rs
eldruin: I replaced `NoAcknowledge` with `NoAcknowledgeAddress` and `NoAcknowledgeData`. Since the `enum` is non-exhaustive, we can add `NoAcknowledge` later if an implemementation cannot distinguish between the two.

TeXitoi: Can you imagine a generic driver handling differently NAK address and NAK data? Because if that's not the case, better to do the contrary: only keep NoAcknowledge for now, and we can add the address and data variant later, no?

eldruin: Yes, depending on the device, NAK address may mean the device is not there at all so bubble up the error to the caller. OTOH, NAK data may mean "I'm busy right now, but you can try again later" so the driver can retry at a later time or return something like `nb::WouldBlock` or so.

TeXitoi: OK, make sense.

yodaldevoid: I think only having `NoAcknowledgeAddress` and `NoAcknowledgeData` without an defined way to handle drivers that cannot differentiate between the two is problematic. If we don't want to add a third `NoAcknowledge`, I would ask that we add documentation saying which one to use by default. Adding `NoAcknowledge` down the line with the current design might cause users to miss the new `NoAcknowledge`, even if adding that variant was considered only a minor change.

---

This may be unacceptable, but perhaps we should instead have `NoAcknowledge` be a tuple enum that specifies if it was on the address or data. e.g.

```rust
#[non_exhaustive]
pub enum ErrorKind {
        /// A bus operation was not acknowledged, e.g. due to the addressed device not being available on
    /// the bus or the device not being ready to process requests at the moment
    NoAcknowledge(NoAcknowledgeSource),
}

enum NoAcknowledgeSource {
    /// The device did not acknowledge its address. The device may be missing.
    Address,
    /// The device did not acknowledge the data. It may not be ready to process requests at the moment.
    Data,
    /// The driver does not know if the no acknowledge was from the address or the data.
    Unknown,
}
```

This would allow users to catch all NACKs if they wanted to or get more information if they wanted that.

therealprof (approved):
LGTM

src/i2c.rs
eldruin: Since we have reached an agreement on this long-standing issue, I would like to merge it as it is.
If you are interested, would you open a separate PR or issue with this proposal?

eldruin:
🎉 
bors r=therealprof

src/i2c.rs
yodaldevoid: Yeah, I will submit an follow-on PR.

PR #297, opened 2021-07-18
Reexport definitions from SPI in blocking SPI module

eldruin:
It makes no sense to offer these only on the `nb` module.
I am also thinking about turning the modules inside out as proposed in the chat room.


eldruin:
Closing now that #298 has been merged.

PR #298, opened 2021-07-20
Turn traits and execution model modules inside out

eldruin:
A proposal for discussion.
## Advantadges
- Place for execution-model-independent type definitions
    - For example useful for SPI modes 
- Fits better with users who might think of peripheral first, then execution model. - @Sh3Rm4n 
- Helps discoverability, since the modules are not "hidden" behind their execution model. - @Rahix
## Disadvantadges
- Breaks every last `embedded-hal`-dependent code
- Users may want to import several `embedded_hal::xxx::blocking` modules, for example, although there is no much reason to do so.
- There might be confusion between `embedded_hal::xxx::nb` and `embedded_hal::nb`.
## Open questions
- Reexport execution-model-independent types in `blocking`/`nb` submodules?
## Alternatives
Keep the structure as-is


Sh3Rm4n:
Even though this change might be quite disruptive, I really like this approach. 

I imagine, normally when a user uses `e-h` they will first choose the interface they want to use (`spi`, `serial`, ...) and **then** they'll choose the implementation, whether it's blocking, `nb` based or `async`.

Sometimes, using mutliple implementations simulatiously could also happen. This change would make the `use` statements cleaner. Instead of 

```rust
use emedded_hal::{
    nb::serial::Write,
    blocking::serial::Write,
};
```

it would be like this:

```rust
use emedded_hal::serial{
    nb::Write,
    blocking::Write,
};
```

therealprof:
@ryankurte Are you aware of this proposal? Any concerns? Otherwise we'd go ahead merge it.

ryankurte:
it is an interesting exercise to consider alternative layouts and this change is neat that it creates an obvious space for common definitions, however, apart from that i'm not convinced there's a strong argument for this, or that the refactoring is a net improvement over the existing layout compared to say, adding a `common` module. 

so i'm opposed to this as it seems to me to be substantial churn for predominantly a stylistic preference. while this is absolutely the time to make breaking changes, it does impact a bunch of downstream users, and i don't see that it makes anyone's e-h experience measurably or demonstrably _better_ as justification for this.


> Users may want to import several embedded_hal::xxx::blocking modules, for example, although there is no much reason to do so.

> I imagine, normally when a user uses e-h they will first choose the interface they want to use (spi, serial, ...) and then they'll choose the implementation, whether it's blocking, nb based or async.

my experience is probably the opposite of the comments on this PR. ime you tend to import a whole pile of peripheral definitions when wiring things up, what you use is primarily defined by the runtime available to you, and it's unlikely (*, except for `serial` atm) one would choose to mix-and-match those from blocking/nb(/async).

there's also talk of feature-gating the upcoming `async` feature, which is totally doable either way but, much simpler with the current layout.

GrantM11235:
Even if this PR is not accepted, there will still be a lot of churn due to #278.

For the record, I do think that this refactor is a net improvement.

therealprof (approved):
@ryankurte Thanks for unblocking this.

I am a fan of the revised model for the reasons mentioned. So let's get the party started. 🎉

bors r+

PR #306, opened 2021-08-12
Add Debug to all spi mode types

Sh3Rm4n:
I've noticed that these were missing, and I see no reason why to not implement `Debug`.


ryankurte:
lgtm, thanks!

bors r+

ryankurte (approved):
whoops...

bors r+

PR #309, opened 2021-08-31
Fix missing `capture` module

GrantM11235:
cc @eldruin 

This must have gotten lost during #298


eldruin (approved):
Indeed. Thanks!
bors r+

PR #310, opened 2021-09-07
Add blanket impls for references

GrantM11235:
I would prefer to use a proc-macro like [auto-impl](https://crates.io/crates/auto_impl), but that doesn't support no_std yet. (see auto-impl-rs/auto_impl#73)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `IoPin` trait for pins that can change between being inputs or outputs
   dynamically.
 - Added `Debug` to all spi mode types.
+- Added blanket impls for references.
 
 ### Changed
 - Swap PWM channel arguments to references
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `IoPin` trait for pins that can change between being inputs or outputs
   dynamically.
 - Added `Debug` to all spi mode types.
-- Added blanket impls for references.
+- Add impls of all traits for references (`&T` or `&mut T` depending on the trait) when `T` implements the trait.
 
 ### Changed
 - Swap PWM channel arguments to references
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -112,23 +112,23 @@ pub mod blocking {
         type Duty = T::Duty;
 
         fn disable(&mut self, channel: &Self::Channel) -> Result<(), Self::Error> {
-            (*self).disable(channel)
+            T::disable(self, channel)
         }
 
         fn enable(&mut self, channel: &Self::Channel) -> Result<(), Self::Error> {
-            (*self).enable(channel)
+            T::enable(self, channel)
         }
 
         fn get_period(&self) -> Result<Self::Time, Self::Error> {
-            (**self).get_period()
+            T::get_period(self)
         }
 
         fn get_duty(&self, channel: &Self::Channel) -> Result<Self::Duty, Self::Error> {
-            (**self).get_duty(channel)
+            T::get_duty(self, channel)
         }
 
         fn get_max_duty(&self) -> Result<Self::Duty, Self::Error> {
-            (**self).get_max_duty()
+            T::get_max_duty(self)
         }
 
         fn set_duty(
@@ -136,14 +136,14 @@ pub mod blocking {
             channel: &Self::Channel,
             duty: Self::Duty,
         ) -> Result<(), Self::Error> {
-            (*self).set_duty(channel, duty)
+            T::set_duty(self, channel, duty)
         }
 
         fn set_period<P>(&mut self, period: P) -> Result<(), Self::Error>
         where
             P: Into<Self::Time>,
         {
-            (*self).set_period(period)
+            T::set_period(self, period)
         }
     }
 
@@ -185,23 +185,23 @@ pub mod blocking {
         type Duty = T::Duty;
 
         fn disable(&mut self) -> Result<(), Self::Error> {
-            (*self).disable()
+            T::disable(self)
         }
 
         fn enable(&mut self) -> Result<(), Self::Error> {
-            (*self).enable()
+            T::enable(self)
         }
 
         fn get_duty(&self) -> Result<Self::Duty, Self::Error> {
-            (**self).get_duty()
+            T::get_duty(self)
         }
 
         fn get_max_duty(&self) -> Result<Self::Duty, Self::Error> {
-            (**self).get_max_duty()
+            T::get_max_duty(self)
         }
 
         fn set_duty(&mut self, duty: Self::Duty) -> Result<(), Self::Error> {
-            (*self).set_duty(duty)
+            T::set_duty(self, duty)
         }
     }
 }
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -102,6 +102,51 @@ pub mod blocking {
             P: Into<Self::Time>;
     }
 
+    impl<T: Pwm> Pwm for &mut T {
+        type Error = T::Error;
+
+        type Channel = T::Channel;
+
+        type Time = T::Time;
+
+        type Duty = T::Duty;
+
+        fn disable(&mut self, channel: &Self::Channel) -> Result<(), Self::Error> {
+            (*self).disable(channel)
+        }
+
+        fn enable(&mut self, channel: &Self::Channel) -> Result<(), Self::Error> {
+            (*self).enable(channel)
+        }
+
+        fn get_period(&self) -> Result<Self::Time, Self::Error> {
+            (**self).get_period()
+        }
+
+        fn get_duty(&self, channel: &Self::Channel) -> Result<Self::Duty, Self::Error> {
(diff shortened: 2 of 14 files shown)


therealprof:
> I would prefer to use a proc-macro like [auto-impl](https://crates.io/crates/auto_impl), but that doesn't support no_std yet. (see [auto-impl-rs/auto_impl#73](https://github.com/auto-impl-rs/auto_impl/pull/73))

As mentioned in the chat I think the overhead of a `proc-macro` is totally inacceptable for this purpose. The blanket impls seem fine to me.

src/pwm.rs
```diff
@@ -102,6 +102,51 @@ pub mod blocking {
...
+        fn disable(&mut self, channel: &Self::Channel) -> Result<(), Self::Error> {
+            (*self).disable(channel)
+        }
+
+        fn enable(&mut self, channel: &Self::Channel) -> Result<(), Self::Error> {
+            (*self).enable(channel)
+        }
+
+        fn get_period(&self) -> Result<Self::Time, Self::Error> {
+            (**self).get_period()
```
Dirbaio, line 123: Why do some require `**` while most require `*`? Is it to avoid recursion?

Maybe the fully qualified syntax is less footgunny? like this: `T::get_period(self)`

almindor: I think this is a typo?

GrantM11235: Yes, it is to avoid recursion. Methods that take `&self` in impls for `&mut T` need `**`, all other cases seem to coerce properly with just a single `*`.

I don't think it is much of a footgun because if it is wrong it will simply fail to compile.

almindor (dismissed):

ryankurte:
hey, seems useful! i _think_ this approach is still better because the implications are much more straightforward, but, wondering whether y'all considered implementing this via `AsRef`/`AsMut` bounds over a generic type?

eldruin (commented):
Seems useful to me, thanks!
I think fully-qualified syntax instead of `**` would be better and less surprising, as previously suggested.
I agree a version of this using `AsRef`/`AsMut` would be interesting to see.

Dirbaio:
How would the AsRef/AsMut work? Having drivers do something like `struct Driver<T> where T: AsMut<impl Write>`? But then they no longer work with owned peripherals.

ryankurte:
okay i had a wee look and it seems `AsRef<T> + AsMut<T>` isn't viable, but, `Deref<Target=T> + DerefMut` achieves the same goal and works for pretty much all references or smart-pointers, and i'm not aware of any drawbacks to this?

something like:
```rust
/// Fake trait for demonstration
pub trait Fake {
    fn idk(&mut self);
}

/// Blanket implementation to cover deref-able types
impl <T, C> Fake for C 
where
    C: Deref<Target=T> + DerefMut,
    <C as Deref>::Target: Fake
{
     fn idk(&mut self) {
        <C as Deref>::Target::idk(self.deref_mut())
    } 
}
```

you can see a full example [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=61fe738cf824fc35f0fb17558699c3e0) for more context.

GrantM11235:
Is there any reason that a generic function would need to be able to take a smart pointer directly? Can't the user simply deref the smart pointer before giving it to the generic function?

ryankurte:
> Is there any reason that a generic function would need to be able to take a smart pointer directly? Can't the user simply deref the smart pointer before giving it to the generic function?

it is possible i have misunderstood the motivation behind this PR? my understanding is that the goal of this is for `&mut _` to fulfill  common driver bounds (`T: Something`) so that drivers can be constructed over (ie. store) references as well as objects, as distinct from passing into a generic function which as you point out you can manually deref. you also shouldn't need any additional trait to call methods on a reference because [deref coercion](https://doc.rust-lang.org/book/ch15-02-deref.html) should magically take care of that for you,.

implementing this over `Deref + DerefMut` seems like the more generic solution for `&mut _` while also allowing `Box<_>` which i think is the most useful example (as well as other deref types, closer to `auto_impl`, without the proc_macro cost).

i've update the [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b62b996661a440c8266a1cc9f021d678) to hopefully demonstrate this better, something like:
```rust
// Driver requires an object implementing the `Test` trait
struct Driver<T: Test> { _t: T }

// Fake implements the Test trait, with a method `.idk()`
let mut f = Fake;
f.idk();

// Object always works
Driver::new(Fake);

// Reference works with `&mut _` or `Deref + DerefMut` blanket impl
Driver::new(&mut Fake);

// The following only work with `Deref + DerefMut` blanket impl

Driver::new(Box::new(Fake));

let shared_fake = Arc::new(RefCell::new(Fake));
Driver::new(shared_fake.borrow_mut());

let sync_fake = Arc::new(Mutex::new(Fake));
Driver::new(sync_fake.lock().unwrap());

```

this does provide blanket support for `RefCell`s and things which will allow people to drop borrows and locked mutexes in, but i don't think this has any impact because it doesn't include the _locking_ which we still need to manage (cc. @eldruin @rahix)

(unrelated to this PR, i do often wish we had a way of genericising over `dyn Sometime` too, but i think it is not so straightforward as this)

GrantM11235:
> my understanding is that the goal of this is for &mut _ to fulfill common driver bounds (T: Something) so that drivers can be constructed over (ie. store) references as well as objects

That is correct.

What I meant to ask is: is it necessary to be able to hold the smart pointer in the driver

```rust
    let _: Driver<Box<Fake>> = Driver::new(Box::new(Fake));

    let _: Driver<RefMut<Fake>> = Driver::new(shared_fake.borrow_mut());

    let _: Driver<MutexGuard<Fake>> = Driver::new(sync_fake.lock().unwrap());
```

instead of

```rust
    let _: Driver<&mut Fake> = Driver::new(&mut Box::new(Fake));

    let _: Driver<&mut Fake> = Driver::new(&mut shared_fake.borrow_mut());

    let _: Driver<&mut Fake> = Driver::new(&mut sync_fake.lock().unwrap());
```

As far as I understand, it is important for the driver to be able to hold an owned peripheral (ie. don't only accept a &mut)
in order to avoid self-reference problems in some cases. I guess the driver should to be able to hold a smart pointer for similar reasons?

Dirbaio:
:+1: to the Deref+DerefMut blanket impl. It's nice that it magically works with all smart pointers!

One minor downside I can think of is: the HALs will no longer be able to impl Deref/DerefMut for the peripheral structs. Arguably they shouldn't since peripheral drivers aren't smart pointers. Sometimes Deref/DerefMut is abused to do fake "inheritance", but that's considered a bad practice so forbidding it is OK imo.

Other than that I really can't think of any downsides. Are we sure we're not missing something though? blanket impls for `&mut` or `Box` or similar are really common throughout the Rust ecosystem but `Deref/DerefMut` ones aren't so common. I wonder why.

btw it can be simplified a bit, no need for 2 generic params:

```rust
impl <C> Test for C 
where
    C: Deref + DerefMut,
    C::Target: Test,
{
     fn idk(&mut self) {
        C::Target::idk(self.deref_mut())
    }   
}
```

almindor:
[Discussion logs from WG meeting](https://libera.irclog.whitequark.org/rust-embedded/2021-09-14#30838560)

Dirbaio:
tl;dr: 

- The Deref blanket impl makes compile errors suck
  - without blanket impl: error[E0277]: the trait bound u32: Traity is not satisfied
  - with blanket impl: error[E0277]: the trait bound u32: DerefMut is not satisfied
  - playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e7c9e223ece82ca3a7d383e43be027c8
- Doing a blanket impl for `Box<T>` is not nice because then `embedded-hal` would have to require `alloc`, or have an `alloc` cargo feature to do so optionally, which is not nice.

So it seems the best way to go is just do the blanket impls for `&mut T`, which don't make the errors suck.

ryankurte (approved):
> So it seems the best way to go is just do the blanket impls for &mut T, which don't make the errors suck.

thanks for investigating the options! given the downside i'm fine with either approach, and with the change from `**` to `T::`, lgtm. @eldruin any other questions / concerns?

(also the uhh, less than useful, Deref error is probably worth a bug report on https://github.com/rust-lang/rust)

> blanket impls for &mut or Box or similar are really common throughout the Rust ecosystem but Deref/DerefMut ones aren't so common

i haven't seen a lot of either tbh, but, picked it up from whitequark's [abstracting over mutability in rust](https://lab.whitequark.org/notes/2016-12-13/abstracting-over-mutability-in-rust/)

almindor (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added `IoPin` trait for pins that can change between being inputs or outputs
   dynamically.
 - Added `Debug` to all spi mode types.
+- Added blanket impls for references.
```
eldruin, line 14: This could be a bit more descriptive.

eldruin (commented):
Other than the changelog entry, looks good to me, thanks!

ryankurte (approved):
lgtm, thanks for all your effort!

therealprof (approved):
Great. Let's get this landed then.

bors r+

PR #314, opened 2021-09-22
Controller Area Network (CAN) Take 4

timokroeger:
Updated to the latest HAL changes:
* Removed `try_` prefix
* Moved non-blocking implementation to `nb` module
* Removed default `blocking` implementaions

## Usage Example
[stm32-fwupdate](https://github.com/timokroeger/pcan-basic-rs/blob/eh-take-4/pcan-basic/examples/stm32-fwupdate.rs)

## Implementations
Updated for this PR:
* [pcan-basic](https://github.com/timokroeger/pcan-basic-rs/blob/eh-take-4/pcan-basic/src/lib.rs) on top of an existing software API
* [bxcan](https://github.com/timokroeger/bxcan/blob/eh-take-4/src/lib.rs#L460)

Based on the very similar predecessor traits `embedded-can` v0.3 ([diff v0.3 -> this PR](https://github.com/timokroeger/embedded-can/compare/eh-take-4))
* [candev](https://github.com/reneherrero/candev) implementing the traits for linux SocketCAN
* [socketcan-isotc](https://github.com/marcelbuesing/socketcan-isotp)

## Previous Discussion
* #212 
* #77
* #21
* #53


adamgreig:
This looks good to me! I have a couple of thoughts but apologies if they've already been discussed in the previous threads...

* Does `new_unchecked` actually need to be unsafe, i.e. might any memory safety rely on it later?
* For CAN, might it make more sense to call it `transmit` and `receive` in both the blocking and nb traits, rather than `read` and `write` in blocking?
* I saw you mentioned this on the [old PR thread](https://github.com/rust-embedded/embedded-hal/pull/212#issuecomment-925167072) but it does seem like it would be nice to add some error types to #296 for things like invalid ID or data-too-long.

src/can/mod.rs
```diff
@@ -45,3 +45,71 @@ pub trait Frame: Sized {
...
+    /// Calculated CRC sequence does not equal the received one.
+    Crc,
+
+    /// A form error shall be detected when a fixed-form bit field contains
+    /// one or more illegal bits.
+    Form,
+
+    /// An ACK  error shall be detected by a transmitter whenever it does not
+    /// monitor a dominant bit during the ACK slot.
+    Ack,
```
eldruin, line 89: For no-acknowledge errors we selected `NoAcknowledge` over at #296 . It would be good to have the same name in both sides.

adamgreig: The CAN 2.0 standard says there are 5 error types, which it calls:

* `BIT ERROR`
* `STUFF ERROR`
* `CRC ERROR`
* `FORM ERROR`
* `ACKNOWLEDGEMENT ERROR`

So matching those names seems nice, while in I²C the condition where there's no acknowledgement is explicitly called `NAK`/no acknowledge, so it makes sense to use NoAcknowledge for I²C. I guess there's a question of whether it should be `Ack` or `Acknowledgement` here, but given how long the latter is, I'd be fine with `Ack`...

PR #316, opened 2021-10-05
I2C NACK error nesting

yodaldevoid:
This idea was first proposed [here](https://github.com/rust-embedded/embedded-hal/pull/296#discussion_r721386337), but broken out to keep things moving.

This merges `ErrorKind::NoAcknowledgeData` and `ErrorKind::NoAcknowledgeAddress` into one variant with a source field to indicate what the NACK was in response to. Some drivers may not be able to differentiate between NACKs on address or data bytes so they should use `NoAcknowledgeSource::Unknown`.

Feel free to bikeshed names and documentation. I used the most obvious names to me, but naming is one of the hardest problems in programming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   implement these `Error` traits, which implies providing a conversion to a common
   set of error kinds. Generic drivers using these interfaces can then convert the errors
   to this common set to act upon them.
+- The `NoAcknowledgeAddress` and `NoAcknowledgeData` variants of the I2C `Error`
+  trait have been merged into `NoAcknowledge` with a `NoAcknowledgeSource` field
+  to differentiate between the two events. An additional `Unknown` variant is
+  provided in `NoAcknowledgeSource` for devices that can't differentiate between
+  the two events.
 
 ### Removed
 - Removed `DelayMs` in favor of `DelayUs` with `u32` as type for clarity.
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -119,7 +119,8 @@ pub trait Error: core::fmt::Debug {
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 #[non_exhaustive]
 pub enum ErrorKind {
-    /// Bus error occurred. e.g. A START or a STOP condition is detected and is not located after a multiple of 9 SCL clock pulses.
+    /// Bus error occurred. e.g. A START or a STOP condition is detected and is not
+    /// located after a multiple of 9 SCL clock pulses.
     Bus,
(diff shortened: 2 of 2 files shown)


adamgreig:
Thanks for the PR!

Do you have any examples of an implementation that can't distinguish address vs data NAKs? I understand they could exist but I couldn't find any, so I wonder if they do exist in practice.

yodaldevoid:
> Do you have any examples of an implementation that can't distinguish address vs data NAKs? I understand they could exist but I couldn't find any, so I wonder if they do exist in practice.

Based off of [this comment](https://github.com/rust-embedded/embedded-hal/issues/229#issuecomment-647428967) most (if not all) HALs currently don't differentiate between NACKs on data vs. address.

As a quick check, I looked at the STM32L0x3 reference manual and based on the possible interrupts it does not differentiate between NACKs on address and data. I suppose you could figure out where the NACK occurred based on if you had previously seen the address match interrupt. This sort of decoding might be possible in other HALs, and if I have time later I will dig through what I can find.

Beyond that, where I work we have an internal device that does not differentiate between NACKs on address vs. data. Sadly, because it is an internal device I can't provide any documentation. I would also not disagree with anyone who says that that device is badly designed to give so little feedback.

adamgreig:
Thanks for the feedback! It's especially useful to hear about custom devices as that's of course exactly the sort of thing we can't find by searching around.

> As a quick check, I looked at the STM32L0x3 reference manual and based on the possible interrupts it does not differentiate between NACKs on address and data. I suppose you could figure out where the NACK occurred based on if you had previously seen the address match interrupt. This sort of decoding might be possible in other HALs, and if I have time later I will dig through what I can find.

This is more or less what I'd expect - the device says it saw a "NAK", but to follow the I2C state machine you would know if it was in the address or data phase. For example in stm32f4xx-hal, a NACK [here](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/i2c.rs#L458) always means an address NACK (because we're waiting for the ADDR bit to be set, indicating the address has been sent and ACK'd), while a NACK in `send_byte` always means a data NACK (because the address has already been ACKd in this transaction).

Does the implementation for your custom device not permit anything similar? Is the interface a combined address+data and you just get back a single transaction status, or do you send an address, get a result, send some data, get a result, etc?

yodaldevoid:
> Does the implementation for your custom device not permit anything similar? Is the interface a combined address+data and you just get back a single transaction status, or do you send an address, get a result, send some data, get a result, etc?

The device we have will indicate if a NACK has occurred in the transaction, but not where. This is true both for "raw" I2C reads and writes in addition to the defacto register read and write scheme that many devices implement. If we are reading data and we have received some of our requested data we can detect a data NACK in that case, but that is it. This is an older API that we are still trying to support so we can't easily add in differentiated NACKs.

therealprof (approved):
LGTM

PR #320, opened 2021-11-03
add default SPI word types

burrbull:
Thought about this reading #295

cc @eldruin  @Sh3Rm4n  @ryankurte 

Should we restrict types be some trait?
Can also split `U16` to `U16LE` and `U16BE`

### DIFF
--- a/src/spi/mod.rs
+++ b/src/spi/mod.rs
@@ -3,6 +3,15 @@
 pub mod blocking;
 pub mod nb;
 
+/// 8-bit SPI Word (default)
+pub type U8 = u8;
+/// 9-bit SPI Word
+pub struct U9;
+/// 16-bit SPI Word
+pub struct U16;
+/// 18-bit SPI Word
+pub struct U18;
+
 /// Clock polarity
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum Polarity {
--- a/src/spi/mod.rs
+++ b/src/spi/mod.rs
@@ -1,7 +1,15 @@
 //! SPI traits
 
+use crate::private::Sealed;
+
 pub mod blocking;
 pub mod nb;
+
+/// Marker trait for SPI Word
+pub trait SpiWord: Sealed {
+    /// Specified data size type
+    type Data: Sized;
+}
 
 /// 8-bit SPI Word (default)
 pub type U8 = u8;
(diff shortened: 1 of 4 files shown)


src/spi/mod.rs
```diff
@@ -3,6 +3,15 @@
...
 pub mod nb;
 
+/// 8-bit SPI Word (default)
+pub type U8 = u8;
+/// 9-bit SPI Word
+pub struct U9;
+/// 16-bit SPI Word
+pub struct U16;
+/// 18-bit SPI Word
+pub struct U18;
```
Rahix, line 13: How are these zero-sized types supposed to be used?  I think you'd need something more like

```rust
pub struct U9(pub u16);
pub type U16 = u16;
pub struct U18(pub u32);
```

burrbull: You are right it was stupid for me. But you suggestion will not work also.
It should be sumething like:
```rust
pub trait SpiWord {
   type Word;
}
pub struct U9;
impl SpiWord for U9 {
    type Word = u16;
}
```

Rahix: I see, this solution also works.  However, now there isn't any implicit type annotation anymore so when calling, for example, `Transfer::transfer()` you'll have to be explicit about the datatype.  This means you cannot call it like

```rust
spi.transfer(&buf);
```

but you'll have to call it as

```rust
Transfer::<U8>::transfer(&mut spi, &buf);
```

My suggestion was to have newtype wrappers around the odd bitwidths and then store an array of those.  This would allow the compiler to infer the type automatically:

```rust
let buf: [U9; 16] = [U9(0), U9(23), ...];
spi.transfer(&buf);
```

burrbull: Wrapping each element of array is a pain

Dirbaio:
1. u32 should be added: stm32h7xx-hal has an impl for Word=u32.
2. This is fixing the in-RAM representation of u9 etc. Perhaps a chip wants the 9 bits be msb-aligned inside the u16 instead of lsb-aligned to be able to DMA. Without the bound, a HAL can define their own WeirdU9 newtype that handles this and take `&[WeirdU9]` which already has the right repr to DMA straight out of it. Restricting word types makes this impossible.
3. I'm not sure if the benefit is worth the extra complexity, with new types, a new trait, and associated types.

src/spi/mod.rs
```diff
@@ -1,8 +1,42 @@
...
+
+impl Sealed for U9 {}
+impl Sealed for U16 {}
```
ryankurte, line 24: i think i would prefer not to alias `u8` and `u16` so that it's still obvious from the interface these are primitive types.
the other question i have is whether these should be _markers_ or _actual data types_, and whether approach this can adequately describe non-byte word sizes.

it seems to me that if you're running in 9-bit mode you're more than likely to be using arrays of 16-bit aligned words, and with 18-bit mode either 24-bit (if you have sub-word addressing) or 32-bit aligned words (if you do not). On one hand, it makes sense to be able to run over say `&[U9]`, on the other, you may need to control the representation of this in memory for things like DMA.

i wonder whether an approach similar to I2C might describe this better, though i also don't think this covers all cases and it _definitely_ makes the bounds more complex...

```rust
trait SpiWord {
  type Data;
}

impl SpiWord for u8 {
  type Data = u8;
}

impl SpiWord for U9 {
  type Data = u16;
}
```

PR #321, opened 2021-11-03
u8 as default SPI Word

burrbull:
Part of #320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


Sh3Rm4n (commented):
I think this is a sensible default and may also be expanded to other traits, like the Serial traits?

Dirbaio (approved):
I like this! I can't think of any downside, and the *vast majority* of SPI uses are u8.

We should do the same for uart and other traits.

ryankurte (approved):
seems reasonable to me, any other opinions @rust-embedded/hal ?

therealprof:
LGTM!

PR #322, opened 2021-11-03
Remove docs mentioning the Default marker traits.

Dirbaio:
These were removed in https://github.com/rust-embedded/embedded-hal/pull/289 but
the doc comments were not updated accordingly.


ryankurte (approved):
lgtm, thanks!

bors r+

ryankurte:
ah, could you add a line to the changelog for this too please?

Issue #201, opened 2020-04-11
Need a unified way to specify frequencies / periods


Ben-Lichtman:
A little while back I was trying to create a UART bitbang library.

I took a look at https://crates.io/crates/bitbang-hal however it was incompatible since for nrf51, the Timers usable for a serial port implement From<Duration> but in most other embedded-hal implementors (such as stm32) the timers have their own newtype Hertz struct which is used to set timers (stm32f4xx_hal::time::Hertz).

The overall result is that it is impossible to have a bitbang-hal library that is generic over many types of processors.

If embedded-hal provided a Hertz / Period struct, or possibly required From<Duration> for the Periodic trait, this incompatibility could be eliminated in the downstream crates.

Ben-Lichtman:
see https://github.com/rust-embedded/embedded-hal/issues/211

AlyoshaVasilieva:
IMO specifying the *Time* type of Timers as a *rate*, like the STM32 HALs do, is incorrect. Requiring `From<Duration>` or similar seems reasonable since it would actually make the timer trait usable. I also ran into this while writing a driver. In addition to not being able to use timers generically (or it being very hard), the STM32 HALs are effectively limited to 1 second being the longest any timer can run, and cannot handle times like 800ms since it's between 1 and 2 Hz.

ryankurte:
requiring `From<Duration>` would be a neat solve, though, there are some [drawbacks](https://github.com/fluentech/embedded-time#duration) to `core::time::Duration` for embedded use. we could perhaps look to using `embedded_time::Duration` for this bound?

newAM:
> requiring `From<Duration>` would be a neat solve, though, there are some [drawbacks](https://github.com/fluentech/embedded-time#duration) to `core::time::Duration` for embedded use. we could perhaps look to using `embedded_time::Duration` for this bound?

`embedded-time` is a big dependency.  I think the idea of an light-weight duration type is sound; but ideally that functionality would be separated into its own crate to reduce bloat.

`embedded-hal` compile time:
```console
$ hyperfine --prepare 'cargo clean' 'cargo build --target thumbv7em-none-eabi'
Benchmark #1: cargo build --target thumbv7em-none-eabi
  Time (mean ± σ):     615.6 ms ±  45.0 ms    [User: 531.1 ms, System: 117.1 ms]
  Range (min … max):   586.7 ms … 737.7 ms    10 runs
```

`embedded-time` compile time:
```console
[...]

PR #323, opened 2021-11-03
spi: group into read-only, write-only and read-write traits.

Dirbaio:
This PR depends on #331

## The problem

Currently, embedded-hal has independent traits for every single method. The rationale is to allow HALs to only implement what they support. For example, a HAL could allow setting up an SPI with a MOSI pin but no MISO pin. This would be a "write-only" SPI, so the resulting struct would implement `write` and `write_iter`, but not other methods like `read` or `transfer`. This is awesome, because it enforces *at compile time* that you can't pass a write-only SPI to a driver that needs a read-write SPI.

However, splitting *one trait per method* is way too granular. It leaves too much room for variability in HAL implementations. I've done a quick survey of which HAL implements which traits, here's the result:

|                         | Transfer | Write | WriteIter | Transactional |
|-------------------------|----------|-------|-----------|---------------|
| stm32f1xx-hal           | x        | x     |           |               |
| stm32f3xx-hal           | x        | x     |           |               |
| stm32f4xx-hal           | x        | x     | x         | x             |
| stm32f7xx-hal           | x        | x     | x         |               |
| stm32h7xx-hal           | x        | x     |           |               |
| stm32l0xx-hal           | x        | x     |           |               |
| stm32l4xx-hal           | x        | x     |           |               |
| stm32g4xx-hal           | x        | x     |           |               |
| David-OConnor/stm32-hal | x        | x     |           |               |
| nrf-hal SPI             | x        | x     | x         |               |
| nrf-hal SPIM            | x        | x     |           |               |
| esp32-hal               | x        | x     | x         |               |
| esp-idf-hal             | x        | x     | x         |               |
| linux-embedded-hal      | x        | x     |           | x             |
| atsamd-hal              | x        | x     | x         |               |
| lpc55s6x-hal            | x        | x     | x         |               |
| imxrt-hal               | x        | x     | x         |               |
| e310x-hal               | x        | x     | x         |               |

As you can see, all HALs implement `write` and `transfer`, but only about half implement `write_iter` and almost none implements `transactional`. 

The result is drivers can't reliably rely on `Transactional` or `WriteIter` being present, so they stick to `Transfer` and `Write` to ensure maximum HAL compatibility. This in turn means HALs have little incentive to implement `Transactional` and `WriteIter` since no drivers use them. This causes a chicken-and-egg situation, which in the end makes `Transactional` and `WriteIter` useless for drivers in practice.

(Too much freedom for HALs causing traits to not be useful for drivers is a common theme in e-h. For example #312 and #201 are the same. IMO e-h should be slightly more opinionated in general. Just enough opinionatedness to force HAL impls to be consistent to be useful for drivers.)

## The solution

The solution is to split the traits *by hardware capability*, not by method.

- If an SPI is capable of `write`, it's capable of `write_iter`. It can write bytes, after all. There's no reason to allow a HAL to implement `write` but not `write_iter`.
- If an SPI is capable of `transfer`, there's no reason it can't  `transfer_inplace`, `transactional`, `read`, `write`, etc. If it can read *and* write, it should be able to do all the kinds of reads/writes we have.

So, I propose we split the traits like this:
- `Read`: read-only SPI.
- `Write`: write-only SPI. It supports all the forms of writing: `write` and `write_iter`.
- `ReadWrite`: read-write SPI. It requires `Read` and `Write`, and supports all the forms of bidirectional transfers: `transfer`, `transfer_inplace`, `transactional`.

This way HALs keep the "good" freedom to declare their SPI is readonly/writeonly, but they no longer have the "bad" freedom of only implementing the basic traits making the advanced traits useless.

## Other traits

I believe this should be also applied to the other traits. I propose discussing this in the context of SPI here, and if the consensus is it's good I'll send PRs doing the same for the other traits.

### DIFF
--- a/src/spi/blocking.rs
+++ b/src/spi/blocking.rs
@@ -2,75 +2,54 @@
 
 use super::ErrorType;
 
-/// Blocking transfer with separate buffers
-pub trait Transfer<Word = u8>: ErrorType {
-    /// Writes and reads simultaneously. `write` is written to the slave on MOSI and
-    /// words received on MISO are stored in `read`.
-    ///
-    /// It is allowed for `read` and `write` to have different lengths, even zero length.
-    /// The transfer runs for `max(read.len(), write.len())` words. If `read` is shorter,
-    /// incoming words after `read` has been filled will be discarded. If `write` is shorter,
-    /// the value of words sent in MOSI after all `write` has been sent is implementation-defined,
-    /// typically `0x00`, `0xFF`, or configurable.
-    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error>;
-}
-
-impl<T: Transfer<Word>, Word: Copy> Transfer<Word> for &mut T {
-    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error> {
-        T::transfer(self, read, write)
-    }
-}
-
-/// Blocking transfer with single buffer (in-place)
-pub trait TransferInplace<Word: Copy = u8>: ErrorType {
-    /// Writes and reads simultaneously. The contents of `words` are
-    /// written to the slave, and the received words are stored into the same
-    /// `words` buffer, overwriting it.
-    fn transfer_inplace(&mut self, words: &mut [Word]) -> Result<(), Self::Error>;
-}
-
-impl<T: TransferInplace<Word>, Word: Copy> TransferInplace<Word> for &mut T {
-    fn transfer_inplace(&mut self, words: &mut [Word]) -> Result<(), Self::Error> {
-        T::transfer_inplace(self, words)
-    }
-}
-
-/// Blocking read
+/// Blocking read-only SPI
 pub trait Read<Word: Copy = u8>: ErrorType {
     /// Reads `words` from the slave.
     ///
     /// The word value sent on MOSI during reading is implementation-defined,
     /// typically `0x00`, `0xFF`, or configurable.
     fn read(&mut self, words: &mut [Word]) -> Result<(), Self::Error>;
+
+    /// Reads all slices in `words` from the slave as part of a single SPI transaction.
+    ///
+    /// The word value sent on MOSI during reading is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
+    fn read_transaction(&mut self, words: &mut [&mut [Word]]) -> Result<(), Self::Error>;
 }
 
 impl<T: Read<Word>, Word: Copy> Read<Word> for &mut T {
     fn read(&mut self, words: &mut [Word]) -> Result<(), Self::Error> {
         T::read(self, words)
     }
-}
 
-/// Blocking write
-pub trait Write<Word: Copy = u8>: ErrorType {
-    /// Writes `words` to the slave, ignoring all the incoming words
-    fn write(&mut self, words: &[Word]) -> Result<(), Self::Error>;
-}
-
-impl<T: Write<Word>, Word: Copy> Write<Word> for &mut T {
-    fn write(&mut self, words: &[Word]) -> Result<(), Self::Error> {
-        T::write(self, words)
+    fn read_transaction(&mut self, words: &mut [&mut [Word]]) -> Result<(), Self::Error> {
+        T::read_transaction(self, words)
     }
 }
 
-/// Blocking write (iterator version)
-pub trait WriteIter<Word: Copy = u8>: ErrorType {
+/// Blocking write-only SPI
+pub trait Write<Word: Copy = u8>: ErrorType {
+    /// Writes `words` to the slave, ignoring all the incoming words
+    fn write(&mut self, words: &[Word]) -> Result<(), Self::Error>;
+
+    /// Writes all slices in `words` to the slave as part of a single SPI transaction, ignoring all the incoming words
+    fn write_transaction(&mut self, words: &[&[Word]]) -> Result<(), Self::Error>;
+
     /// Writes `words` to the slave, ignoring all the incoming words
     fn write_iter<WI>(&mut self, words: WI) -> Result<(), Self::Error>
     where
         WI: IntoIterator<Item = Word>;
 }
 
-impl<T: WriteIter<Word>, Word: Copy> WriteIter<Word> for &mut T {
+impl<T: Write<Word>, Word: Copy> Write<Word> for &mut T {
+    fn write(&mut self, words: &[Word]) -> Result<(), Self::Error> {
+        T::write(self, words)
+    }
+
+    fn write_transaction(&mut self, words: &[&[Word]]) -> Result<(), Self::Error> {
+        T::write_transaction(self, words)
+    }
+
     fn write_iter<WI>(&mut self, words: WI) -> Result<(), Self::Error>
     where
         WI: IntoIterator<Item = Word>,
@@ -79,7 +58,7 @@ impl<T: WriteIter<Word>, Word: Copy> WriteIter<Word> for &mut T {
     }
 }
 
-/// Operation for transactional SPI trait
+/// Operation for ReadWrite::transaction
 ///
 /// This allows composition of SPI operations into a single bus transaction
 #[derive(Debug, PartialEq)]
@@ -91,18 +70,46 @@ pub enum Operation<'a, Word: 'static + Copy = u8> {
     /// Write data out while reading data into the provided buffer
     Transfer(&'a mut [Word], &'a [Word]),
     /// Write data out while reading data into the provided buffer
-    TransferInplace(&'a mut [Word]),
+    TransferInPlace(&'a mut [Word]),
 }
 
-/// Transactional trait allows multiple actions to be executed
-/// as part of a single SPI transaction
-pub trait Transactional<Word: 'static + Copy = u8>: ErrorType {
-    /// Execute the provided transactions
-    fn exec<'a>(&mut self, operations: &mut [Operation<'a, Word>]) -> Result<(), Self::Error>;
+/// Blocking read-write SPI
+pub trait ReadWrite<Word: Copy = u8>: Read<Word> + Write<Word> {
+    /// Writes and reads simultaneously. `write` is written to the slave on MOSI and
+    /// words received on MISO are stored in `read`.
+    ///
+    /// It is allowed for `read` and `write` to have different lengths, even zero length.
+    /// The transfer runs for `max(read.len(), write.len())` words. If `read` is shorter,
+    /// incoming words after `read` has been filled will be discarded. If `write` is shorter,
+    /// the value of words sent in MOSI after all `write` has been sent is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
+    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error>;
+
+    /// Writes and reads simultaneously. The contents of `words` are
+    /// written to the slave, and the received words are stored into the same
+    /// `words` buffer, overwriting it.
+    fn transfer_in_place(&mut self, words: &mut [Word]) -> Result<(), Self::Error>;
+
+    /// Execute multiple actions as part of a single SPI transaction
+    fn transaction<'a>(
+        &mut self,
+        operations: &mut [Operation<'a, Word>],
+    ) -> Result<(), Self::Error>;
 }
 
-impl<T: Transactional<Word>, Word: 'static + Copy> Transactional<Word> for &mut T {
-    fn exec<'a>(&mut self, operations: &mut [Operation<'a, Word>]) -> Result<(), Self::Error> {
-        T::exec(self, operations)
+impl<T: ReadWrite<Word>, Word: Copy> ReadWrite<Word> for &mut T {
+    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error> {
+        T::transfer(self, read, write)
+    }
+
+    fn transfer_in_place(&mut self, words: &mut [Word]) -> Result<(), Self::Error> {
+        T::transfer_in_place(self, words)
+    }
+
+    fn transaction<'a>(
+        &mut self,
+        operations: &mut [Operation<'a, Word>],
+    ) -> Result<(), Self::Error> {
+        T::transaction(self, operations)
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed blanket impl of `DelayUs` not covering the `delay_ms` method.
 ### Changed
 - `spi`: traits now enforce all impls on the same struct (eg `Transfer` and `Write`) have the same `Error` type. 
+- `spi/blocking`: unified traits into `Read`, `Write`, `ReadWrite`.
+- `spi/blocking`: renamed Transactional `exec` to `batch`.
+- `spi/blocking`: Added `read_batch`, `write_batch` methods.
 
 ### Changed
 - `digital`: traits now enforce all impls on the same struct have the same `Error` type.


ryankurte:
huh, definitely an interesting concept. i like the idea of grouping, it's a neat solution to offer default impls for approaches like `transactional`, and it _seems_ like combining these could substantially decrease the complexity for drivers using the trait types?

for example, SPI bindings like:
```rust
where
  T: Read<u8, Error=E> + Write<u8, Error=E> + Transactional<u8, Error=E>
```
would become:
```rust
where
  T: WriteRead<u8, Error=E>
```

it also raises the question for other traits, obviously most of these are orthogonal (and we've already done this for things like stateful pins) but for example, is there any reason to have separate `i2c` `write` and `write_read` traits when you can't usually interact with an `i2c` device without both?

Dirbaio:
> is there any reason to have separate i2c write and write_read traits when you can't usually interact with an i2c device without both?

Yeah I think i2c should be a single trait. I don't think "write-only i2c" exists, right?

Dirbaio:
The bound would be even simpler, just

```rust
where T: WriteRead
```

The `Error=E` was needed to enforceforce all traits to have the same error type. Now the traits already enforce that, so the `E` generic param in the driver struct can be removed, using `T::Error` instead.

ryankurte:
> Yeah I think i2c should be a single trait. I don't think "write-only i2c" exists, right?

not afaik

> The Error=E was needed to enforce all traits to have the same error type. Now the traits already enforce that, so the E generic param in the driver struct can be removed, using T::Error instead.

i _think_ you're still going to need the error bound at a driver level to effectively wrap bus / driver errors anyway (for driver methods, like `pub fn do_something(&mut self) -> DeviceError<BusError, PinError>`)


this seems like an improvement to me, any other votes @rust-embedded/hal ?

Dirbaio:
You still need it in error enums, but you don't need it in the driver struct itself:

```rust
enum MyDriverError<SpiError, PinError> {
    Spi(SpiError),
    Pin(PinError),
}

struct MyDriver<S, P> { spi: S, pin: P }

impl<S: spi::ReadWrite, P: OutputPin> MyDriver<S, P> {
   fn do_something(&mut self) -> Result<(), MyDriverError<S::Error, P::Error> {
     ...
   }
}

File: src/spi/blocking.rs [hunk 1]
```diff
@@ -1,98 +1,76 @@
...
     /// typically `0x00`, `0xFF`, or configurable.
     fn read(&mut self, words: &mut [W]) -> Result<(), Self::Error>;
+
+    /// Reads all slices in `words` from the slave as part of a single SPI transaction.
+    ///
+    /// The word value sent on MOSI during reading is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
+    fn read_batch(&mut self, words: &mut [&mut [W]]) -> Result<(), Self::Error> {
+        for buf in words {
+            self.read(buf)?;
```
eldruin, line 29: This default impl would not work for a `Read + ManagedCS` impl as `read()` would assert/deassert CS for each slice.

src/spi/blocking.rs
```diff
@@ -1,98 +1,76 @@
...
-
-impl<T: Write<W>, W> Write<W> for &mut T {
-    type Error = T::Error;
 
-    fn write(&mut self, words: &[W]) -> Result<(), Self::Error> {
-        T::write(self, words)
+    /// Writes all slices in `words` to the slave as part of a single SPI transaction, ignoring all the incoming words
+    fn write_batch(&mut self, words: &[&[W]]) -> Result<(), Self::Error> {
+        for buf in words {
+            self.write(buf)?;
```
eldruin, line 49: Same thing with `Write+ManagedCS`

src/spi/blocking.rs
```diff
@@ -1,98 +1,76 @@
...
-    type Error: crate::spi::Error;
 
     /// Writes `words` to the slave, ignoring all the incoming words
     fn write_iter<WI>(&mut self, words: WI) -> Result<(), Self::Error>
     where
-        WI: IntoIterator<Item = W>;
+        WI: IntoIterator<Item = W>,
+    {
+        for word in words {
+            self.write(&[word])?;
```
eldruin, line 60: same.

src/spi/blocking.rs
```diff
@@ -117,20 +95,47 @@ pub enum Operation<'a, W: 'static = u8> {
...
-    /// Execute the provided transactions
-    fn exec<'a>(&mut self, operations: &mut [Operation<'a, W>]) -> Result<(), Self::Error>;
+    /// Execute multiple operations as part of a single SPI transaction
+    fn batch<'a>(&mut self, operations: &mut [Operation<'a, W>]) -> Result<(), Self::Error> {
+        for op in operations {
+            match op {
+                Operation::Read(words) => self.read(words)?,
+                Operation::Write(words) => self.write(words)?,
+                Operation::Transfer(read, write) => self.transfer(read, write)?,
+                Operation::TransferInplace(words) => self.transfer_inplace(words)?,
```
eldruin, line 122: For completeness: same

eldruin (commented):
I disagree with a few statements in the problem description.
For example, a good driver which needs `transactional` would encourage HAL impls to support it.
I see the ecosystem development to be much more of a hand-in-hand thing.
I should note that we used to have 20-line-ish blanket impls for `Transactional`, so supporting it given `Write` and `Transfer` is pretty trivial.

Nevertheless, I think the solution is a good idea, including the default method impls and forcing the error types to match.

The default impls here would clash with `ManagedCS`, though.

Dirbaio:
>  For example, a good driver which needs transactional would encourage HAL impls to support it.

A driver author writes a driver because they need it for a project they're working on *now*. The author has two options:
1. Fork their HAL to add support for Transactional, PR it, then use Transactional in the driver.
2. Use Write/Transfer as needed, do the `for` loop themselves in their driver.

Option 2 is way way easier. The author wants their driver to work *now*. Why would they choose option 1?

Dirbaio:
> The default impls here would clash with ManagedCS, though.

`ManagedCs` would specify "for each method call of the SPI traits, CS has to be driven low before the entire method call, then driven high after". This would apply to *all* methods in the trait, including the ones with a defualt impl.

This means ManagedCs impls MUST override all the spi methods, not using any default impl. If they don't, they're wrong. It's not a conflict.

eldruin:
> Option 2 is way way easier. The author wants their driver to work _now_. Why would they choose option 1?

I agree option 2 can be easier but I, for one, did implement `transactional` for `linux-embedded-hal` including doing it in `rust-i2cdev` _while doing driver contract work_. And that was way more work than taking a blanket impl.
Not everybody that is writing a driver is in the same situation and I would not portray it as such.

re. `ManagedCS`: I agree all HALs have to do is implement the methods themselves again. I just wanted to highlight that there is an use-case where the default impls here do the wrong thing. We should be aware of this and document it clearly albeit in #245.

Dirbaio:
> re. `ManagedCS`: I agree all HALs have to do is implement the methods themselves again. I just wanted to highlight that there is an use-case where the default impls here do the wrong thing. We should be aware of this and document it clearly albeit in #245.

Agreed, it's kind of a footgun. Maybe we could not have default method impls? Adding a trait method with a default impl is backwards-compatible, but arguably not in this case because it would cause ManagedCs impls that were previously correct to become incorrect?

Maybe a different ManagedCs design could avoid these advantages? A closure-based one? Not sure if it's worth it, the existing design is super nice and simple, anything else wlil be more complex.

Sympatron:
> Yeah I think i2c should be a single trait. I don't think "write-only i2c" exists, right?

There are devices where a write-only  I²C would be enough to use them, because they can't be read anyway.
But from a HAL's perspective I don't think there is any MCU that has a write-only I²C mode. So any sane HAL implementation could implicitly read, too. Which would make restricting a driver to `Write` kind of pointless even if it only used `write()`.

GrantM11235:
I think the best solution is to remove the default method implementations for now.

If we can find a solution to the chipselect problem (and I am hopeful that we can), we should be able to add the default method implementations back as a non-breaking change.

Dirbaio:
Removed the default method impls.

eldruin (commented):
Looking good, thanks!
Left to do would be:
- Changelog entry
- Adaption of a HAL (ideally 2) to make sure we are not missing anything, since this touches traits that currently work fine, like I said earlier.

burrbull:
> Removed the default method impls.

I'm fine with this. But instead of deleted code, I'd want to see implementation examples in docs.

GrantM11235:
I don't think that it is necessary to include implementation examples in the docs, it would just be extra clutter for users of the traits to sift through. I think that HAL authors will have no trouble implementing the traits using the existing documentation.

Sh3Rm4n:
Good point, that this documentation is not really relevant for the user. The documentation of the trait should gear towards the user of the trait just as well as the implementor. 

We could either add a `# Implementation Example` section to the docs or link to another block  or to `examples/` for implementation examples. With [`-Z rustdoc-scrape-examples=examples`](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/unstable.md#scrape-examples) in sight, this might be the best place for the documentation.

Dirbaio:
I've split the "unify error types" to #331, also including unifying across `nb` and `blocking`.

Dirbaio:
> Left to do would be the adaption of a HAL (ideally 2) to make sure we are not missing anything, since this touches traits that currently work fine, like I said earlier.

Working HAL implementation for nRF chips here https://github.com/embassy-rs/embassy/pull/552. No issues discovered.

The boilerplate for the `batch` and `iter` methods is slightly annoying, but IMO is still worth it. I plan on changing it to do [DMA chaining](https://infocenter.nordicsemi.com/topic/ps_nrf52840/spim.html?cp=4_0_0_5_24_3#concept_svj_wt2_wr) on nrf in embassy, but probably not soon.

eldruin:
Thanks @Dirbaio!
What is the reason to add `read_batch` and `write_batch` other than being a slightly simpler transactional `batch` call?
I have gone through https://github.com/embassy-rs/embassy/pull/552 and the impls seem straight-forward. However, the PR includes lots of unrelated stuff and I am unsure if I missed something. It would be more understandable to see the adaptions for this separately.

Dirbaio:
Without `read_batch/write_batch` it's impossible to read/write multiple buffers in a single transaction for impls that impl only Write or only Read. This will be a problem when combined with ManagedCs: it would force users to fall back to manual CS management.

Arguably this is only a problem for Read, since Write has write_iter though, which you can use to chain multiple buffers. I don't see an easy way to make a `read_iter` though, it seems it requires GATs.

I have no strong opinion, I'm fine with removing them, I just thought they'd be handy.

ryankurte:
> Without read_batch/write_batch it's impossible to read/write multiple buffers in a single transaction for impls that impl only Write or only Read. This will be a problem when combined with ManagedCs: it would force users to fall back to manual CS management.

i suspect for these cases you could use [`Iterator::Chain`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain) with the `write_iter` or `read_iter` methods?

i have a weak preference for eliding them to minimise the API surface but, would be okay either way.

Dirbaio:
> @ryankurte  I suspect for these cases you could use Iterator::Chain with the write_iter or read_iter methods?

- There's no read_iter (it'd require GATs, nightly-only)
- `iter`-based APIs work really badly with DMA so in the #334 PR I've proposed only having read_batch/write_batch which can be DMA'd efficiently. Having only `iter` methods in blocking and only `batch` methods in async would be an inconsistency.

Dirbaio:
Rebased.

Also, renamed `read_batch`, `write_batch`, `batch` to `read_transaction`, `write_transaction`, `transaction`. It highlights the "single transaction" part which is quite important and the entire reason the methods exist.

ping @eldruin @therealprof @ryankurte thoughts, now that #331 is merged?

Dirbaio:
Rebased for `W->Word` rename.

What's the final decision regarding `read_transaction, write_transaction`? Do we keep them or not?

src/spi/blocking.rs
```diff
@@ -91,18 +70,46 @@ pub enum Operation<'a, Word: 'static + Copy = u8> {
...
+pub trait ReadWrite<Word: Copy = u8>: Read<Word> + Write<Word> {
+    /// Writes and reads simultaneously. `write` is written to the slave on MOSI and
+    /// words received on MISO are stored in `read`.
+    ///
+    /// It is allowed for `read` and `write` to have different lengths, even zero length.
+    /// The transfer runs for `max(read.len(), write.len())` words. If `read` is shorter,
+    /// incoming words after `read` has been filled will be discarded. If `write` is shorter,
+    /// the value of words sent in MOSI after all `write` has been sent is implementation-defined,
+    /// typically `0x00`, `0xFF`, or configurable.
+    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error>;
```
ryankurte, line 86: another small bikeshed... if we have `TransferInPlace` and `transfer_in_place` would it make sense to have `TransferSplit` and `transfer_split` as the reciprocal?

```suggestion
    fn transfer_split(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error>;
```

src/spi/blocking.rs
```diff
@@ -2,75 +2,54 @@
...
     fn write(&mut self, words: &[Word]) -> Result<(), Self::Error>;
-}
 
-impl<T: Write<Word>, Word: Copy> Write<Word> for &mut T {
-    fn write(&mut self, words: &[Word]) -> Result<(), Self::Error> {
-        T::write(self, words)
-    }
-}
+    /// Writes all slices in `words` to the slave as part of a single SPI transaction, ignoring all the incoming words
+    fn write_transaction(&mut self, words: &[&[Word]]) -> Result<(), Self::Error>;
```
ryankurte, line 36: in `transaction` we call them `operations` rather than `words`, here they're not exactly operations but also not really words. non critical, but, are we happy with this / is there any clearer name?

ryankurte (commented):
> Without read_batch/write_batch it's impossible to read/write multiple buffers in a single transaction for impls that impl only Write or only Read. This will be a problem when combined with ManagedCs: it would force users to fall back to manual CS management.

is this a thing in practice? while many _peripherals_ might be unidirectional i can't think of an SPI  controller i have used that only supports write _or_ read.

i can't find the latest in my mailbox but iirc the latest approach to `ManagedCS` looks like some kind of closure, so you'll be able to call `.with_cs( |s| { s.read(..)?; s.read(..)? })` to chain operations without using `transactional`.
(@GrantM11235 do you have a branch / PR for the work you were doing on this?) 

> What's the final decision regarding read_transaction, write_transaction? Do we keep them or not?

i really don't know, gut feeling is they're not needed / demonstrated, but i can see they'd be more annoying to add later. @eldruin any opinions?

src/spi/blocking.rs
Dirbaio: The name was `transfer` before. IMO separate-buffers transfer is the "normal" one, in-place is the "weird" one, so I like the `transfer` name more.

Sympatron:
I like the `transaction` methods and think they would be useful for driver authors.

src/spi/blocking.rs
eldruin: At first sight `transfer_split` sounds like the transfer itself is somehow split, which is rather confusing. I would also keep the `transfer` name.

eldruin:
After further discussion with @ryankurte, we would favor removing the `read_transaction`/ `write_transaction` methods.
We appreciate the handiness of the methods but the boilerplate increase to use the transactional interface instead is rather minimal and we would prefer to keep the API surface smaller.
```
let a = [0; 2];
let b = [0; 3];
write_transaction(&[&a, &b]);
transaction(&[Operation::Write(&a), Operation::Write(&b)]);
```
As pointed out above, any SPI controller supports read and write in practice, so using the transactional interface would be possible.
As for restricting use to only read _or_ write with the `Read`/`Write` traits, I do not see a real chance of making a mistake and using the wrong method or any other real "misuse" protection.

Dirbaio:
Do we want to support readonly, writeonly, and "readwrite but in readonly/writeonly mode because it's missing pins" SPI peripherals?

If the answer is yes, then we should have the Read, Write traits be "first class citizens", complete with with their `read_transaction, write_transaction` methods.

If the answer is no (it seems you're arguing for no), then we should merge all SPI traits in one, like #339. (and then not have `read_transaction, write_transaction`, of course).

Having separate traits but without separate `read_transaction, write_transaction` is a weird "yes but no but yes" answer to the above question.

eldruin:
I think merging all traits into one (1.) makes sense except there is a general problem with implementing `Transactional`, then I would prefer (2.).
Again the options:
1. `Spi` trait with `Read`, `Write` and `Transactional`
2. `ReadWrite`, `Transactional`
3. `Read`, `Write`, `Transactional` (status quo)

Opinions @rust-embedded/hal?

Dirbaio:
What's the problem with Transactional? AFAICT all hardware that can impl read/write/transfer should be able to impl transactional.

burrbull:
Saying about `transfer` and `transactional`.
If all methods will be in 1 trait, we can restore default implementations removed in #289 as now they will not conflict with external implementations.

Dirbaio:
I don't think making it a single trait fixes the original issue raised in #289. Also, the arguments against the Default I raised in https://github.com/rust-embedded/embedded-hal/pull/289#issuecomment-879322878 still apply IMO

eldruin:
AFAIK there is no problem with `Transactional`, I was just wondering.
If we merge all into one trait, what would be the problem with some default implementations like the following?
```rust
// I just wrote this down, please ignore trivial mistakes
pub trait Spi {
  fn write(&mut self, data: &[Word]) -> Result<(), Self::Error> {
    self.transaction(&[Operation::Write(data)])
  }
  fn read(&mut self, data: &mut [Word]) -> Result<(), Self::Error> {
    self.transaction(&[Operation::Read(data)])
  }
  fn transaction(&mut self, ops: &[Operation]) -> Result<(), Self::Error>;
}
```

src/spi/blocking.rs
ryankurte: :+1:

ryankurte:
> Doesn't the same argument apply to the (read-write) transaction method too though? 

yep, though it is also useful of itself for DMA-ing a set of transactions on platforms that support this.

> If we merge all into one trait, what would be the problem with some default implementations like the following?

one of the side effects of the newer option for `ManagedCS` is you have to be very careful with default implementations because you can fall through to the wrong ones, i think it's probably better to leave them off for now.

> Having separate traits but without separate read_transaction, write_transaction is a weird "yes but no but yes" answer to the above question.

i still prefer this, though in practice i have never only used `Write` or `Read` (and tend to just use `Transactional`). i think it is less "yah nah yeah" and more that there are still changes to come ^_^

Dirbaio:
There's another use case for readonly/writeonly SPI: limited DMA channels.

STM32 needs one DMA channel for read, another for write. If the user only wants to write it's very rude to force them to waste one DMA channel for writing. Channels are very scarce on low-end chips (smallest G0, L0 only have 5!). In chips without DMAMUX, even if there are more channels, they have no/limited remappability, so the wasted channel could cause a conflict.

`embassy-stm32` SPI driver is generic on the channels, and only impls the traits if the channels in the needed directions are set: https://github.com/embassy-rs/embassy/blob/ddf8c99a938f7d2c4bb6e3ea9476c487c5b1c047/embassy-stm32/src/spi/mod.rs#L608-L651

This applies more to async, rather than blocking, but I'd like async+blocking to be consistent on this.

Given this, I no longer think unifying SPI into a single trait is a good idea...

----

On `read_transaction`, `write_transaction`: I still don't see why `transaction` is useful but `read/write_transaction` is not. They're both equally useful. 

If HALs are going to have write-only/read-only SPI, you can't use `transaction` on them, so `read/write_transaction` is the solution for writing multiple buffers.

burrbull:
In any case, I believe that it is better to solve the problem of missing implementations in organizational way.
Revisit all known repositories with hals and open PRs or/and issues with list of traits that are not implemented.

eldruin:
Thanks for that important argument @Dirbaio. Let's keep `Read`, `Write` and `Transactional` separate.
As for `read_transaction`/`write_transaction`, @ryankurte had some additional concerns regarding managed CS compatibility. He also made a new proposal at #350

Dirbaio:
Closing in favor of #351 which incorporates all the ideas from here.

PR #324, opened 2021-11-03
Remove traits with unconstrained `Time` associated types.

Dirbaio:
As discussed in https://github.com/rust-embedded/embedded-hal/issues/201 and
https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-910396479 ,
the traits that have an unconstrained `type Time` associated type end up
not being usable for HAL-independent drivers in practice, since there is too
much variation across HALs of what the actual Time type is.

These should be bound to something, probably a `Duration` trait, so that
drivers can actually create and use them. Alternatively embedded-hal could have
actual `struct Duration` and so.

Either way, it's unclear what the final solution would look like. We shouldn't leave
these traits in the 1.0 release, since fixing them later will be a breaking change.
We should temporarily remove them, and add them back once we have figured this out
which won't be a breaking change.


eldruin:
Uh, this removes a few traits.
Before considering this, I would like to see an assessment of which public HALs implement them and which drivers use any of them to see how much downstream code would break.

Sh3Rm4n:
I've created this [source graph query](https://sourcegraph.com/search?q=context:global+timeout:60s+repo:github.com/%28esp-rs%7Cstm32-rs%7Cnrf-rs%7Catsamd-rs%7Cimxrt-rs%7Cxmc-rs%29+%28lang:rust+%5Cb%28WatchdogEnable%7CCountDown%7CPwm%7CCapture%29%5Cb%29+AND+%28lang:rust+type+Time%5Cb+%3D%29&patternType=regexp&case=yes) on the bases of https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-910396479 to see, which crate is implementing these traits and also to give an overview, which `Time` types are used. 

Based on the curated drivers list on [awesome-embedded-rust](https://github.com/rust-embedded/awesome-embedded-rust#driver-crates) and a [quick and dirty script](https://gist.github.com/e7987c28883e35e67c62299ea9cf7106), 
I also created a similar [query](timeout:3m0s repo:github.com/(rahul-thakoor/mma7660fc|JoshMcguigan/tsl256x|eldruin/kxcj9-rs|braun-embedded/rust-dw1000|dbrgn/shtcx-rs|japaric/enc28j60|danielgallagher0/hts221|rainbow-hat-rs/|aw/sram23x|sousandrei/iis2mdc|jamwaffles/ssd1306|japaric/l3gd20|eldruin/tmp1x2-rs|redpfire/usbd-webusb|e-matteson/keypad|kallemooo/adafruit-7segment|mvniekerk/microchip-eeprom-25x|eldruin/tmp006-rs|almindor/st7789|mvniekerk/microchip-tc72-rs|master/tcn75a|cs2dsb/adafruit-alphanum4.rs|eldruin/ads1x1x-rs|daschl/tsic-rs|eldruin/ds1307-rs|leshow/rotary-encoder-hal|japaric/lsm303dlhc|JitterCompany/adxl355-rs|chocol4te/aht20|jamwaffles/sh1106|eldruin/opt300x-rs|NeoBirth/ADXL343.rs|sousandrei/ism330dhcx|mvirkkunen/usbd-serial|twitchyliquid64/usbd-hid|agalakhov/usbd-hid-device|michaelbeaumont/dht-sensor|thejpster/embedded-sdmmc-rs|pcein/adc-mcp3008|eupn/bno055|flott-motion/stepper|eldruin/mcp794xx-rs|JoshMcguigan/shift-register-driver|dbrgn/mcp3425-rs|rust-iot/rust-radio-sx128x|lonesometraveler/max6955|eldruin/embedded-ccs811-rs|eldruin/veml6030-rs|eldruin/veml6075-rs|mvniekerk/adxl313-rs|danielgallagher0/bluenrg|ryankurte/rust-radio-sx127x|eldruin/ad983x-rs|eldruin/eeprom24x-rs|eldruin/pwm-pca9685-rs|dbrgn/sgp30-rs|eldruin/si4703-rs) lang:rust \b(WatchdogEnable|CountDown|Pwm|Capture)\b) for drivers the driver part (which sadly times out). 

But [even](timeout:3m0s repo:github.com/(JoshMcguigan/tsl256x|agalakhov/usbd-hid-device|mvirkkunen/usbd-serial|redpfire/usbd-webusb|eldruin/veml6030-rs|twitchyliquid64/usbd-hid|eldruin/si4703-rs|rahul-thakoor/mma7660fc|braun-embedded/rust-dw1000|eldruin/tmp006-rs|dbrgn/sgp30-rs|eldruin/veml6075-rs|eldruin/opt300x-rs|master/tcn75a|mvniekerk/microchip-tc72-rs|eldruin/pwm-pca9685-rs|jamwaffles/ssd1306|rainbow-hat-rs/|leshow/rotary-encoder-hal|ryankurte/rust-radio-sx127x|kallemooo/adafruit-7segment|eldruin/tmp1x2-rs|aw/sram23x|almindor/st7789|jamwaffles/sh1106|JoshMcguigan/shift-register-driver|dbrgn/shtcx-rs|daschl/tsic-rs|flott-motion/stepper|rust-iot/rust-radio-sx128x) lang:rust \b(WatchdogEnable|CountDown|Pwm|Capture)\b) [smaller](https://sourcegraph.com/search?q=context:global+timeout:3m0s+repo:github.com/%28eldruin/veml6075-rs%7Cbraun-embedded/rust-dw1000%7Cdbrgn/shtcx-rs%7Ckallemooo/adafruit-7segment%7Celdruin/opt300x-rs%7Cmvirkkunen/usbd-serial%7Cagalakhov/usbd-hid-device%7Calmindor/st7789%7Credpfire/usbd-webusb%7Ctwitchyliquid64/usbd-hid%29+lang:rust+%5Cb%28WatchdogEnable%7CCountDown%7CPwm%7CCapture%29%5Cb&patternType=regexp&case=yes)  queries did not lead to any matches at all. So this might be the wrong way to get an idea, which driver uses these traits. (Or there are just no matches at all)

But hopefully this creates a basis for discussion.

ryankurte:
thanks for the analysis! there's definitely room for improvement on em, which either way will be a breakage i guess, but, i still feel a little wary of removing all these (also particularly interested in whether this has flow-on impact on RTIC timer definitions?)

i'm not _sure_ this achieves all goals, but, what if instead of associated types we elected here to define a really basic timestamp / duration type (maybe something like `Ts{ s: u32, us: u32 }` inspired by the classic `struct timespec`?) and then to bound the traits by `Into<Ts>`. that way we're providing a simple primitive and if `embedded-time` (or any other approach) want to work with this it can be implemented outside the hal.

Sh3Rm4n:
I feel like, in previous discussion and  PRs it was made clear, that the most important priority `e-h` should solve, is to provide a **consistent interface between drivers and MCUs**, especially to be able to independently implement a half of the `interface<->driver` combination. 

Everything, that **misses that goal**, should not be part of this crate. And, as of now, the traits which will be removed in this PR **do miss the goal** with the unconstrained `Time` type.

The **fear of breaking many downstream** crates while removing these traits is very real and very reasonable, so this problem should be solved.

> i'm not sure this achieves all goals, but, what if instead of associated types we elected here to define a really basic timestamp / duration type [...]

Even though this sounds like an interesting idea, I propose we should not go this route.  Any discussion about an alternative simple `Time` type sets us back to square one, where we have to come to an agreement **how** this simple solution should look like. With a discussion like that started, this is **still** a blocker for a `v1.0` release, which this PR is **deliberately trying to solve**, by just removing these traits for now. 

IMHO, more important right now is to solve the breakage issue for downstream HAL implementations and maybe inter-op between HAL impls and RTIC apps. First of, I don't think this is a blocker for this PR in particular, but we should have a solution for that concern before `v1.0` is released. 

I have a few ideas, but I'm not really particularly happy with anyone of these (maybe someone else can iterate ontop of these ideas): 

1. **Break the world**. Just remove them. This breakage could either lead to a big hurdle to move to `v1.0` or to the situation, that the implementations just vendor / copy & paste these traits into the HAL implementations. This could lose the uniformity between these implementations over time, as adjustments to these traits can be made anytime. I'd argue this problem already exists, because of the widely different `Time` types used. 
2. **Move** these traits behind a `unstable` **feature flag**, for which the semantic versioning rules are loosened, and breaking changes are ok, at least between minor versions. This reminds me of the `unproven` flag, so this might be a bad idea. Anyway, this would allow the HAL implementations to go-on their lives to keep their current `Time` based implementations and the churn is not unnecessarily higher as it is already to move to `v1.0`.
3. **Move** these traits to a **companion crate** (could still be part of this repository)  to solve point 1., which is advertised as **temporary** from the beginning **until** the `Time` situation is solved. This crate could also be used as a testing bed to iterate on the  `Time` design, including breaking changes until the WG is comfortable to move these into `e-h` again. 

I'm out of ideas, but maybe there are other solutions as well.

burrbull:
Need testers https://github.com/stm32-rs/stm32f4xx-hal/pull/381

burrbull:
Have anyone saw this?

https://github.com/BlackbirdHQ/atat/blob/e5d509bf3871f540fe88f069b409555bee93ab3e/atat/src/lib.rs#L281

burrbull:
`PwmPin` does not contain associated `Time` type. Why it was removed?

Dirbaio:
@burrbull it was designed to work together with `Pwm` which does have the `type Time` problem. IMO the safest is to remove both,  I'm not sure the `PwmPin` trait makes sense "standalone".

We can redesign and add it back later, either as a "Pwm" peripheral with multiple pins/channels, or a standalone "PwmPin" that just allows setting duty % without "caring" about the full peripheral...

burrbull:
> it was designed to work together with `Pwm`

It is not true. `PwmPin` was added much more earlier than `Pwm`

Dirbaio:
I stand corrected then!

I still think it suffers from a similar issue: it has an unconstrained `type Duty;` that can be integer, float... How can a HAL-independent driver make effective use of it?

eldruin (commented):
As discussed in yesterday's meeting, let's remove all traits with unconstrained associated types for the 1.0 release and add them back later on piece by piece as we figure out what constraints we can specify on them, which would make them usable for HAL-independent code.
@therealprof agreed yesterday as well.
Opinions @ryankurte ?

eldruin:
Yeah, we should start some kind of "Get me those traits back" initiative and detail what the types could be :)

eldruin (approved):
Ok, let's merge this as-is and I will update the changelog separately.
For anyone finding this in the future, if you are interested in any of the removed traits, please speak up in the issues we will open about this.
bors r+

niclashoyer:
@eldruin please point to the issues you mentioned. I have several drivers I'd like to publish in the next months that depend on timer traits (especially `CountDown`). I'm using `embedded-time` with https://github.com/drogue-iot/drogue-embedded-timer as a compatibility layer to circumvent the problems mentioned here. I want to speak up to get (reasonably bound) timer traits back in!

eldruin:
Thanks for the reminder @niclashoyer.
Here is the issue with the overall roadmap for these traits: #357

PR #325, opened 2021-11-13
Update readme notes on alpha / 0.2.x compatibility

ryankurte:
i wasn't sure whether to remove the comment about implementing both completely as it seems superfluous given the availability of a `compat` layer?

cc. @eldruin


eldruin (approved):
Looks good to me as-is. That way people have all options.
Thank you!

bors r+

PR #326, opened 2021-11-16
Require all SPI and Serial word types to be `Copy`

GrantM11235:
The word type should always be a builtin integer (`u8`, `u16`, etc) or a newtype to represent an integer with an unusual size, which will almost certainly be `Copy`.

I also removed the `'static` bound from the spi transactional stuff because it appears to be unnecessarily now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Use `u8` as default SPI as Serial Word type
 - The Minimum Supported Rust Version (MSRV) is now 1.46.0
+- Require all SPI and Serial word types to be `Copy`.
 
 ### Added
 - Added `Can` Controller Area Network traits.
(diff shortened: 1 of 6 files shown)


eldruin (commented):
Requiring `Copy` for word types seems fine to me.
I am not sure what the reason for the `'static` bound was so I will leave that to @ryankurte

Dirbaio:
Perhaps we want to require `'static` for all words? I can't imagine a legit reason for it not to be static

ryankurte:
i had to go digging [[1](https://github.com/rust-embedded/embedded-hal/pull/191), [2](https://github.com/rust-embedded/embedded-hal/pull/178#pullrequestreview-348972945)], it appears the justification was to separate the lifetime of the type `W` from that of the buffers `&'a [W]`, i can't imagine i would have added it if unnecessary, but, if the compiler can infer the lifetime of the word type now perhaps that is no longer required.

there aren't a lot of reasons for `W` not to be static, but, it's also seemingly an unnecessary bound so, provided it is demonstrated to still work on both sides i'd be okay with removing the `'static`.

GrantM11235:
I decided to leave the `'static` bound unchanged for now because I don't fully understand it, and it seems to be unrelated to this PR.

eldruin (approved):
Looks good to me, thanks!
bors r+

PR #331, opened 2021-12-19
spi: enforce all traits have the same Error type.

Dirbaio:
Previously discussed in #323, split off to its own PR now.

This PR enforces all SPI trait impls for a given type use the same Error associated type, by moving it to its own ErrorType trait.

## How breaking is this?

I believe this is not very breaking in practice (it won't make upgrading existing code to 1.0 harder), because:

- HALs already use the same error type for all methods (I haven't seen one that doesn't)
- Drivers already often add bounds to enforce the errors are the same. [Example](https://github.com/rust-iot/rust-radio-sx127x/blob/8188c20c89603dbda9592c40a8e3fc5b37769a00/src/lib.rs#L122). Without these bounds, propagating errors up would be more annoying, drivers would need even more generic types `SpiWriteError, SpiReadError, SpiTransferError, SpiTransactionalError`...

## Why is this good?

Traits being able to have different Error types is IMO a case of "bad freedom" for the HALs. Today's traits don't stop HALs from implementing them with different error types, but this would cause them to be incompatible with drivers with these bounds. If traits force error types to be the same, the problem is gone.

## What about other traits?

I believe this should be also applied to the other traits. I propose discussing this in the context of SPI here, and if the consensus is it's good I'll send PRs doing the same for the other traits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Fixed blanket impl of `DelayUs` not covering the `delay_ms` method.
+### Changed
+- `spi`: traits now enforce all impls on the same struct (eg `Transfer` and `Write`) have the same `Error` type. 
 
 ## [v1.0.0-alpha.6] - 2021-11-19
 
--- a/src/spi/blocking.rs
+++ b/src/spi/blocking.rs
@@ -1,10 +1,9 @@
 //! Blocking SPI API
 
-/// Blocking transfer with separate buffers
-pub trait Transfer<W: Copy = u8> {
-    /// Error type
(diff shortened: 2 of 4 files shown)


Dirbaio:
Working HAL implementation for nRF chips here https://github.com/embassy-rs/embassy/pull/552. Updating was quite straightforward as the HAL already used the same error type for all impls. No issues discovered.

burrbull (dismissed):

eldruin (approved):
Sounds like a good idea to me.
Opinions @rust-embedded/hal?

Dirbaio:
Friendly ping @eldruin @therealprof @ryankurte I guess if no one has any concerns in 23 days it's good to go?

Rebased

ryankurte (approved):
lgtm, thanks!

bors r+

PR #332, opened 2021-12-19
Remove unresolved link to `toggleable`

Sh3Rm4n:
In 282320b `mod toggleable` as the default implementation was removed,
which invalidated the link in the doc comment to it.

This just removes the reference to this non-existing mod, to fix the
issue.


Sh3Rm4n:
Should probably have the label "no changelog"

eldruin (approved):
Good catch, thank you!
bors r+

PR #333, opened 2021-12-19
delay: fix blanket impl not covering delay_ms.

Dirbaio:
If an impl overrides the default impl for delay_ms, it was ignored when used through the blanket impl.


eldruin (approved):
Nice catch, thanks!
bors r+

PR #334, opened 2021-12-20
Async

Dirbaio:
Depends on #331 #323

## Separate crate

Following the discussions on #285, this adds the async traits in a separate crate. TLDR of the reasons:
- We want to be able to major-bump EHA independently, to leave the main EH crate at 1.0.
- At the very least we'll have to major-bump when switching from GATs to async-fn-in-traits.
- GAT-based traits require nightly, so they'd have to be behind an "unstable" Cargo feature if they were in EH.

Once we're confident EHA is "1.0 ready", we can move the traits to the main EH crate. This can be done backwards-compatibly: move the traits, then do one last update to EHA making it just reexport them from EH.

## Single repo

This moves `embedded-hal` to a subdir, then adds another for `embedded-hal-async`.

The advantage of having both crates in this repository is that a single PR can update both the blocking and async traits. Since they are close mirrors of each other, it's very likely a change to one (e.g. docs) will affect the other.

## Trait design

The design closely mirrors the blocking traits. 

The only differrence is I have removed the `iter` methods such as `write_iter` because they don't work well with async. Fundamentally operate byte by byte, so the core has to wake up for each one. It can't kick off a single bit DMA transfer and then sleep or do something else.

## HAL use

GAT-based async traits have been in use for a while at [embassy-traits](https://github.com/embassy-rs/embassy/tree/master/embassy-traits) and have worked great. IMO the approach is proven.

Working HAL implementation for nRF chips here https://github.com/embassy-rs/embassy/pull/552. STM32 coming soon.

## TODO:

- [ ] Crate-level docs, README
- [ ] Fix CI scripts
- [ ] Add changelog for the -async crate. Should it go at the root, or should each crate have its own?

### DIFF
--- a/embedded-hal-async/src/serial.rs
+++ b/embedded-hal-async/src/serial.rs
@@ -1,0 +1,74 @@
+//! Serial interface
+
+use core::future::Future;
+pub use embedded_hal::serial::{Error, ErrorKind};
+
+/// Read half of a serial interface
+///
+/// Some serial interfaces support different data sizes (8 bits, 9 bits, etc.);
+/// This can be encoded in this trait via the `Word` type parameter.
+pub trait Read<Word: 'static = u8> {
+    /// Read error
+    type Error: Error;
+
+    /// The future associated with the `read` method.
+    type ReadFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Reads words from the serial interface into the supplied slice.
+    fn read<'a>(&'a mut self, read: &'a mut [Word]) -> Self::ReadFuture<'a>;
+}
+
+impl<T: Read<Word>, Word: 'static> Read<Word> for &mut T {
+    type Error = T::Error;
+    type ReadFuture<'a>
+    where
+        Self: 'a,
+    = T::ReadFuture<'a>;
+
+    fn read<'a>(&'a mut self, read: &'a mut [Word]) -> Self::ReadFuture<'a> {
+        T::read(self, read)
+    }
+}
+/// Write half of a serial interface
+pub trait Write<Word: 'static = u8> {
+    /// Write error
+    type Error: Error;
+
+    /// The future associated with the `write` method.
+    type WriteFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// The future associated with the `flush` method.
+    type FlushFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Writes a single word to the serial interface
+    fn write<'a>(&'a mut self, words: &'a [Word]) -> Self::WriteFuture<'a>;
+
+    /// Ensures that none of the previously written words are still buffered
+    fn flush<'a>(&'a mut self) -> Self::FlushFuture<'a>;
+}
+
+impl<T: Write<Word>, Word: 'static> Write<Word> for &mut T {
+    type Error = T::Error;
+    type WriteFuture<'a>
+    where
+        Self: 'a,
+    = T::WriteFuture<'a>;
+    type FlushFuture<'a>
+    where
+        Self: 'a,
+    = T::FlushFuture<'a>;
+
+    fn write<'a>(&'a mut self, words: &'a [Word]) -> Self::WriteFuture<'a> {
+        T::write(self, words)
+    }
+
+    fn flush<'a>(&'a mut self) -> Self::FlushFuture<'a> {
+        T::flush(self)
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 **/*.rs.bk
 .#*
-/target/
+target
 Cargo.lock
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - `spi`: traits now enforce all impls on the same struct (eg `Transfer` and `Write`) have the same `Error` type. 
+- `spi/blocking`: unified traits into `Read`, `Write`, `ReadWrite`.
+- `spi/blocking`: renamed Transactional `exec` to `batch`.
+- `spi/blocking`: Added `read_batch`, `write_batch` methods.
 
 ## [v1.0.0-alpha.6] - 2021-11-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `Error` traits for Can, SPI, I2C and Serial are implemented for Infallible
 
+### Changed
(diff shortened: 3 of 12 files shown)


burrbull:
This is excellent. But moving `embedded-hal` to subdir breaks all existent PRs. Same as all links.
When we merge crates, we will have the same problem or useless subdir instead.

eldruin:
Thank you for this proposal and the write-up.
However, @ryankurte and I discussed yesterday about creating a separate `embedded-hal-async` repository.
Advantages:
1. This repository stays "single-purpose".
2. No file shuffling now and whenever `embedded-hal-async` is absorbed, since that is the plan.
Drawbacks:
1. Not possible to change blocking/nb traits as well as async traits in a single commit.

Dirbaio:
Git is able to track moved files, I think it should be able to merge across renames? :thinking: 

If moves are a concern I'd leave the subdir in when merging the crates. It's not useless, it'll avoid more file moves if/when we want to maintain more crates here. (Perhaps embedded-hal-compat? Perhaps stuff for sharing spi/i2c buses?)

and re single-purpose: I'd argue async and blocking traits are two sides of the same coin, same purpose :) 

There are more advantages to have stuff in a single repo:

- `embedded-hal-async` depends on `embedded-hal` (for the Error types/traits, and shared stuff like the SPI Mode/Polarity/etc enums). When are `embedded-hal` changes propagated to `embedded-hal-async`? In a single repo it's trivial, you just update both atomically. With separate repos, however: 
  - Option 1: wait for release. Change is done to `embedded-hal` repo, wait for when the change is released in `embedded-hal` crates.io, `embedded-hal-async` repo is updated to the new `embedded-hal` release, corresponding changes are then finally done. Problem: very slow.
  - Option 2: git dep. Change is done to `embedded-hal` repo, `embedded-hal-async` repo is changed to a git dep pointing to the new `embedded-hal` repo, `embedded-hal-async` is updated. Problem: needs manual Cargo.toml changes when releasing since you can't have git deps in crates.io.
- For users using git for both crates, Cargo automatically does the right thing with `[patch.crates-io]` with a single repo. With multiple repos, users have to micromanage the git revisions for both since it's no longer guaranteed that `embedded-hal-async` latest master works with `embedded-hal` latest master. 

From my experience in embassy, managing a set of interdependent crates in separate repos gets really annoying really fast. The probe-rs project is also moving to single-repo for the same reasons. I strongly suggest doing single-repo :)

Dirbaio:
Another option is leaving `embedded-hal` at the root and having `embedded-hal-async` in a subdir. This means no file moving, but the file structure ends up a bit anti-intuitive IMO.

huntc:
I would also like to voice my support for a single repo as the start point. CI, release cadence, team structure and perhaps other factors can influence the decision on multiple vs single repositories. However, I’ve also come to learn that starting with a single repository generally works out well.

Also, async and non-async embedded development solves the same problem, sometimes together, so it feels right to combine them in this repository. We can avoid promoting a division between async and non-async development. :-)

eldruin:
If `embedded-hal-async` would be long-lived, I agree with all the pain points. Especially the inter-dependency problems.
Somehow I imagined `embedded-hal-async` to not live that long in the grand scheme of things, though.
Just have a few major bumps and then integrate into `embedded-hal`, at which point a separate repository would be archived and async traits would just be a regular part of `embedded-hal`.
That is what we are doing with `embedded-can`, for example.

Dirbaio:
`embedded-hal-async` will live for at least min(e-h1.0 gets released, async-fn-in-traits is stable in Rust). I guess the limiting factor will be the second, which can be easily 1 year or more.

Dirbaio:
so perhaps the sweet spot is to have only `async-embedded-hal` in a subdir? I've change the PR to do that.

No pain of moving files, no pain of multiple repos. The only pain is the "weird" file structure, but that's much less than separate repos IMO, and it's temporary until they get merged.

burrbull:
> so perhaps the sweet spot is to have only `async-embedded-hal` in a subdir?
> 

Yes, should be the best way.

eldruin:
That would be an acceptable compromise, thanks!

ryankurte (approved):
lgtm, thanks for bearing with us!

embedded-hal-async/src/serial.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Read error
+    type Error: Error;
+
+    /// The future associated with the `read` method.
+    type ReadFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
+    where
+        Self: 'a;
+
+    /// Reads words from the serial interface into the supplied slice.
+    fn read<'a>(&'a mut self, read: &'a mut [Word]) -> Self::ReadFuture<'a>;
```
ryankurte, line 20: i don't really want to derail this PR further / we can totally discuss elsewhere, but, it seems to me this needs to be able to return a _partially filled_ result.

say you're reading serial payloads of size up to N, to use this trait you're going to need to always poll with `&[Word; 1]` to guarantee you'll return, undermining the benefits of buffering / DMA.

i think in this case the trait would be more useful with `Result<usize, Self::Error>` that tells you how full the buffer is, then the HAL can choose to return on `EOF` / timeout / buffer full as is relevant for the hardware.

Dirbaio: The problem with that is it's not specified when it returns, so different impls will do different things, making it useless for drivers.

In Embassy we specified it [like this](https://github.com/embassy-rs/embassy/blob/master/embassy-traits/src/uart.rs): 
- `read` returns on buffer full.
- `read_until_idle` returns on buffer full, or when at least 1 byte has been received and then the line goes idle.

`read_until_idle` is a separate trait because some chips (nrf) can't do it natively so they need a hardware timer to do it. Forcing everyone to waste a timer when not needed would be rude, so there's a [special impl](https://github.com/embassy-rs/embassy/blob/master/embassy-nrf/src/uarte.rs#L426) supporting it that you can opt-in to use.

in my experience, `read` is still useful if you know your packets have a fixed size, and it maps to a single DMA read operation which is the lowest overhead possible. 

(this also applies to the uart blocking traits, IMO we should have a way to do reads without `nb`. The current `nb` read trait works really badly in the nRF because it can *only* do DMA reads.)

ryankurte: > The problem with that is it's not specified when it returns, so different impls will do different things, making it useless for drivers.

i don't think adding `this _may_ return a partial buffer on timeout or idle` is useless? this is how reads under linux typically work, it just implies the need for a check and buffering at the driver level if you're not doing byte-wise processing.

> ... some chips (nrf) can't do it natively so they need a hardware timer to do it

rude of them (nrf) imo

> in my experience, read is still useful if you know your packets have a fixed size, and it maps to a single DMA read operation which is the lowest overhead possible.

the problem with this approach appears when you know your packet sizes but you somehow end up out-of-sync or with corrupted data, then you have to have application timeouts to recover anyway.

Dirbaio:
Converting to draft, will open PRs for each individual trait, first is #344 . Will close this when they're all opened.

Dirbaio:
Closing in favor of #344 #346 #347 #348 #349

PR #335, opened 2021-12-20
Set merge=union Git attribute for CHANGELOG.md

Dirbaio:
Currently all PRs add a line to CHANGELOG.md, usually at the end. This means
they almost always run into merge conflicts since they are adding different lines at the exact same
place in the file.

With merge=union Git will automatically resolve these merge conflicts by keeping all lines from both sides.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.md merge=union


dbrgn (approved):

eldruin (approved):
That is really neat, thanks!
bors r+

Dirbaio:
needs a "skip changelog" maybe?

PR #338, opened 2021-12-20
digital: enforce all traits have the same Error type.

Dirbaio:
Equivalent of #331 for GPIO traits.

This one is a bit trickier, so it maybe warrants some discussion (this is the reason I've opened everything as separate PR's, sorry for the spam!):

- The bounds on `IoPin` are becoming very cursed, but I *think* they're correct... Is there some HAL out there implementing it, so that I can verify this doesn't break it?
- This forces the "input" and "output" errors to be the same. Perhaps we want to only unify Output and Input separately? In practice I don't think this will be an issue, as structs usually impl only Input or Output but not both (except with IoPin maybe?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Fixed blanket impl of `DelayUs` not covering the `delay_ms` method.
 
+### Changed
+- `digital`: traits now enforce all impls on the same struct have the same `Error` type.
+
 ## [v1.0.0-alpha.6] - 2021-11-19
 
 *** This is (also) an alpha release with breaking changes (sorry) ***
--- a/src/digital.rs
+++ b/src/digital.rs
@@ -2,6 +2,21 @@
 
 use core::{convert::From, ops::Not};
 
+/// GPIO error type trait
+///
+/// This just defines the error type, to be used by the other traits.
+pub trait ErrorType {
+    /// Error type
+    type Error: core::fmt::Debug;
+}
+
+impl<T: ErrorType> ErrorType for &T {
+    type Error = T::Error;
+}
+impl<T: ErrorType> ErrorType for &mut T {
+    type Error = T::Error;
+}
+
 /// Digital output pin state
 ///
 /// Conversion from `bool` and logical negation are also implemented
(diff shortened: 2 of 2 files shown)


burrbull:
> * The bounds on `IoPin` are becoming very cursed, but I _think_ they're correct... Is there some HAL out there implementing it, so that I can verify this doesn't break it?
>

https://github.com/stm32-rs/stm32f4xx-hal/blob/dc1552b2fc2e14901203a13ab42e76a5121a08eb/src/gpio.rs#L574

Dirbaio:
Thanks for the link @burrbull . It seems the bounds are OK, the obvious changes worked with no fuss! https://github.com/Dirbaio/stm32f4xx-hal/commit/091d9450fc92e85e5ad0271db6a761ff2959bef0

eldruin (approved):
Seems correct to me, thanks!

Dirbaio:
> Breaking up the unified error type requirement for IoPin in the future would be a non-breaking change, right?

I have no idea to be honest. Removing the `ErrorType` bounds on `TInput/TOutput` seems like it could break consumers relying on it, so probably not..?

eldruin:
> > Breaking up the unified error type requirement for IoPin in the future would be a non-breaking change, right?
> 
> I have no idea to be honest. Removing the `ErrorType` bounds on `TInput/TOutput` seems like it could break consumers relying on it, so it's maybe breaking...?

Hmm, I see.
There might be an `IoPin` impl wanting to define a different error type. Something like:
```rust
enum IoPinError {
  Input(...),
  Output(...),
  FailedToChangeMode
}
```
Given this, I would tend not to unify the error type for `IoPin`.

Dirbaio:
Changed to leave `IoPin` alone.

Anyway IMO IoPin requires a redesign.
- It's incredibly unwieldy to impl #340 
- It's incredibly unwieldy to use: you can't simply own one in a driver struct, because every time you use it it changes type. Tshi forces you to store Options or enums and constantly `.take()`ing it out and putting it back, which results in panicky code with lots of runtime checks.
- There's still no way to switch between PullUp/PullDown/NoPull, and PushPull/OpenDrain

but that's a different story I guess...

eldruin:
Hmm, I still see `IoPin` inherit from `ErrorType`, though.
Could you add that criticism of `IoPin` to https://github.com/rust-embedded/embedded-hal/issues/340?
Have you read through the original discussion at https://github.com/rust-embedded/embedded-hal/issues/29?

Dirbaio:
I screwed up the rebase somehow, it's OK now

Dirbaio:
Friendly ping @eldruin @therealprof @ryankurte are there any pending concerns on this now that IoPin is left untouched?

therealprof (approved):
LGTM, thanks.

bors r+

PR #336, opened 2021-12-20
i2c: enforce all traits have the same Error type.

Dirbaio:
Equivalent of #331 for i2c.

PR #339, opened 2021-12-20
I2c unify

Dirbaio:
Depends on #336 

Equivalent of #323 but for I2c.

I think for i2c unifying everything in a single trait makes the most sense. The i2c bus is specified to be bidirectional, I believe no hardware out there can "only write" or "only read" (and writing requires *reading* ACK bits anyway!).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - `i2c`: traits now enforce all impls on the same struct have the same `Error` type.
+- `i2c`: unify all traits into a single `I2c` trait.
 
 ## [v1.0.0-alpha.6] - 2021-11-19
 
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -22,22 +22,70 @@
 //! Here is an example of an embedded-hal implementation of the `Write` trait
 //! for both modes:
 //! ```
-//! # use embedded_hal::i2c::{ErrorKind, ErrorType, SevenBitAddress, TenBitAddress, blocking::Write};
+//! # use embedded_hal::i2c::{ErrorKind, ErrorType, SevenBitAddress, TenBitAddress, blocking::{I2c, Operation}};
 //! /// I2C0 hardware peripheral which supports both 7-bit and 10-bit addressing.
 //! pub struct I2c0;
 //!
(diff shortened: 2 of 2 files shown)


eldruin (commented):
Seems like a good idea, although the trait has a lot of methods now. 
Are we sure it is always possible to implement the iterator methods?
Also, for consistency, if we rename the `exec` method in `SPI` `batch` we should do so here as well.

Dirbaio:
> Are we sure it is always possible to implement the iterator methods?

For stm32, rp2040 yes. 

For nRF it'd need to drain the iterator into a temp buffer then DMA that out. However there are other cases that nRF has to do similar copies to temp bufs (for example, it can't DMA out of flash, so if TXing from flash it has to be copied to RAM) so I don't think it's that bad.

If the iterator methods were separate traits, we'd have the usual problem: some HALs wouldn't impl them, so driver authors wouldn't use them since they want their driver to work everywhere. IMO they should be either in the main trait, or removed.

> Also, for consistency, if we rename the exec method in SPI batch we should do so here as well.

Agreed they should be consistent! I'm not 100% happy with the `batch` name though. I liked "Transactional", it emphasizes it's a single transaction, not equivalent to just doing multiple read/write calls. Perhaps `transaction`? `exec_transaction`?

burrbull:
> For stm32

Yes for writing. Probably not for reading.

For correct ending of read operation we need to know length of remaining piece.

https://github.com/stm32-rs/stm32f1xx-hal/blob/e790b27663c03362b7dfac924f55b02d00869c08/src/i2c/blocking.rs#L291

Dirbaio:
There's no iterator-based reading.

burrbull:
> 

For this reason it is not there. Same as Transactional.

Dirbaio:
@burrbull not sure if I understand. The trait proposed in this PR has a write_iter, but no read_iter, so AFAIK all the methods should be implementable for stm32s. Are you saying a method is not implementable, or just "if there was a read_iter method, it wouldn't be"?

burrbull:
> The trait proposed in this PR has a write_iter, but no read_iter

Oh, sorry. I was sure I'd seen it.

Dirbaio:
Renamed `exec` to `transaction`, naming is now consistent with #323

eldruin (commented):
Great work, thank you!
bors r+

eldruin (approved):
bors r+

PR #342, opened 2021-12-21
Add embedded-hal-async crate in subdirectory

eldruin:
This is a simpler version of #334 where an empty `embedded-hal-async` crate is added inside a subfolder.
I propose merging this and add all the traits and docs piece by piece so that they can be discussed comfortably.
The CI adaption went fine except for the changelog check, which cannot check that the changelog in the `embedded-hal-async` subfolder was touched.

All credit for this idea goes to @Dirbaio !

### DIFF
--- a/.github/workflows/ci-async.yml
+++ b/.github/workflows/ci-async.yml
@@ -1,0 +1,32 @@
+on:
+  push:
+    branches: [ staging, trying, master ]
+  pull_request:
+
+name: CI Async
+
+env:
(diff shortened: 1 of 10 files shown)


Dirbaio:
:+1: :+1: 

So are we merging this? Shall I split up #334 into several PRs on top of this?

burrbull:
> So are we merging this? Shall I split up #334 into several PRs on top of this?

This is the plan

ryankurte (approved):
lgtm!

bors r+

PR #344, opened 2022-01-11
async: add DelayUs

Dirbaio:
Add async DelayUs, mirroring the blocking one.

Unlike blocking, `delay_ms` has no default impl because it's not possible to have default impls in GAT-based async traits.

### DIFF
--- a/.github/workflows/clippy-async.yml
+++ b/.github/workflows/clippy-async.yml
@@ -5,16 +5,15 @@ on:
 
 name: Clippy check
 jobs:
-  clippy_check:
+  clippy_check-async:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
           override: true
           components: clippy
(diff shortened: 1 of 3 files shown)


Dirbaio:
How do we handle the changelog check for the async crate? No need to changelog anything until first release, I guess.

eldruin (commented):
Seems fine to me, thanks!
Do you really need two different futures for the `us` and `ms` variants?

The changelog check does not seem to support more than one changelog file in a repository so I think we can just ignore it for e-h-a changes. It is not required for a bors merge, just slightly annoying for maintainers.

Dirbaio:
> Do you really need two different futures for the us and ms variants?

an impl may want to have a loop for the `ms` variant like in the blocking default impl, to avoid overflow. In that case they'd be different futures.

Also, for consistency I'd rather always do "1 method = 1 future associated type". This is what the future "async fn in traits" will desugar to. Also, users will almost never use the associated types, they'll always do `foo.method().await`, so having many of them is not bad for usability. This gives max flexibility to implementors with no downside.

eldruin (approved):
I see, thanks for the clarification.
Looks good to me!
bors r+

PR #345, opened 2022-01-12
Rename generic W type Word for consistency

### LINKED ISSUE #343: Generic param naming consistency: `Word` vs `W`.
- Serial: `Word`
- ADC: `Word`
- SPI: `W`

Which one do we pick?

eldruin:
For increased readability and consistency, I think the best would be to use `Word` everywhere.
Closes #343


ryankurte (approved):
lgtm!

bors r+

PR #347, opened 2022-01-12
async: add SPI

Dirbaio:
This is an exact mirror of the blocking SPI trait (including the proposed changes in #351 ), except the following differences:

- `W: 'static` is required everywhere, otherwise complicated lifetime bounds are required in the future GATs.

### DIFF
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -9,8 +9,8 @@ use embedded_hal::{digital::blocking::OutputPin, spi::blocking};
 
 /// SPI device trait
 ///
-/// SpiDevice represents ownership over a single SPI device on a (possibly shared) bus, selected
-/// with a CS pin.
+/// `SpiDevice` represents ownership over a single SPI device on a (possibly shared) bus, selected
+/// with a CS (Chip Select) pin.
 ///
 /// See (the docs on embedded-hal)[embedded_hal::spi::blocking] for important information on SPI Bus vs Device traits.
 pub trait SpiDevice: ErrorType {
@@ -30,17 +30,18 @@ pub trait SpiDevice: ErrorType {
                 ),
             > + 'a;
 
-    /// Start a transaction against the device.
+    /// Perform a transaction against the device.
     ///
     /// - Locks the bus
     /// - Asserts the CS (Chip Select) pin.
     /// - Calls `f` with an exclusive reference to the bus, which can then be used to do transfers against the device.
+    /// - [Flushes](SpiBusFlush::flush) the bus.
     /// - Deasserts the CS pin.
-    /// - Unlocks the bus,
+    /// - Unlocks the bus.
     ///
-    /// The lock mechanism is implementation-defined. The only requirement is it must prevent two
+    /// The locking mechanism is implementation-defined. The only requirement is it must prevent two
     /// transactions from executing concurrently against the same bus. Examples of implementations are:
-    /// critical sections, blocking mutexes, async mutexes, or returning an error or panicking if the bus is already busy.
+    /// critical sections, blocking mutexes, async mutexes, returning an error or panicking if the bus is already busy.
     fn transaction<'a, R, F, Fut>(&'a mut self, f: F) -> Self::TransactionFuture<'a, R, F, Fut>
     where
         F: FnOnce(&'a mut Self::Bus) -> Fut + 'a,
@@ -74,14 +75,16 @@ impl<T: SpiDevice> SpiDevice for &mut T {
     }
 }
 
-/// Flush for SPI bus
+/// Flush support for SPI bus
 pub trait SpiBusFlush: ErrorType {
     /// Future returned by the `flush` method.
     type FlushFuture<'a>: Future<Output = Result<(), Self::Error>> + 'a
     where
         Self: 'a;
 
-    /// Flush
+    /// Wait until all operations have completed and the bus is idle.
+    ///
+    /// See (the docs on embedded-hal)[embedded_hal::spi::blocking] for information on flushing.
     fn flush<'a>(&'a mut self) -> Self::FlushFuture<'a>;
 }
 
@@ -104,6 +107,9 @@ pub trait SpiBusRead<Word: 'static + Copy = u8>: SpiBusFlush {
     ///
     /// The word value sent on MOSI during reading is implementation-defined,
     /// typically `0x00`, `0xFF`, or configurable.
+    ///
+    /// Implementations are allowed to return before the operation is
+    /// complete. See (the docs on embedded-hal)[embedded_hal::spi::blocking] for details on flushing.
     fn read<'a>(&'a mut self, words: &'a mut [Word]) -> Self::ReadFuture<'a>;
 }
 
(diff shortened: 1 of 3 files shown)


Dirbaio:
Pushed new version

- Rebased on top of #351
- Added async SpiDevice trait, mirroring the blocking one.

Dirbaio:
New version

- Rebased on #351, #365
- Added ExclusiveDevice impl

I've had LOTS of trouble with the "async closure". The right bound would be something like `where F: for<'b> FnOnce(&'b mut Bus) -> (impl Future<Output=Result<R, Bus::Error>> + 'a)`, but it's impossible to express that in today's Rust.

The workaround is to use the outer `'a` lifetime, and have the future "give back" the bus: `where F: FnOnce(&'a mut Bus) -> Fut, Fut: Future<Output=(&'a mut Bus, Result<R, Bus::Error>)> + 'a`.

This has the nasty implication that it's no longer possible to use `?` inside the closure. Also rustc can't infer the result of `transaction()`, so you have to assign it to a temp variable. It can't infer the closure arg type either, so you have to write `|bus: &mut T::Bus| `...

It's not possible to add the helper default methods to `SpiDevice` to hide this for single-transfer transactions either.

blocking:
```rust
    let mut buf = [0;64];
    device.transaction(|bus: &mut T::Bus| {
        bus.write(&[0x42])?;
        bus.read(&mut buf)?;
        Ok(())
    }).unwrap()
```

async:
```rust
    let mut buf = [0;64];
    let res: Result<(), T::Error> = device.transaction(|bus: &mut T::Bus| async {
        if let Err(e) = bus.write(&[0x42]).await {
            return (bus, Err(e));
        }
        if let Err(e) = bus.read(&mut buf).await {
            return (bus, Err(e));
        }
        (bus, Ok(())
    }).await;
    res.unwrap()
```

Not pretty, but it works. I'm afraid it's the best we can do with today's async traits.

Updated embassy [here](https://github.com/embassy-rs/embassy/pull/627)
Example of a dumb driver [here](https://gist.github.com/Dirbaio/49d1cd45c7fa838f200a5cb451bbb69e)

embedded-hal-async/src/spi.rs
```diff
@@ -0,0 +1,340 @@
...
+    BUS: blocking::SpiBusFlush,
+    CS: OutputPin,
+{
+    type Bus = BUS;
+
+    fn transaction<R>(
+        &mut self,
+        f: impl FnOnce(&mut Self::Bus) -> Result<R, <Self::Bus as ErrorType>::Error>,
+    ) -> Result<R, Self::Error> {
+        self.cs.set_low().map_err(ExclusiveDeviceError::Cs)?;
```
ryankurte, line 277: should these be async pins too?

ryankurte (commented):
ohh wow that gets gnarley with the bus callbacks huh. i'm a little cautious that folks are going to have bad experiences trying to use it but, seems well documented, and given it is -for now- it doesn't seem like there's a better option (and your examples look great!).

in my cursed async experience sometimes things can be simplified by concretising async return types, and using `&'static dyn` dispatch, however, only sometimes and i don't really have the time to bash my head on it at the moment...

Dirbaio:
You can do it with `Box<dyn Future<..>>`, but that requires alloc. `&'static dyn Future` won't work without horrible hacks, as the impl has to allocate the future somewhere... Also `dyn` itself should be avoided, it causes some bloat with the vtable and virtual calls, and prevents optimizations.

Also, in the long-term, first-class support for [async closures](https://rust-lang.github.io/async-fundamentals-initiative/roadmap/async_closures.html) will make this even more ergonomic.

I've been looking deeper at this. It turns out it is possible to express `where F: for<'b> FnOnce(&'b mut Bus) -> (impl Future<Output=Result<R, Bus::Error>> + 'a)`, by desugaring it manually like this: [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=50dfee178d6d6abbca31c907b390b3c5)

```rust
where 
  for<'b> F: FnOnce<(&'b mut Bus,)>,
  for<'b> <F as FnOnce<(&'b mut Bus,)>>::Output: Future<Output = Result<R, Bus::Error>> + 'b,
```

but there's a few issues:
- Rust's current trait solver isn't smart enough to process these bounds. It kinda works with `-Zchalk` (if you change it to not use futures, as Chalk can't handle generators right now..)
- It works with a standalone `async fn` but not with an `|| async { ... }` closure, because Rust is not inferring it to be higher-ranked. https://github.com/rust-lang/rust/issues/70263

It's likely this will improve in the future, we can change the trait to take advantage of whatever solution comes up later...

Dirbaio:
Rebased on latest master.

Update on the "Rust's current trait solver isn't smart enough" issue above: By reducing the `where` clauses from the post above I've arrived to [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0dfa5101035db25314774cde5f154570)

Seems the issue is bounds on higher-ranked associated types don't work in general. There's quite a few issues open, closest is maybe https://github.com/rust-lang/rust/issues/23481 . Seems it's blocked on "lazy normalization": https://github.com/rust-lang/rust/issues/60471 . These bounds all work fine with Chalk, so maybe they just plan to just wait for Chalk...? Either way, this is going to take quite a while... 

I think it's best to take the unfortunate usability hit for now, and not block this. When either Chalk, lazy normalization, or first-class async closures land, we'll be able to remove the hack (with a breaking change, of course...).

embedded-hal-async/src/spi.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// - Locks the bus
+    /// - Asserts the CS (Chip Select) pin.
+    /// - Calls `f` with an exclusive reference to the bus, which can then be used to do transfers against the device.
+    /// - Deasserts the CS pin.
+    /// - Unlocks the bus,
+    ///
+    /// The lock mechanism is implementation-defined. The only requirement is it must prevent two
+    /// transactions from executing concurrently against the same bus. Examples of implementations are:
+    /// critical sections, blocking mutexes, async mutexes, or returning an error or panicking if the bus is already busy.
```
eldruin, line 43:
```suggestion
    /// critical sections, blocking mutexes, async mutexes, returning an error or panicking if the bus is already busy.
```

eldruin (dismissed):
Thank you! The code is difficult to grasp, though.
I just added a few nitpicks but it looks good to me.

embedded-hal-async/src/spi.rs
ryankurte: yeah it's probably not critical, it just occurred to me that it is possible to have async pins too. later date sounds good.

ryankurte:
> I think it's best to take the unfortunate usability hit for now, and not block this. When either Chalk, lazy normalization, or first-class async closures land, we'll be able to remove the hack (with a breaking change, of course...).

hugely appreciate your effort in exploring these options @Dirbaio! :+1: to landing this once those small changes are applied.

Dirbaio:
New version

- Rebased on #371 
- Doc fixes from review comments. Since most of them applied to both blocking+async, I've fixed them in both.

eldruin (dismissed):
Looks good to me, thanks! Let's merge #371 and then this one

eldruin (approved):
Looks good to me, thanks!
bors r+

PR #348, opened 2022-01-12
async: add i2c

Dirbaio:
This mirrors the blocking i2c trait (including the changes proposed in #339 ), except the following:

- Removed all `iter` methods since they don't play well with DMA.


eldruin (approved):
Seems straight-forward to me, thanks!
We could add  an `I2cIter` trait in the future if there is a good use case.

bors r+

eldruin:
What? I just approved this

eldruin (approved):
bors r+

Issue #299, opened 2021-07-20
RFC: Use the SPI traits to represent a single peripheral on an SPI bus


GrantM11235:
# What?

Currently, the SPI traits are used to represent an entire SPI bus. I propose using them to represent a single peripheral on an SPI bus.

# Why?

Generic device drivers want to be able to communicate with an SPI device or devices, they don't need access to the whole bus. In fact, giving the entire bus to a driver has a number of downsides:
- The driver must manually toggle it's chip select. This is inconvenient and can lead to problems like #264.
- Because the driver must hold the SPI bus struct (or at least a `&mut`), nothing else can use that bus for the lifetime of the driver. [`shared-bus`](https://crates.io/crates/shared-bus) exists to work around this problem, but it has limitations (**edit: and it is unsound** Rahix/shared-bus#23) due to the manual handling of chip selects.
- A driver that communicates with two SPI devices must decide whether to take one bus (and only work if both devices are on the same bus) or two buses (and only work if the devices are on separate buses).
- A driver has no way to communicate with two SPI devices on the same bus if they use different bus configurations (polarity, phase, speed) because there is no way to change the configuration of the bus using embedded_hal.

# How?

## embedded_hal

We don't need to actually change any code in embedded_hal, we just need to mandate that any implementations of the spi traits must hold their chip select pins low for the duration of the transfer.

This is similar to #180/#245, but it would be mandatory, not optional.

## Generic device drivers

Generic device drivers can be updated by simply removing any manual toggling of chip select pins.

## HALs

This is where it gets interesting...

# HAL implementations

Instead of implementing the SPI traits directly on an spi bus, HALs will instead create device structs. Here are some simplified examples:

#### A device struct which shares it's bus:
```rust
pub struct Device {
	bus: Mutex<Spi1>,
	cs_pin: OutputPin,
	config: Config,
}

impl spi::Write<u8> for Device {
	fn write(&mut self, words: &[u8]) {
		self.bus.lock(|bus| {
			bus.reconfigure(self.config);
			self.cs_pin.set_low();
			bus.write(words);
			self.cs_pin.set_high();
		});
	}
}
```

#### A device struct with exclusive access to it's bus, which doesn't need a mutex and doesn't need to reconfigure the bus every time:
```rust
pub struct ExclusiveDevice<'a> {
	bus: &'a mut Spi1,
	cs_pin: OutputPin,
}

impl<'a'> spi::Write<u8> for ExclusiveDevice<'a> {
[...]

Issue #264, opened 2021-01-15
Clarify SPI timing behaviour (currently completely broken!)


mgottschlag:
### TLDR
Currently, it is unclear when exactly the SPI transfer functions return. Some drivers assume that the functions return after the clock line has returned to idle state. This is not how some (most?) HALs implement SPI. The documentation should clarify the expected behaviour.

### The Problem
I am currently building an application with [https://github.com/astro/embedded-nrf24l01](embedded-nrf24l01) on a system with an STM32F429 clocked at 168 MHz. I clocked the SPI bus at 500 KHz -- much slower than the system clock, which to a certain degree caused the following problems -- and subsequently saw very weird behaviour. embedded-nrf24l01 would pull the CS pin high and then low again *before the SCK signal reached idle*, so the device only ever registered one single long transfer instead of multiple small transfers as required to send individual commands. All but the first command was therefore ignored or incorrectly processed by the device.

The corresponding code in embedded-nrf24l01 is as follows:

        self.csn.set_low().unwrap();
        let transfer_result = self.spi.transfer(buf).map(|_| {});
        self.csn.set_high().unwrap();

Changing the code as follows solves all problems, because then all CS changes occur while the SPI bus is idle:

        cortex_m::asm::delay(200);
        self.csn.set_low().unwrap();
        cortex_m::asm::delay(200);
        let transfer_result = self.spi.transfer(buf).map(|_| {});
        cortex_m::asm::delay(200); // <- this line is what fixes the problem, the other delays are just to make the signal look more like in the datasheet
        self.csn.set_high().unwrap();

### The Solution(s)

1. The documentation could state that the SPI functions return as soon as the last bit was read from the bus. Device drivers would then have to add delays where necessary, similar to the code above. I do not know how many devices would require this type of delay -- if many devices need it, this solution is a *huge* footgun and substantially increases complexity. I hardcoded the delay cycles in the example above, but that's only possible because I know the clock speed of the device.
2. The documentation could state that the SPI functions return as soon as the SCK signal is idle again. Probably, all HALs would have to be modified.

### Other Problems
If I understand it correctly, this bug is not completely identical to https://github.com/stm32-rs/stm32f0xx-hal/issues/130, although that ticket shows a similar problem with SPI documentation: The documentation should (or should not?) state that transfer() finishes after the last bit has been transferred, not after the bits have been shoved into a FIFO.

mgottschlag:
[...]

PR #351, opened 2022-01-20
SPI: split into device/bus, allows sharing buses with automatic CS

Dirbaio:
Because why not?

This separates the concepts of "SPI bus" and "SPI device". I got the idea of having traits for SPI devices from #299. It has an excellent breakdown of the issues of the status quo, so go read that first.

The difference from #299 is that proposes "repurposing" the existing traits to represent devices, while this PR proposes specifying the existing traits represent buses, and introduces a new trait for representing devices on top of that.

- HALs impl the bus traits, just like now. No extra boilerplate.
- Crates like `shared-bus` add the "bus mutexing" logic, to impl `SpiDevice` on top of `SpiBus` (+ `OutputPin` for CS). There's a wide variety of possible impls, all can satisfy the `SpiDevice` requirements:
   - Exclusive single-device, that just owns the entire bus
   - Single-thread, on top of `RefCell`
   - Multithreaded std, using std `Mutex`
   - Multithreaded embedded, using some `AtomicBool` taken flag.
   - Async embedded, using an async mutex (this'd require an async version of SpiDevice).
- Drivers take `T: SpiDevice` (or `SpiDeviceRead`, `SpiDeviceWrite`), then they start a `.transaction()`, then use it to read/write/transfer data. The transaction ends automatically when returning. Example:

```rust
fn read_register(&mut self, addr: u8) -> Result<u8, Error> {
    let mut buf = [0; 1];
    self.spi.transaction(|bus| {
        bus.write(&[addr])?;
        bus.read(&mut buf)?;
    })?;
    Ok(buf[0])
}
```

## No Transactional needed

Previous proposals boiled down to: for each method call (read/write/transfer), it automatically assets CS before, deasserts it after. This means two `write`s are really two transactions, which might not be what you want. To counter this, `Transactional` was added, so that you can do multiple operations in a single CS assert/deassert.

With this proposal, you have an explicit "Transaction" object, so you can easily do multiple reads/writes on it in a single transaction. This means `Transactional` is no longer needed.

The result is more flexible than `Transactional`, even. Take an imaginary device that responds to "read" operations with a status code, and the data only arrives on success. With this proposal you can easily read data then take conditional action.  With `Transactional` there's no way, best you can do is read the data unconditionally, then discard it on error.

It allows for smaller memory usage as well. With `Transactional` you had to allocate buffer space for all operations upfront, with this you can do them as you go.

```rust
fn read_data(&mut self, addr: u8, dest: &mut [u8]) -> Result<(), Error> {
    self.spi.transaction(|bus| {
        bus.write(&[addr])?;

        // Read status code, 0x00 indicates success
        let mut buf = [0; 1];
        bus.read(&mut buf)?;
        if buf[0] != 0x00 {
            return Err(Error::ErrorCode(buf[0]));
        }

        // If the operation is successful, actually read the data, still in the same transaction
        bus.read(dest)
    })
}
```


ryankurte:
hey another interesting option!

> Because why not?

🤣

> Drivers take T: SpiDevice (or SpiReadonlyDevice, SpiWriteonlyDevice), then they start a .transaction(), then use it to read/write/transfer data. The transaction ends automatically when dropped. 

i guess you'd also have a `fn done(mut self)` to make it obvious how to explicitly close the transaction, or you'd have to document `.drop()`.

> With this proposal, you have an explicit "Transaction" object, so you can easily do multiple reads/writes on it in a single transaction. This means Transactional is no longer needed.

i believe this is achieved by both ManagedCS proposals, though it is still useful to have a mechanism for representing an arbitrary sequence of operations.

> An alternative would be using a closure like in #350, but I think the "end transaction on drop" proposed here is better:

with this approach the CS is effectively scoped to the parent function rather than the transaction and requires explicit overriding with manual `drop` to maintain ordering, which would seem to make the API easier to misuse / opens opportunities for bugs like not closing and re-opening the transaction or not closing the transaction prior to moving on.

```rust
/// Code to do some SPI stuff then wait for 100ms, note that the missing `drop` means the delay will occur within the CS assertion
fn something(&mut self) {
  let t = spi.transaction();
  // do some SPI stuff

  delay_ms(100);
}
```

i prefer the closure because it make it obvious when the end of the transaction is, rather than whenever the compiler chooses to drop the object. the timing of when a mutex is dropped  / reopened doesn't typically matter because it's just a software concept, but CS assertion and deassertion very much does.

eldruin (commented):
Nice work! this looks pretty cool.
A few questions:
- While I like the idea of closing the transaction when falling out of scope, do you have any plan on how to inform the driver/what to do about a CS pin deasertion error on `drop()` (if not using a `done(mut self)` method like @ryankurte suggested)?
- Any reason why a similar approach would not work for `I2C`?

I see the point of people leaving the transaction open while doing something else, but given the popularity of this way of holding onto mutexes with guards and the fact that the SPI bus cannot be used for anything else as long as the transaction is not completed, it would be fine for me. The peripheral would just be listening to silence for longer than necessary. Attempting to create a new transaction should result in a compilation error.

Some TODOs:
- Ensure that implementations for `&mut T` do not introduce problems like those highlighted in #350 
- Ensure it is not possible to call `transaction` twice before the first one is dropped.
- For async, ensure it is not possible to call `transaction` again while a transaction is still ongoing but an operation is being awaited.
- Since it is a whole new concept it would be especially important to provide PoC implementations for HAL, drivers and bus mutexing.
- Wait for stable GATs :coffee:

Dirbaio:
> with this approach the CS is effectively scoped to the parent function rather than the transaction and requires explicit overriding with manual drop to maintain ordering, which would seem to make the API easier to misuse / opens opportunities for bugs like not closing and re-opening the transaction or not closing the transaction prior to moving on.

They're scoped to the parent block, not function. You can also add explicit `{}` braces to force dropping earlier, instead of `drop(x)`. Also, many times the block scope naturally aligns with what you want to do, for example drivers often have `read_reg`, `write_reg` fns that do just that, so the transaction gets dropped at the right place.

> While I like the idea of closing the transaction when falling out of scope, do you have any plan on how to inform the driver/what to do about a CS pin deasertion error on drop() (if not using a done(mut self) method like @ryankurte suggested)?

Oh, this is a problem indeed. `done` would work, but there's no way to enforce the user calls it, so you still have to do something on drop, probably panicking. Not great. Maybe it's not that bad because fallible GPIO is very uncommon...? Still, I don't like that there are two ways to accomplish the same thing (`done` and dropping) with non-obvious pros and cons.

Perhaps the closure is better after all. The analogy with Mutex/RefCell breaks down because dropping Transaction causes I/O side effects which are fallible, while dropping Mutex/RefCell guards doesn't...

> Any reason why a similar approach would not work for I2C?

It'd work too, I guess? Same for the closure version. i2c has the same problems outlined in #299 so the device/bus split would help a lot there too. Same for the closure version if we decide to go with that. It'd be nice if i2c+spi transactions were consistent. i2c is a bit more picky with the repeated-start stuff etc, that might make it more complicated.

> Ensure that implementations for &mut T do not introduce problems like those highlighted in #350

I would *not* do `impl SpiBus for T where T: SpiDevice` impls (which is the logical equivalent of the ones conflicting in #350). If you have an `SpiDevice`, you *have* to start a transaction to use it. Reasons:

- It's conceptually wrong: Devices are not buses. If they were, you could give a device to a bus mutexer as if it was a bus, and have it mutexify *that*. You end up with two layers of mutexing, with the "inner" CS being auto-toggled at each read/write op, which makes no sense whatsoever.
- It's relatively common to do multiple ops in the same transaction. Most devices are some form of "Write addr + read data, write addr+write data". Having it autostart the transaction if you want to do a single write is not that valuable IMO.
- A driver might bound on SpiBus and do their own manual CS management. By doing so, it declares it wants exclusive ownership of the whole bus. The driver knows its manual CS management is correct thanks to owning the entire bus. If we had these impls, you'd be able to feed it an SpiDevice on a shared bus, which could cause transactions from other drivers to get interleaved in between without the driver knowing. This is the issue at https://github.com/Rahix/shared-bus/issues/23 .
- The impls do conflict as well, hehe :see_no_evil: 

> - Ensure it is not possible to call transaction twice before the first one is dropped.
> - For async, ensure it is not possible to call transaction again while a transaction is still ongoing but an operation is being awaited.

These are already enforced by the `'a` lifetime in `Transaction`: it borrows `self`, so you get a borrow checker error if you try.

> - Wait for stable GATs :coffee:

Yep... :sleeping:

Dirbaio:
Pushed a new version!

- Changed to use closures.
   - Removes the requirement on nightly-only GATs.
   - Solves the issue of propagating errors when deasserting CS.
- Removed `ManagedCs` trait. After thinking about it more, I think the "device" abstraction is enough, and having an additional `ManagedCs` trait would be counterproductive / too restrictive.
   - For example: a driver for a chip with CS will require ManagedCs. However, the user might want to tie CS low on the PCB if they're not planning on sharing the bus. The CS pin is there, just not accessible from the MCU. The user would be unable to use the driver, unless doing hacks like "fake" ManagedCs impls that don't actually manage CS.
- Changed `SpiDevice`, `SpiDeviceRead`, `SpiDeviceWrite` from traits to trait aliases. The problem with them being traits is the Device impls must explicitly impl them. We don't want that, we want them to be just a "pre-made set of trait bounds", which is exactly what trait aliases are. They're nightly-only, so perhaps it's best to not include them for now, and have drivers add bounds like this instead: `where T: SpiDeviceBase, T::Bus: SpiBus`.

I also wrote a working end-to-end test case [here](https://github.com/embassy-rs/embassy/blob/bae26c46edbb3f439cd5deb8b3d04c41403fdd5f/examples/rp/src/bin/spi_display.rs). Targets the Raspberry Pi Pico RP2040 with the [Pico-ResTouch-LCD-2.8](https://www.waveshare.com/wiki/Pico-ResTouch-LCD-2.8). It's the perfect use case because it has an SPI display and an SPI resistive touch sensor on the same SPI bus with separate CS pins.

- The [HAL impl](https://github.com/embassy-rs/embassy/blob/bae26c46edbb3f439cd5deb8b3d04c41403fdd5f/embassy-rp/src/spi.rs#L335-L365) stays nice and small, it only cares about the Bus part. It doesn't have to worry about Device stuff.
- [mod shared_spi](https://github.com/embassy-rs/embassy/blob/bae26c46edbb3f439cd5deb8b3d04c41403fdd5f/examples/rp/src/bin/spi_display.rs#L110-L174) impls SpiDevice on top of SpiBus+OutputPin, in a HAL-independent way. This is something that could go in crates like `shared_bus`.
- [mod touch](https://github.com/embassy-rs/embassy/blob/bae26c46edbb3f439cd5deb8b3d04c41403fdd5f/examples/rp/src/bin/spi_display.rs#L177-L233) implements the touch driver, generic on a SpiDevice. This shows how a small simple driver would look like. The code is simplified thanks to Device dealing with CS for us. It also shows [doing multiple transfers in the same transaction](https://github.com/embassy-rs/embassy/blob/bae26c46edbb3f439cd5deb8b3d04c41403fdd5f/examples/rp/src/bin/spi_display.rs#L212-L215) with the new API (this would be done with the Transactional trait before, which this PR removes because it's no longer needed.)
- [mod my_display_interface](https://github.com/embassy-rs/embassy/blob/bae26c46edbb3f439cd5deb8b3d04c41403fdd5f/examples/rp/src/bin/spi_display.rs#L235) contains the display interface impl, generic on a SpiDeviceWrite. Of note, this is a use case of the write-only SPI traits.

It all seems to work nicely together! :D

Open questions:
- Do we want to add a "CS" variant to ErrorKind?
- Do we want to add pre-made SpiDevice impls? Probably yes, at least for the "exclusive" case (no sharing), with and without CS.

ryankurte:
approach looks pretty good to me! though still a couple of questions...

i don't think the `device` vs `bus` nomenclature is super clear, i think the biggest complaint is that we're going to end up with situations where we're talking about (a driver for) _an spi device_ (eg. a radio) with bounds on `SpiDevice`.

i think we can resolve this with better module and per-function documentation wrt. the `SpiDeviceX` and `SpiBusX` functions to describe their purpose / use, and i'd be in favour of `s/SpiDevice/Spi/g` in the trait names since that's what _most_ people will see.

> Changed SpiDevice, SpiDeviceRead, SpiDeviceWrite from traits to trait aliases. The problem with them being traits is the Device impls must explicitly impl them.

agreed that (for now at least) we're going to need to drop the trait aliases... we _could_ replicate this behaviour with marker traits and bounded default implementations, though they conflict with `&mut T` impls and i'm not sure whether it would be breaking to swap to aliases later. again i think documentation is probably the solution (show people exactly how to use em).

> For example: a driver for a chip with CS will require ManagedCs. However, the user might want to tie CS low on the PCB if they're not planning on sharing the bus. The CS pin is there, just not accessible from the MCU. The user would be unable to use the driver, unless doing hacks like "fake" ManagedCs impls that don't actually manage CS.

i am pretty sure this is still necessary. the idea of `ManagedCS` is for drivers (hal consumers) to signal that they expect CS to be managed at a level below them, the only way to write that's abstract over software or hardware CS is to push that requirement out of the driver.

Dirbaio:
> i think we can resolve this with better module and per-function documentation wrt. the SpiDeviceX and SpiBusX functions to describe their purpose / use

Agreed, will add more docs.

> and i'd be in favour of s/SpiDevice/Spi/g in the trait names since that's more consistent with other e-h traits and what most people will see.

I think "SpiDevice, SpiBus" is the clearest. At least clearer than "Spi, SpiBus". What's "a SPI"?

If we do the Device/Bus thing for i2c as well, then there's no more consistency issues: the "bus" concept only applies to i2c and spi. 

I don't agree SpiDevice is what most people will use:
- HALs will impl SpiBus, and rarely touch SpiDevice. 
- There's still use cases for drivers using SpiBus (see below).

> agreed that (for now at least) we're going to need to drop the trait aliases... we could replicate this behaviour with marker traits and bounded default implementations, though they conflict with &mut T impls and i'm not sure whether it would be breaking to swap to aliases later. again i think documentation is probably the solution (show people exactly how to use em).

Blanket impls don't work, yeah... I'll remove them, then add more docs.

> i am pretty sure this is still necessary. the idea of ManagedCS is for drivers (hal consumers) to signal that they expect CS to be managed at a level below them, the only way to write that's abstract over software or hardware CS is to push that requirement out of the driver.

I would argue `SpiDevice` already very strongly implies there is a CS pin. A `SpiDevice` impl on a shared bus without a CS pin is broken. Perhaps we can simply document that SpiDevice MUST have a CS pin? For edge cases like "there's a CS but it's tied down in the PCB" the user can impl their own as an escape hatch, I don't think it's that common..?

Drivers that want an SpiDevice intentionally without CS shouldn't be a thing:
- Drivers that manually manage CS should use SpiBus instead of SpiDevice. Manually managing CS on a shared bus is impossible to do correctly.
- Drivers that want to abuse SPI for "cursed" stuff like WS2812B waveform generation, logic analyzer, weird bitbangings should use SpiBus as well.

ryankurte:
> I don't agree SpiDevice is what most people will use:

sure, but the number of HALs << the number of drivers << the number of applications. HAL authors are already deep in the technical side, and i would expect orders of magnitude more users at the application end, so it seems worth weighting our solutions towards that end.

> I would argue SpiDevice already very strongly implies there is a CS pin. A SpiDevice impl on a shared bus without a CS pin is broken. Perhaps we can simply document that SpiDevice MUST have a CS pin? For edge cases like "there's a CS but it's tied down in the PCB" the user can impl their own as an escape hatch, I don't think it's that common..?

i think we are looking at this from the opposite direction. so your expectation is that HALs will provide `Spi::with_sw_cs` and `Spi::with_hw_cs(..., pin: Pin)` functions so they _cannot_ be constructed without some CS control?

because in many cases you will choose whether to use the hw implementation based on the functionality you require, so can't be _required_ to use the hw approach. and building on that what happens if you had a cursed situation like CS via GPIO expander, how would you model this?

> Drivers that manually manage CS. Manually managing CS on a shared bus is impossible to do correctly.

this is the whole point of `SpiDevice` and `transaction`... once you're in the closure you have exclusive access to the bus to do what you will with CS (or any other IOs)?

Dirbaio:
Pushed new version

- Removed trait aliases
- Renamed `SpiDeviceBase` to `SpiDevice`
- Added `ExclusiveDevice`. Impls a `SpiDevice` out of a `SpiBus` and a `OutputPin`, with no bus sharing.
- Added docs.

@ryankurte hopefully the added docs clarifies the confusion. The TLDR is:

- `SpiBus` = exclusive ownership of the whole bus. NEVER manages CS. Implemented by HALs
- `SpiDevice` = ownership of a single device in a possibly-shared bus. ALWAYS manages CS. Implemented by either:
    - Using Hardware CS, in the HALs.
    - Using software CS, in either
         - `embedded-hal`, the `ExclusiveDevice`
         - Third-party crates, like `shared-bus`.

Dirbaio:
ugh, MSRV is 1.46+ but `#[doc = include_str!(...)]` is 1.54+ ... What do we do?

ryankurte:
> Since SpiDevice ALWAYS manages CS, there's no need for a ManagedCs marker trait. There's no such thing as a "SpiDevice without CS", you just use SpiBus instead.

alright i think that's clearer. and if your hardware does CS but you don't use it you just ignore it?

> ugh, MSRV is 1.46+ but #[doc = include_str!(...)] is 1.54+ ... What do we do?

would it work to chuck the images here then link to them instead of embedding for now?

Dirbaio:
> alright i think that's clearer. and if your hardware does CS but you don't use it you just ignore it?

Yep, The HAL would only enable hardware-CS if you ask for it. You can just use the SpiBus part.

> would it work to chuck the images here then link to them instead of embedding for now?

:+1:

Dirbaio:
Pushed new version:

- Link image from github (note: it'll be broken until merged because it points to git master... maybe there's a better way?)
- Added changelog.

ryankurte (approved):
lgtm, thanks for bearing with me ^_^

@rust-embedded/hal all happy?

eldruin (commented):
This looks great, thank you!
I just added a couple nitpicks.

re. CS error kind, since we have `ErrorKind::ModeFault` (Multiple devices on the SPI bus are trying to drive the slave select pin, e.g. in a multi-master setup) and the CS pin management is so common, I think we should add some kind of CSFault variant.

A thing to clarify would be the timing behavior (#264) as we discussed. Somehow `SPIDevice` implementations must be able to ensure that CS is only deasserted after SCK is idle.

re. MSRV, I think it would make sense to top up the MSRV before the 1.0.0 release so we can have as many goodies as possible, I would say feel free to do so and include the svg in the docs properly, so that docs are self-contained.

We should think about which `SPIDevice` implementations we should include here, but I think we are fine with `ExclusiveDevice` for now. Maybe we can add a `RefCell` variant as well, but we should discuss this in a separate issue.

I look forward to using and implementing these traits :)

Dirbaio:
Pushed new version:
- Applied suggestions from review
- Added `ErrorKind::ChipSelectFault`
- Bumped MSRV to 1.54
- Changed the svg back to `include_str!`

> A thing to clarify would be the timing behavior (https://github.com/rust-embedded/embedded-hal/issues/264) as we discussed. Somehow SPIDevice implementations must be able to ensure that CS is only deasserted after SCK is idle.

Agreed this is needed for correct SpiDevice operation. I was planning on sending a follow-up PR, as this PR is quite big already. (probably adding a `.flush()` method, which AFAICT should work fine with the SpiDevice design)

eldruin (approved):
Great, thank you!
@ryankurte already approved this.
@therealprof what do you think?

ryankurte (approved):

src/spi/blocking.rs
```diff
@@ -1,108 +1,287 @@
...
+//!         Self { spi }
+//!     }
+//!
+//!     pub fn read_foo(&mut self) -> [u8; 2] {
+//!         let mut buf = [0; 2];
+//!
+//!         // `transaction` asserts and deasserts CS for us. No need to do it manually!
+//!         self.spi.transaction(|bus| {
+//!             bus.write(&[0x90]).unwrap();
+//!             bus.read(&mut buf).unwrap();
```
eldruin, line 82:
```suggestion
//!             bus.write(&[0x90])?;
//!             bus.read(&mut buf)
```
Can we rather do this in this and the rest of examples?

Dirbaio:
Pushed new version.

- Changed examples to use error handling best practices (error enum instead of unwrap).
- Changed transaction closure from `FnOnce(&mut Self::Bus) -> R` to `FnOnce(&mut Self::Bus) -> Result<R, <Self::Bus as ErrorType>::Error>`.

The reason for the closure change is that `SpiDevice::Error` and `SpiDevice::Bus::Error` are different types (and they can't be the same type because SpiDevice has to be able to add the CS errors). This forces drivers to deal with both errors, which is quite ugly:

```rust
enum Enc28j60Error<SPI>
where
    SPI: SpiDevice,
    SPI::Bus: SpiBus,
{
    SpiDeviceError(SPI::Error),
    SpiBusError(<SPI::Bus as ErrorType>::Error),
}

struct Enc28j60<SPI> {
    spi: SPI,
}

impl<SPI> Enc28j60<SPI>
where
    SPI: SpiDevice,
    SPI::Bus: SpiBus,
{
    pub fn do_it(&mut self) -> Result<(), Enc28j60Error<SPI>> {
        self.spi
            .transaction(|bus| bus.write(&[0xFF]))
            .map_err(Enc28j60Error::SpiDeviceError)?
            .map_err(Enc28j60Error::SpiBusError)?;
        Ok(())
    }
}
```

WIth the new API, the SpiDevice impl converts the `SpiDevice::Bus::Error` into a `SpiDevice::Error`, so drivers only have to deal with the `SpiDevice::Error`. The result looks like this, which is much cleaner IMO.

```rust

enum Enc28j60Error<SPI> {
    Spi(SPI),
}

struct Enc28j60<SPI> {
    spi: SPI,
}

impl<SPI> Enc28j60<SPI>
where
    SPI: SpiDevice,
    SPI::Bus: SpiBus,
{
    pub fn do_it(&mut self) -> Result<(), Enc28j60Error<SPI::Error>> {
        self.spi.transaction(|bus| bus.write(&[0xFF])).map_err(Enc28j60Error::Spi)?;
        Ok(())
    }
}
```

The downside is the new way requires nested Results if the user wants to return their own errors from within the transaction. However, the previous way would *always* return a nested error from `transaction` in practice ALL the time, so I think it's a net win.

src/spi/blocking.rs
Dirbaio: Done

Dirbaio:
New version
- Add changelog entries to Unreleased instead of to alpha.7 (oops!)

src/spi/blocking.rs
```diff
@@ -1,108 +1,315 @@
...
+//! SPI bus sharing doesn't make sense at all in this case. By requiring [`SpiBus`] you disallow sharing, ensuring correct operation.
+//!
+//! # For HAL authors
+//!
+//! HALs **must** implement [`SpiBus`], [`SpiBusRead`] and [`SpiBusWrite`]. Users can combine the bus together with the CS pin (which should
+//! implement [`OutputPin`]) using HAL-independent [`SpiDevice`] implementations such as [`ExclusiveDevice`].
+//!
+//! HALs may additionally implement [`SpiDevice`] to **take advantage of hardware CS management**, which may provide some performance
+//! benefits. (There's no point in a HAL implementing [`SpiDevice`] if the CS management is software-only, this task is better left to
+//! the HAL-independent implementations).
```
Rahix, line 139: I do not think HALs should ever implement `SpiDevice`.  The reason is that a HAL has no idea about the execution environment.  But for bus sharing, it needs to make assumptions about some mutex primitive to use.  This will quickly lead to compatibility problems between HALs and "OS" crates.  I'll leave a comment below to detail how I would approach hardware CS support instead (that works with this PR's design as it is!).

Rahix (commented):
Hi, sorry for chiming in so late.  I have to say, I _really_ like this design.  It solves so many of the issues we've been seeing.  And it is very elegant in doing that :)

Rahix:
As mentioned in an inline comment above, I do not think the bus-sharing code for hardware chip-select support should be implemented in the HALs.  I fear this will quickly lead to compatibility problems because bus-sharing always requires some sort of synchronization primitive and a HAL has no idea what synchronization methods an application would prefer to use.

I've been thinking about this whole story for quite some time and the most promising idea I came up with is this:  HALs should provide concrete APIs on their `SpiBus`-implementing type for configuring the active hardware-CS line.  The bus-sharing solution (for example, ugh, `shared-bus`) then provides a way to call a small function after locking but before operating on the bus.  This function is user-provided and configures the bus for the impending operation.  To let some code speak:

```rust
// this is what application code would look like:
// ---------------------------------------------
let spi_bus = BusManager::new(SomeSpiBus::new());

let dev_a: impl SpiDevice = spi_bus.acquire_with_hwcs(|b| {
    // set_active_cs() just marks which CS pin should be asserted when a transfer is started
    b.set_active_cs(1);
});
let dev_b: impl SpiDevice = spi_bus.acquire_with_hwcs(|b| {
    b.set_active_cs(2);
});

// acquire_with_hwcs() is defined like this:
// ----------------------------------------
impl<BUS> BusManager<BUS> {
    fn acquire_with_hwcs<'a>(&'a self, cfg: fn(BUS)) -> SharedSpiDevice<'a, BUS, NoCs> { ... }

    // ofc. there is a variant with GPIO chip-select
    fn acquire<'a, CS: OutputPin>(
        &'a self, cs: CS, cfg: fn(BUS)
    ) -> SharedSpiDevice<'a, BUS, CS> { ... }
}

// and the function is called like this:
// ------------------------------------
impl<'a, BUS> SpiDevice<BUS> for SharedSpiDevice<'a, BUS> {
    fn transaction<R>(&mut self, f: ...) -> Result<R, Self::Error> {
        self.bus.lock(|bus: &mut BUS| {
            // do note that CS can be a no-op implementation for acquire_with_hwcs()
            self.cs.set_low().map_err(ExclusiveDeviceError::Cs)?;

            //     |||
            //     VVV
            self.cfg(bus);

            let f_res = f(bus);
            let cs_res = self.cs.set_high();
            let f_res = f_res.map_err(ExclusiveDeviceError::Spi)?;
            cs_res.map_err(ExclusiveDeviceError::Cs)?;

            Ok(f_res)
        })
    }
}
```

---

This design pattern additionally provides a solution to a problem which was not discussed here at all so far:  How do you talk to multiple devices which need different bus configurations (SPI mode, clock speed)?  With this design it is simple:

```rust
// application code:
let spi_bus = BusManager::new(SomeSpiBus::new());

let cs_flash = gpio.P42.into_output();
let flash = SpiFlashDriver::new(spi_bus.acquire(cs_flash, |b| {
    b.set_speed(40.MHz());
    b.set_mode(spi::MODE_0);
}));

let cs_sensor = gpio.P23.into_output();
let sensor = SensorDriver::new(spi_bus.acquire(cs_sensor, |b| {
    b.set_speed(1.MHz());
    b.set_mode(spi::MODE_3);
}));
```

src/spi/blocking.rs
```diff
@@ -1,108 +1,315 @@
...
-    /// Write data out while reading data into the provided buffer
-    Transfer(&'a mut [Word], &'a [Word]),
-    /// Write data out while reading data into the provided buffer
-    TransferInplace(&'a mut [Word]),
+/// This is the most straightforward way of obtaining an [`SpiDevice`] from an [`SpiBus`],
+/// ideal for when no sharing is required (only one SPI device is present on the bus).
+pub struct ExclusiveDevice<BUS, CS> {
+    bus: BUS,
+    cs: CS,
 }
```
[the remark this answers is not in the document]
ryankurte: ooh, i like the idea of a `NoCS`, but i'd definitely make it opt-in rather than opt-out to avoid accidentally using this default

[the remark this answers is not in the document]
Dirbaio: The problem is there are SPI devices that do require you toggling CS for every transaction. The current docs say impls MUST toggle CS for each transaction, so drivers can rely on that. If an impl ties it low permanently, it'll break drivers for such devices.

If you know your particular device tolerates tying CS low permanently, you can do that impl yourself and it'll work, but that impl is "officially incorrect", so it's not something we should add to EH.

An alternative would be to allow these SpiDevice impls, and have a `ActuallyTogglesCsForEachTransaction` marker trait, so drivers can require that if needed. I'm not sure if it's worth it though...

src/spi/blocking.rs
```diff
@@ -1,108 +1,315 @@
...
+    /// - Deasserts the CS pin.
+    /// - Unlocks the bus.
+    ///
+    /// The lock mechanism is implementation-defined. The only requirement is it must prevent two
+    /// transactions from executing concurrently against the same bus. Examples of implementations are:
+    /// critical sections, blocking mutexes, or returning an error or panicking if the bus is already busy.
+    fn transaction<R>(
+        &mut self,
+        f: impl FnOnce(&mut Self::Bus) -> Result<R, <Self::Bus as ErrorType>::Error>,
+    ) -> Result<R, Self::Error>;
```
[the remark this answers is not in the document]
Dirbaio: They require `type Bus: SpiBus`, so then you can't use `SpiDevice` with read-only or write-only buses.

They could be done with "extension traits" that are impl'd with the right bounds on Bus, but I think that's too much complexity.

Dirbaio:
> HALs should provide concrete APIs on their SpiBus-implementing type for configuring the active hardware-CS line.

One challenge with HW CS is you can do multiple transfers in the same transaction:

```rust
  self.spi.transaction(|bus| {
        bus.write(&[addr])?;
        bus.read(&mut buf)?;
    })?;
```

So, we can't simply have HALs add a `set_active_cs`, and then have them assert/deassert CS for each read/write/transfer. That'll assert CS 2 times (for the write and for the read) in the above case, not 1 time.

The trait would have to look something like this. Implementing it at the HAL side would require fun tracking state so CS gets enabled/disabled at the right times...

```rust
trait SpiBusWithHardwareCs {
    fn set_active_cs(&mut self, cs: ???);
    fn start_transaction(&mut self);
    fn end_transaction(&mut self);
}
```

Is all this complexity really worth it? Are performance gains from hardware-CS really that big? And how common is it to care so much about performance that you REALLY want hardware-CS AND you still want to share the bus? Shared buses are bad for performance!

With the current design, HALs impl SpiBus for sharing with software-CS. They also have the option to impl SpiDevice without bus sharing with hardware CS (which I predict will be rare). If someone really REALLY wants hardware-cs AND sharing (which I predict will be ULTRA rare) then yes, they'll have to impl it all in the HAL, replicating the mutex logic for their particular environment.

I don't think HAL-independent hardware-CS bus sharing is worth the extra complexity to add support for that in embedded-hal.

Dirbaio:
The "configure bus" closure is super neat! That's a valid SpiDevice impl (allowed by the trait contract) indeed. 

I think it's best to leave the issue of configuring the bus to external crates, so we don't lock ourselves to a particular solution. One such solution is HALs providing `set_freq, set_mode` methods and `shared-bus` providing the "configure bus" closure as you say!

In the future we could add traits for `set_freq, set_mode` so `SpiDevice` impls can do that in a HAL-independent way, but I don't think it's needed now.

src/spi/blocking.rs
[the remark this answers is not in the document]
GrantM11235: This works if you add `where Self::Bus: SpiBusWrite` and `where Self::Bus: SpiBusRead` to the respective methods.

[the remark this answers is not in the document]
Dirbaio: Oooh, awesome idea! Added!

Dirbaio:
New version
- Added helper methods to SpiDevice to do transactions with single transfers

src/spi/blocking.rs
```diff
@@ -1,108 +1,363 @@
...
-    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error> {
-        T::transfer(self, read, write)
+    /// Do a write within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(|bus| bus.write(buf))`.
+    ///
+    /// See also: [`SpiDevice::transaction`], [`SpiBusWrite::write`]
+    fn write(&mut self, buf: &[u8]) -> Result<(), Self::Error>
+    where
+        Self::Bus: SpiBusWrite,
```
GrantM11235, line 183:
```suggestion
    fn write<Word>(&mut self, buf: &[Word]) -> Result<(), Self::Error>
    where
        Self::Bus: SpiBusWrite<Word>,
        Word: Copy,
```

src/spi/blocking.rs
```diff
@@ -1,108 +1,363 @@
...
-    /// `words` buffer, overwriting it.
-    fn transfer_inplace(&mut self, words: &mut [Word]) -> Result<(), Self::Error>;
+    /// Do a read within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(|bus| bus.read(buf))`.
+    ///
+    /// See also: [`SpiDevice::transaction`], [`SpiBusRead::read`]
+    fn read(&mut self, buf: &mut [u8]) -> Result<(), Self::Error>
+    where
+        Self::Bus: SpiBusRead,
```
GrantM11235, line 195:
```suggestion
    fn read<Word>(&mut self, buf: &mut [Word]) -> Result<(), Self::Error>
    where
        Self::Bus: SpiBusRead<Word>,
        Word: Copy,
```

src/spi/blocking.rs
```diff
@@ -1,108 +1,363 @@
...
+    }
+
+    /// Do a transfer within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(|bus| bus.transfer(read, write))`.
+    ///
+    /// See also: [`SpiDevice::transaction`], [`SpiBus::transfer`]
+    fn transfer(&mut self, read: &mut [u8], write: &[u8]) -> Result<(), Self::Error>
+    where
+        Self::Bus: SpiBus,
```
GrantM11235, line 207:
```suggestion
    fn transfer<Word>(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error>
    where
        Self::Bus: SpiBus<Word>,
        Word: Copy,
```

src/spi/blocking.rs
```diff
@@ -1,108 +1,363 @@
...
+    }
+
+    /// Do an in-place transfer within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(|bus| bus.transfer_in_place(buf))`.
+    ///
+    /// See also: [`SpiDevice::transaction`], [`SpiBus::transfer_in_place`]
+    fn transfer_in_place(&mut self, buf: &mut [u8]) -> Result<(), Self::Error>
+    where
+        Self::Bus: SpiBus,
```
GrantM11235, line 219:
```suggestion
    fn transfer_in_place<Word>(&mut self, buf: &mut [Word]) -> Result<(), Self::Error>
    where
        Self::Bus: SpiBus<Word>,
        Word: Copy,
```

GrantM11235 (commented):
The convenience methods should be generic over the word type

Dirbaio:
New version
- Make SpiDevice helper methods generic over word type. (Thanks @GrantM11235, very nice catch! :D)

src/spi/blocking.rs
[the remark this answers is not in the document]
ryankurte: it seems to me like the cases we're talking about here is outside of the norm, and there are plenty of ways to work around the trait to achieve this (use `SpiBus` directly, provide a `NoOpPin` to ExclusiveDevice as described) without fundamental changes here, so i would like to call this resolved?

ryankurte (approved):
still lgtm, shall we get this done?

almindor (approved):
Yes please! We can incorporate this into https://github.com/riscv-rust/e310x-hal/pull/42 and test things out soon.

therealprof (approved):
Looks like we're all agreed.

Let's get this merged. 🎉

bors r+

PR #352, opened 2022-02-04
Remove Into<Time> generics

burrbull:
Reason:
`fugit` uses it's own convertion technic and works bad with `Into`. Same as `embedded-time` don't use `Into` for potentially fallible convertions


eldruin:
Thank you for bringing this up @burrbull.
As we agreed yesterday to merge #324 and thus remove the traits for now, I will close this.
However, we should definitely remember about this when we discuss the associated type constraints when adding the traits back.
Please bring this back up for proper discussion then if we forget about it.

burrbull:
> If you look at [fugit::ExtU32](https://docs.rs/fugit/latest/fugit/trait.ExtU32.html) it let you pass `10.millis()` type inference for automatic detecting of `NOM` and `DENOM` regardless of the resulting type. It works very well if resulting type is **known** (without any `Into<T>`). If you don't know resulting type than type inference don't work and you need to specify it by hand.

Looks like this was fixed. But can't understand how.

PR #354, opened 2022-02-09
Remove Qei trait as it has an unconstrained associated type

eldruin:
For now. See #324 for the reasoning.


Dirbaio (approved):
:ok_hand: :+1:

eldruin:
If you are interested in bringing this trait back, please participate in #362

PR #365, opened 2022-02-16
spi: allow impls returning early, add flush. 

### LINKED ISSUE #264: Clarify SPI timing behaviour (currently completely broken!)

Dirbaio:
Fixes #264 
Depends on #351 

- Document that all methods may return early, before bus is idle.
- Add a `flush` method to wait for bus to be idle.

The alternative is to document that methods MUST wait for bus idle before returning, but that would be bad for performance IMO. Allowing early return means multiple successive writes can "overlap" so that bytes come out back-to-back without gaps, which is great for workloads with tons of small writes, like SPI displays with `embedded-graphics`.

Drivers using `SpiDevice` won't have to worry about this, it'll Just Work since Device impls must flush before deasserting CS. Drivers using SpiBus will have to care if they coordinate SPI activity with other GPIO activity.

### DIFF
--- a/src/spi/blocking.rs
+++ b/src/spi/blocking.rs
@@ -143,7 +143,7 @@
 //!
 //! # Flushing
 //!
-//! To improve performance, Bus implementations are allowed to return before the operation is finished, i.e. when the bus is still not
+//! To improve performance, [`SpiBus`] implementations are allowed to return before the operation is finished, i.e. when the bus is still not
 //! idle.
 //!
 //! When using a [`SpiBus`], call [`flush`](SpiBusFlush::flush) to wait for operations to actually finish. Examples of situations
--- a/src/spi/blocking.rs
+++ b/src/spi/blocking.rs
@@ -140,6 +140,32 @@
 //!
 //! HALs **must not** add infrastructure for sharing at the [`SpiBus`] level. User code owning a [`SpiBus`] must have the guarantee
 //! of exclusive access.
+//!
+//! # Flushing
+//!
+//! To improve performance, Bus implementations are allowed to return before the operation is finished, i.e. when the bus is still not
+//! idle.
+//!
+//! When using a [`SpiBus`], call [`flush`](SpiBusFlush::flush) to wait for operations to actually finish. Examples of situations
+//! where this is needed are:
+//! - To synchronize SPI activity and GPIO activity, for example before deasserting a CS pin.
+//! - Before deinitializing the hardware SPI peripheral.
+//!
+//! When using a [`SpiDevice`], you can still call [`flush`](SpiBusFlush::flush) on the bus within a transaction.
+//! It's very rarely needed, because [`transaction`](SpiDevice::transaction) already flushes for you
+//! before deasserting CS. For example, you may need it to synchronize with GPIOs other than CS, such as DCX pins
+//! sometimes found in SPI displays.
+//!
+//! For example, for [`write`](SpiBusWrite::write) operations, it is common for hardware SPI peripherals to have a small
+//! FIFO buffer, usually 1-4 bytes. Software writes data to the FIFO, and the peripheral sends it on MOSI at its own pace,
+//! at the specified SPI frequency. It is allowed for an implementation of [`write`](SpiBusWrite::write) to return as soon
+//! as all the data has been written to the FIFO, before it is actually sent. Calling [`flush`](SpiBusFlush::flush) would
+//! wait until all the bits have actually been sent, the FIFO is empty, and the bus is idle.
+//!
+//! This still applies to other operations such as [`read`](SpiBusRead::read) or [`transfer`](SpiBus::transfer). It is less obvious
+//! why, because these methods can't return before receiving all the read data. However it's still technically possible
+//! for them to return before the bus is idle. For example, assuming SPI mode 0, the last bit is sampled on the first (rising) edge
+//! of SCK, at which point a method could return, but the second (falling) SCK edge still has to happen before the bus is idle.
 
 use core::fmt::Debug;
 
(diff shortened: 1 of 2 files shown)


ryankurte:
a [diff](https://github.com/embassy-rs/embedded-hal/compare/spi-transaction...spi-flush) against #351 so this is easier to review pre that landing

src/spi/blocking.rs
```diff
@@ -1,108 +1,338 @@
...
+//! benefits. (There's no point in a HAL implementing [`SpiDevice`] if the CS management is software-only, this task is better left to
+//! the HAL-independent implementations).
+//!
+//! HALs **must not** add infrastructure for sharing at the [`SpiBus`] level. User code owning a [`SpiBus`] must have the guarantee
+//! of exclusive access.
+//!
+//! # Flushing
+//!
+//! To improve performance, Bus implementations are allowed to return before the operation is finished, i.e. when the bus is still not
+//! idle. You must call [`flush`](SpiBusFlush::flush) to wait for operations to actually finish, for example before deasserting CS in a
```
ryankurte, line 135: i'm not sure exactly how to reframe this but, it might be nice to separate the `Bus` and `Device` components, something like:

```suggestion
//! idle. When using a `Bus` directly if you need to ensure the bus is idle you can manually call [`flush`](SpiBusFlush::flush).
//!  [`SpiDevice`] implementations **must** call [`flush`](SpiBusFlush::flush) prior to de-asserting CS to maintain ordering guarantees.
```

Dirbaio: You may still have to flush even if using a SpiDevice, for example in an SPI display to ensure the DCX pin toggles at the right time wrt the spi data. It's rare though....

src/spi/blocking.rs
```diff
@@ -1,108 +1,338 @@
...
+//! [`SpiDevice`] implementation, or before deinitializing the hardware SPI peripheral.
+//!
+//! For example, for [`write`](SpiBusWrite::write) operations, it is common for hardware SPI peripherals to have a small
+//! FIFO buffer, usually 1-4 bytes. Software writes data to the FIFO, and the peripheral sends it on MOSI at its own pace,
+//! as the specified SPI frequency. It is allowed or an implementation of [`write`](SpiBusWrite::write) to return as soon
```
adamgreig, line 140:
```suggestion
//! at the specified SPI frequency. It is allowed for an implementation of [`write`](SpiBusWrite::write) to return as soon
```

src/spi/blocking.rs
ryankurte:
```
, while [`SpiDevice`] users may choose to flush within an transaction if required.
```

Dirbaio:
New version

- Rebased on master
- Expanded docs on when flush might be needed.

src/spi/blocking.rs
Dirbaio: I've expanded docs on when flush might be needed, hopefully it's much straightforward now!

ryankurte (approved):
lgtm~! any other comments @hal? it'd be good to get this merged before too many folks refactor their SPI impls.

src/spi/blocking.rs
```diff
@@ -140,6 +140,32 @@
 //!
 //! HALs **must not** add infrastructure for sharing at the [`SpiBus`] level. User code owning a [`SpiBus`] must have the guarantee
 //! of exclusive access.
+//!
+//! # Flushing
+//!
+//! To improve performance, Bus implementations are allowed to return before the operation is finished, i.e. when the bus is still not
```
eldruin, line 146:
```suggestion
//! To improve performance, [`SpiBus`] implementations are allowed to return before the operation is finished, i.e. when the bus is still not
```

eldruin (commented):
Other the nitpick below, looks good to me!

eldruin (approved):
Thank you!

therealprof (approved):
LGTM, too.

bors r+

PR #368, opened 2022-02-23
Document MSRV upgrade and improve readme

### LINKED ISSUE #364: MSRV for 1.0
Current MSRV is "1.46+", #351 bumps it to "1.54+". 

Bumping MSRV is a breaking change, so it'd be "technically" wrong to do it on a minor/patch release, this means we'll be stuck with the same MSRV for the entire lifetime of 1.0, which is not feasible (for example stable async traits will require a MSRV bump).

Perhaps we should document it as "1.54+ for now, but we may bump it in minor versions" or simply "latest stable Rust" like the [WG policy](https://github.com/rust-embedded/wg/blob/master/ops/msrv.md).

eldruin:
I moved some of the sections in the README out into separate documents in a `docs` folder.
I have left the section about alpha releases in there, though. I think it is important enough to grant it its placement.
I have added a document describing a policy to upgrade the MSRV, since this will eventually be important once we have released the `1.0.0` version.

TL;DR: The MSRV may be updated in major or minor releases but not on patch releases.

Closes #364

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ version and an `-alpha` version side by side in a HAL.
 
 - [How-to: add a new trait](docs/how-to-add-a-new-trait.md)
 - [Version policy](docs/version-policy.md)
+- [MSRV](docs/msrv.md)
 
 ## Implementations and drivers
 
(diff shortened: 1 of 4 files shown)


Dirbaio (dismissed):

therealprof (dismissed):
LGTM

ryankurte (approved):
lgtm!

bors r+

PR #373, opened 2022-03-11
async/spi: add helper methods to SpiDevice.

Dirbaio:
It's not possible to add them as default methods to `SpiDevice` itself,
but it's possible to do via an extension trait!

With this, you can avoid interacting with the horrible async closure hack in most
cases. You can now do `my_device.write(&buf).await`, just like with bare buses.

### DIFF
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -53,6 +53,164 @@ pub trait SpiDevice: ErrorType {
             > + 'a;
 }
 
+/// Helper methods for SpiDevice.
+///
+/// This is automatically implemented for all `T: SpiDevice`, you can directly
+/// use the methods on any `SpiDevice`.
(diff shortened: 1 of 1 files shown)


eldruin (commented):
Nice!
Can it be a problem downstream to have the automatic implementation of `SpiDeviceExt` for all `T: SpiDevice` ?

Dirbaio:
Yes. Actually, all SpiDeviceExt impl downstream will conflict with the one in this crate. It's impossible to impl `SpiDeviceExt` downstream.

That's exactly what we want though. Downstream is supposed to impl just SpiDevice. That makes `SpiDeviceExt` automatically available.

ryankurte:
oooh, looks neat!

(gosh i can't wait for actual async traits 😂)

eldruin (approved):
Alright, let's get this merged and gather experiences. Thank you!
bors r+

PR #374, opened 2022-03-15
spi/blocking: Add word methods and default impls that use them

GrantM11235:
- Rename all methods that deal with slices to include "slice".
- Add `read_word`, `write_word`, and `transfer_word` methods.
- Add default implementations for slice methods based on word methods.

This will also allow us to add new methods in the future with default implementations as a non-breaking change.


Dirbaio:
I'm not convinced the `_word` variants are useful. 
- They can't be implemented with DMA, so I would *not* add them to the async traits, but this makes async+blocking inconsistent.
- I'm not sure it has a performance advantage. If you call `write` with a slice with `len=1` the compiler can inline `write` and then simplify all the code baking in the assumption that `len=1`, for example eliminating the loop. This would give final compiled code similar to what `write_word` would've.

> This will also allow us to add new methods in the future with default implementations as a non-breaking change.

Why is that not the case already, with the current slice-based methods?

eldruin:
I share @Dirbaio's analysis. I would rather not increase the API surface if there is no very good reason for it.
Could you elaborate what that would be?

GrantM11235:
As discussed [in the meeting](https://libera.irclog.whitequark.org/rust-embedded/2022-03-22#31970001;), I find the arguments against these methods to be convincing.

PR #375, opened 2022-04-12
Prepare 1.0.0-alpha.8 release

eldruin:
Even though the I2C rework similar to SPI is not available, it would be good to start gathering some experiences.
This release is a requirement for an initial alpha release of `embedded-hal-async`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


ryankurte (approved):
i think it would be great to have the i2c too, but, lgtm

eldruin:
Yes, I initially wanted to wait too but it is also easier to adapt HALs if you only have to adapt SPI to get it working and not SPI and I2C at once.
Since we are still in the alpha phase, I think it would be fine.
bors r=ryankurte

PR #376, opened 2022-04-12
Remove ADC traits

eldruin:
The traits are quite complicated to use in generic code.
It seems users would be better served by inherent methods.
Also, these traits are `nb`-only.
I would propose to remove these traits before 1.0.0 and readd/redesign them later if needed.
TODO if merged:
- Add tracking issue like for the other removed traits.


ryankurte (approved):
:+1:

therealprof (approved):
Okay then.

bors r+

PR #379, opened 2022-04-15
Prepare e-h-async 0.1.0-alpha.0 release

eldruin:
Since it is based on an `embedded-hal` alpha release, it seems appropriate for it to also be an alpha release.


ryankurte (approved):

eldruin:
bors r=ryankurte

PR #380, opened 2022-04-26
async/spi: replace the "give back the Bus" hack with a raw pointer.

Dirbaio:
The async SPI trait contains a "hack" to workaround limitations in Rust's borrow checker: #347 

It turns out the hack doesn't allow many shared bus implementations, for example when using an async Mutex: The `transaction` method gets `&'a mut self`, and the closure wants `&'a mut Self::Bus`. This only works if the Bus is a direct field in `self`. In the mutex case, the Bus has to come from inside the `MutexGuard`, so it'll live for less than `'a`.

See https://gist.github.com/kalkyl/ad3075182d610e7b413b8bbe1228ab90 which fails with the following error:

```
error[E0597]: `bus` does not live long enough
  --> src/shared_[spi.rs:78](http://spi.rs:78/):34
   |
63 |     fn transaction<'a, R, F, Fut>(&'a mut self, f: F) -> Self::TransactionFuture<'a, R, F, Fut>
   |                    -- lifetime `'a` defined here
...
78 |             let (bus, f_res) = f(&mut bus).await;
   |                                --^^^^^^^^-
   |                                | |
   |                                | borrowed value does not live long enough
   |                                argument requires that `bus` is borrowed for `'a`
...
89 |         }
   |         - `bus` dropped here while still borrowed
```

This is an alternative hack. If lifetimes don't work, simply don't use lifetimes at all! 

- Downside: it needs `unsafe{}` in all callers of transaction (but these should be rare, many users will use the `SpiDeviceExt` helpers.
- Upside: it's now possible to write sound shared bus impls.
- Upside: it no longer requires the "give back the bus" hack, it's now possible again to use `?` inside the closure for error handling.

cc @kalkyl

### DIFF
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -22,13 +22,8 @@ pub trait SpiDevice: ErrorType {
     where
         Self: 'a,
         R: 'a,
-        F: FnOnce(&'a mut Self::Bus) -> Fut + 'a,
-        Fut: Future<
-                Output = (
-                    &'a mut Self::Bus,
-                    Result<R, <Self::Bus as ErrorType>::Error>,
-                ),
-            > + 'a;
+        F: FnOnce(*mut Self::Bus) -> Fut + 'a,
+        Fut: Future<Output = Result<R, <Self::Bus as ErrorType>::Error>> + 'a;
 
     /// Perform a transaction against the device.
     ///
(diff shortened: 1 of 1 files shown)


ryankurte:
hmmm interesting, so mutexes in embassy are a singleton with shared references rather than a cheaply cloned object? 
ime in the async world you'd usually clone the mutex before the `async move` to avoid the issue, just having a play to see if there are any other strange workarounds.

Dirbaio:
> mutexes in embassy are a singleton with shared references rather than a cheaply cloned object?

It works the same as std or tokio's Mutex. Maybe you're confusing it with `Arc<Mutex<..>>`?

ryankurte (approved):
> Maybe you're confusing it with Arc<Mutex<..>>?

riight, yep that's it. this _feels_ like something we should be able to solve without pointers but, i can't see how rn so lgtm!

bors r+

PR #382, opened 2022-05-03
async/spi: make helpers default methods instead of an extension trait.

Dirbaio:
It turns out you *can* have default methods on a GAT-based async trait. All
you have to do is move the TAITs out of the trait so Rust infers them
instead of the impl specifying them.

Impls can't override the methods, but they already couldn't before with
the extension trait.

This makes async SpiDevice more consistent with blocking, and makes the helpers
more discoverable.


ryankurte (approved):
wow nice improvement, lgtm!

bors r+

PR #383, opened 2022-05-04
async: remove useless `+ 'a` in TAITs.

Dirbaio:
These aren't needed, they're already implied by the type having `'a` as a generic param.


ryankurte (approved):
:+1:

bors r+

PR #384, opened 2022-05-10
Make `can::Id` usable as key in `BTreeMap` and `HashMap`

usbalbin:
Implement `Ord` and `Hash` for `can::Id` to make it usable as key in `BTreeMap` and `HashMap`

### DIFF
--- a/src/can/id.rs
+++ b/src/can/id.rs
@@ -85,13 +85,37 @@ impl ExtendedId {
 }
 
 /// A CAN Identifier (standard or extended).
-#[derive(Debug, Copy, Clone, Eq, PartialEq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub enum Id {
     /// Standard 11-bit Identifier (`0..=0x7FF`).
     Standard(StandardId),
 
     /// Extended 29-bit Identifier (`0..=0x1FFF_FFFF`).
     Extended(ExtendedId),
+}
+
+impl Ord for Id {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        let split_id = |id: &Id| {
+            let (standard_id_part, ide_bit, extended_id_part) = match id {
+                Id::Standard(StandardId(x)) => (*x, 0, 0),
+                Id::Extended(x) => (
+                    x.standard_id().0,
+                    1,
+                    x.0 & ((1 << 18) - 1), // Bit ID-17 to ID-0
+                ),
+            };
+            (standard_id_part, ide_bit, extended_id_part)
+        };
+
+        split_id(self).cmp(&split_id(other))
+    }
+}
+
+impl PartialOrd for Id {
+    fn partial_cmp(&self, other: &Id) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
 }
 
 impl From<StandardId> for Id {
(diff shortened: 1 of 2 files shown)


timokroeger:
I think having `Ord` for `StandardId` and `ExtendedId` makes perfect sense.
For `can::Id` the question is if `Ord` should be implement according to CAN bus arbitration rules. In that case the automatic derive would not work.

usbalbin:
Does that mean that 
`Id::Standard(StandardId::ZERO) > Id::Standard(StandardId::MAX)` should be true? Thus small number meaning important means "large"? :)

usbalbin:
If in that case, would it not make more sense if the ordering was the same for all three types?

timokroeger:
Unfortunately arbitration rules are a bit more complex than that if I understand the CAN specification correctly.
When two messages are transmitted at the same time the ID with the first dominant bit (0bit) wins.
To compare standard and extended ids you would need to shift the standard ID by 18bits to align them with the MSB of the extended ID and also account the IDE bit.

Example:
Std(0) < Ext(0) < Ext(1337) < Std(1) < Ext(555555) < Std(4)

usbalbin:
Oh, ok thanks
I will have a closer look at the protocol :)

usbalbin:
So something more like
```rust
impl Ord for Id {
    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
        let split_id = |id: &Id| {
            let (standard_id_part, ide_bit, extended_id_part) = match id {
                Id::Standard(StandardId(x)) => (*x, 0, 0),
                Id::Extended(x) => (
                    x.standard_id().0,
                    1,
                    x.0 & ((1 << 18) - 1), // Bit ID-17 to ID-0
                ),
            };
            (standard_id_part, ide_bit, extended_id_part)
        };

        split_id(self).cmp(&split_id(other))
    }
}
```

eldruin (commented):
Good idea, thank you!
I have not looked in detail but a couple of things that would be necessary for the merge would be:
- Documentation on how to use this comparison and how it behaves with regard to the CAN arbitration rules and so on.
- An entry in the changelog about this.

timokroeger:
Looks good to me now. Can you please fix the two tests too:
```
failures:
    can::id::tests::cmp_extended
    can::id::tests::cmp_standard
```

usbalbin:
I hope I have put the documentation in the right place. Otherwise please let me know.

The failing test have been removed, they were forgotten left overs from my first attempt

usbalbin:
Anything more I can do?

src/can/id.rs
```diff
@@ -94,6 +94,44 @@ pub enum Id {
...
+/// from the beginning. A bit with the value 0 is dominant and a bit with
+/// value of 1 is recessive.
+///
+/// When two devices are sending frames at the same time, as soon as the first
+/// bit is found which differs, the frame with the corresponding dominant
+/// 0 bit will win and get to send the rest of the frame.
+///
+/// This implementation of `Ord` for `Id` will take this into consideration
+/// and when comparing two different instances of `Id` the "smallest" will
+/// always be the id which would form the most dominant frame, all other
```
eldruin, line 109:
```suggestion
/// always be the ID which would form the most dominant frame, all other
```

eldruin (commented):
Looks good to me, just a couple of tiny nitpicks that I can merge myself.

eldruin (approved):
Thank you very much @usbalbin and @timokroeger!
bors r+

PR #385, opened 2022-05-11
[Proposal] I2S traits - take 2

eldruin:
_(This was all done and discussed in #204 but this crate continued its evolution before the merge. Here I just adapted the traits to how we do things now: 1 blocking trait with all methods, no `nb` variant as that will be deprecated eventually.
The proof implementations need to be adapted (sorry!) but this should be pretty easy)_

I added some I2S traits to kick off the discussion started in #203.
The traits accept two words for left and right and are otherwise similar to SPI as of the 0.2.x version.

An interleaved version can be found in the [`i2s-interleaved`](https://github.com/eldruin/embedded-hal/tree/i2s-interleaved) branch. There the data should be passed interleaved. e.g. `[left0,right0,left1,right1,...]`

A version of these traits for the embedded-hal 0.2.x version can be found in the [`i2s-v0.2.x`](https://github.com/eldruin/embedded-hal/tree/i2s-v0.2.x) branch.

There is another MCU implementation in [`stm32f4xx-hal`](https://github.com/stm32-rs/stm32f4xx-hal/pull/265)

TODO:
- [ ] Demo microcontroller HAL implementation:
    - [ ] [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal/pull/99) - [discussion issue](https://github.com/stm32-rs/stm32h7xx-hal/issues/90)
    - [ ] [[WIP] stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/pull/145)
- [ ] [Bit-banging HAL implementation supporting 16-bit and 32-bit packets](https://github.com/eldruin/bitbang-hal/blob/i2s-v0.2.x/src/i2s.rs)
    - [ ] [Working example tested using both MAX98357A and PCM5102A](https://github.com/eldruin/driver-examples/blob/i2s-bitbang-bp/stm32f1-bluepill/examples/i2s-audio-bitbang.rs)
- [ ] [Working example of saw wave](https://github.com/maxekman/stm32f407g-disc/blob/dac-support-with-cs43l22/examples/audio.rs)
    - Uses driver: https://github.com/maxekman/cs43l22
- [ ] Full-duplex trait implementation
    - [ ] [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal/pull/99) - [discussion issue](https://github.com/stm32-rs/stm32h7xx-hal/issues/90)
- [X] Changelog entry.
- [ ] Add documentation about what happens if the buffer sizes differ.

Thoughts?
cc. @maxekman @samcrow @YruamaLairba

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- I2S traits
+
 ## [v1.0.0-alpha.8] - 2022-04-15
 
 *** This is (also) an alpha release with breaking changes (sorry) ***
--- a/src/i2s.rs
+++ b/src/i2s.rs
@@ -1,0 +1,35 @@
+//! I2S API
+
+/// Blocking I2S traits
+pub mod blocking {
+
+    /// Blocking I2S trait
+    pub trait I2s<W> {
+        /// Error type
+        type Error: core::fmt::Debug;
+
+        /// Reads enough bytes to fill `left_words` and `right_words`.
+        fn read<'w>(
+            &mut self,
+            left_words: &'w mut [W],
+            right_words: &'w mut [W],
+        ) -> Result<(), Self::Error>;
+
+        /// Sends `left_words` and `right_words`.
+        fn write<'w>(
+            &mut self,
+            left_words: &'w [W],
+            right_words: &'w [W],
+        ) -> Result<(), Self::Error>;
+
+        /// Sends `left_words` and `right_words`.
+        fn write_iter<LW, RW>(
+            &mut self,
+            left_words: LW,
+            right_words: RW,
+        ) -> Result<(), Self::Error>
+        where
+            LW: IntoIterator<Item = W>,
+            RW: IntoIterator<Item = W>;
+    }
+}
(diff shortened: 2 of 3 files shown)


YruamaLairba:
Hello, is the API still open for change ? I think an API based on `[(W,W)]` instead of `left:[W], right:[W]` or `interleaved:[W]` will prevent some stupid situation, for example :
- an user may want only left channel and use `left: [W;128]` and `right: [W;0]`, does implementation expect that ?
- a left or right sample is corrupted. Naive implementation may just drop It, causing synchronization issue between channel.
- a malicious user use an interleaved with odd number of elements, how methods should behave ?

eldruin:
Changes are still possible here. The reason I chose separate buffers for left and right instead of `[(W,W)]` is because they are much easier to deal with than an array of tuples and from my experience, both channels are probably being handled with separate buffers in the rest of the application already.

Of course this leads to situations like the ones you listed and I would expect any implementation to either deal with them in some way or impose additional restrictions like requiring that both buffers have the same length.

I will add some more documentation to the trait about this.

eldruin:
I published this trait in a separate [`embedded-i2s`](https://github.com/eldruin/embedded-i2s-rs) crate.
I would encourage anybody to continue the discussion over there.
A discussion about integration into `embedded-hal` would come much later.

PR #391, opened 2022-07-26
Change missing_docs from deny to warn.

Dirbaio:
It is annoying to get hard errors while developing, it prevents testing things out
without documenting them because it fails the build.

This changes it to warn, which has the same effect in CI (because it already has `RUSTFLAGS: '--deny warnings'`),
but still allows buildilng.


eldruin (approved):
Seems fine to me, thanks!
bors r+

Issue #340, opened 2021-12-20
IoPin trait not practical to implement with multiple input/output types


davidlattimore:
The IoPin trait - added in #29, seems like it would be impractical to implement in a system with multiple input / output pin types. The STM32 HALs that I'm familiar with (g0, l4, h7) use different types for OpenDrain and PushPull pin types. Because IoPin is parameterized on both input and output pin types, it appears that you would need to implement every combination NxM (where N is the number of input types and M is the number of output types).

Is there a reason why it's one trait rather than two? I guess having a single trait can be good if you want to construct a trait object - although I don't think you can do that in this case anyway, since the methods take self by value.

The only implementation I managed to find was in [linux-embedded-hal](https://github.com/rust-embedded/linux-embedded-hal/pull/60/commits/112aa9e91ff7da41d2fb6aa8d86ff57cfebf00cb) which seems to only have a single type for pins.

Dirbaio:
There's one in stm32f4xx-hal. I agree the trait is quite unwieldy, yes: https://github.com/stm32-rs/stm32f4xx-hal/blob/dc1552b2fc2e14901203a13ab42e76a5121a08eb/src/gpio.rs#L574-L683

It also feels strange to me `TInput, TOutput` are trait generic params instead of associated types. 

Also, taking `self` interacts badly with having driver structs own the pin, because it changes type on the fly so you have to store an enum in the struct to store all the possible states...

I wonder if it'd be better to have a trait for a pin that you can switch at runtime without changing type.

```rust
/// A pin that can switch between input and output modes at runtime
pub trait IoPin: InputPin + OutputPin {
    /// Configures the pin to operate in input mode
    fn as_input(&mut self);
    /// Configures the pin to operate in output mode
    fn as_output(&mut self)
}
```

ryankurte:
> I wonder if it'd be better to have a trait for a pin that you can switch at runtime without changing type.

this seems preferable to me, maybe with `set_(input|output)` and `is_(input|output)`?

i tried to do some trickier type-based stuff but, i think there's elegance in this simplicity

kelnos:
I've run into some difficulties with `IoPin` as well, but from a different angle: a crate that uses an `IoPin` and wants to hold onto it for the life of a struct.

Initially I implemented the struct (simplified) as

```rust
pub struct Device<E, P>
where
    E: core::fmt::Debug,
    P: InputPin<Error = E> + OutputPin<Error = E>,
{
    pin: P,
}
[...]

PR #393, opened 2022-07-26
Remove IoPin

Dirbaio:
As discussed in #340 it's unclear if the current IoPin trait is the best way to go (tldr: due to the fact it changes type it's hard to own it in driver structs).

I propose temporarily removing it, to unblock the 1.0 release. When discussion settles on a final design, we can add it back.


eldruin (approved):
Thanks! Seems reasonable, though unfortunate, to me in light of #340.
Especially the argument about pins being dropped for good.

TODO if merged:
- [ ] Add tracking issue

Further thoughts @rust-embedded/hal?

Issue #387, opened 2022-05-25
It's not clear when `can::ErrorKind::Acknowledge` should be returned


jonas-schievink:
Its docs say "An ACK error shall be detected by a transmitter whenever it does not monitor a dominant bit during the ACK slot.", but `Can::transmit` only enqueues a frame for transmission, so it's impossible to detect this condition for the frame passed there. `Can::transmit` *could* check if any previous frame caused an ACK error and return an error, but then its return value indicates whether a *different* frame was transmitted successfully.

eldruin:
cc: @timokroeger

timokroeger:
I fully agree with jonas here.
The error kinds were chosen to map to the error conditions defined by the CAN spec (same wording in the doc strings as the description in the spec).
For our `transmit()` interface returning an ACK error does not make sense as those should be handled internally by the controller.

eldruin:
We discussed in the WG meeting that we could keep the variant but improve its documentation.
Any change suggestions?

Issue #381, opened 2022-04-27
Additional CAN trait for extended features


mlsvrts:
The current embedded-hal CAN trait seems to only support 'transmit' and 'receive', but commonly higher level applications will need support at least:

- Opening and closing the node
- Setting the baudrate

Additionally, it can be useful to have support for:

- Reading the bus state flags
- Putting the device into listen-only mode

For instance, consider the options supported by the linux can-utils slcand daemon: https://github.com/linux-can/can-utils/blob/master/slcand.c

I think that it would be reasonable to have some `CanState` trait, that supports open/close, alongside a `CanSpeed` trait that allows setting (+getting?) the device baudrate -- but I would be interested to hear other opinions/comments.

ryankurte:
howdy, thanks for the issue!

the main goal with `e-h` is to abstract what is required for [drivers](https://docs.rs/embedded-hal/latest/embedded_hal/#design-goals) rather than how to interact with a given platform, because that differs so substantially between platforms and cross platform applications must address this configuration -anyway- (think mapping pins, clocks, etc.)

i have not seen a specific need for these yet, but if there are drivers with a demonstrable need for these APIs at runtime the first steps would be to ensure your proposed traits could be supported by a collection of platforms and used by a driver or two, then to open the appropriate PRs.

mlsvrts:
@ryankurte Thanks for the feedback!

Maybe I'm misunderstanding the goals of this project. I expect a stack like:

```
host system
------
comm. transport (usb, uart, i2c, etc.)
------
can hal  <------------- embedded hal goes here
------
can controller (embedded hardware specific)
```

I'm not actually aware of many CAN stacks that *don't* support programmable baudrates: kvaser, pcan, linux vcan, etc. all support programmable bitrates, and [Zephyr RTOS also seems to consider it a integral CanBus API feature](https://docs.zephyrproject.org/latest/hardware/peripherals/canbus/controller.html?highlight=can#setting-the-bitrate).

So if I want to write a `uart --> can` driver using `e-h`, I'll find myself unable to implement all of the supported features (namely, programming the bitrate and open/close).

If all you need is drivers + hardware that support a shared feature-set, I'm happy to provide a list and open a PR with my suggested traits. Or do you need something more concrete than that?

timokroeger:
Related discussion: https://github.com/rust-embedded/embedded-hal/pull/212#issuecomment-635822962

adamgreig:
[...]

PR #394, opened 2022-07-26
Split nb and can traits to separate crates.

Dirbaio:
Following up from https://github.com/rust-embedded/embedded-hal/issues/177#issuecomment-1162212193, this PR splits `nb` and `can` traits to separate crates.

I propose keeping these crates (and `embedded-hal-async`) separate forever. (instead of merging EHA into EH when stable as we originally planned when we created EHA).

Reasons for splitting `nb`:
- its future is unclear. Some people who would like to see `nb` deprecated. By splitting, if we do deprecate it, we won't get stuck with it in the main `embedded-hal` crate forever.
- It makes the module paths in the main crate cleaner.
- Traits tied to a particular "execution model" in their own crates, which make the overall structure easier to understand IMO.

Reasons for splitting `can`:
- It has some open concerns: #381 #387 
- Even if we do solve these, it's nice to have separate crates for the "more specialized" traits so that they can be major-bumped in the future if more concerns arise, without having to major-bump the main `embedded-hal` crate.

TODO:

- [x] Name it `embedded-can`, not `embedded-hal-can`.
- [x] Move/remove docs/examples from the main crate that mention `nb`.
- [x] Mention the subcrates from the main crate's docs.
- [x] Add CI
- [x] move `embedded-hal` to a subdir as well
- [x] What should the nb, can crate versions be? 1.0? (no opinion on this from me).
   - [x] can: 0.4
   - [x] nb: 1.0


burrbull:
Could you add links to `-async` and `-nb` crates in `embedded-hal` README.md ?

adamgreig:
We discussed this a bunch at today's meeting (logs starting [here](https://libera.irclog.whitequark.org/rust-embedded/2022-07-26#32629470;)), a few key points:

* Lots of support for splitting `nb` traits into `embedded-hal-nb` (and therefore keeping `embedded-hal-async` separate for good)
* For CAN, it seems like it might make most sense to use a new repository and publish to the old `embedded-can` crate rather than an `embedded-hal-can`; in general this approach seems favourable for more complicated protocols/peripherals/concepts like CAN, USB, DMA, etc (see also embedded-dma, embedded-i2s, etc). This way the CAN traits can change/have breaking releases/develop without impacting the embedded-hal crate versioning, without being much more work for HAL/driver/end users to actually use.

ryankurte:
outcomes from https://github.com/rust-embedded/embedded-hal/pull/394#issuecomment-1195879618 sound good to me!

(looking forward to daylight savings here in a couple of months so the meetings are a paletable 08:00 instead of 06:00 here 🤣)

adamgreig:
Discussed in today's meeting: Let's go with putting the CAN traits into an `embedded-can` folder in this repository, and then we can publish a new version of `embedded-can` with the latest traits from `embedded-hal` 0.2.

Dirbaio:
bors try

Dirbaio:
This should be now ready to go :) 
ping @rust-embedded/hal 

The changelog checker is angry because I moved `CHANGELOG.md` to the subdir. How do we do this? Ideally the checker would require updating changelog for crate X only if you've touched files from subdir X, but it doesn't seem that's supported...

embedded-can/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "embedded-can"
+version = "0.4.0"
+edition = "2018"
+
+description = "HAL traits for Controller Area Netowrk (CAN) devices."
+categories = ["embedded", "hardware-support", "no-std"]
+documentation = "https://docs.rs/embedded-hal-can"
```
eldruin, line 8:
```suggestion
documentation = "https://docs.rs/embedded-can"
```

embedded-can/README.md
```diff
@@ -0,0 +1,33 @@
+[![crates.io](https://img.shields.io/crates/d/embedded-hal-nb.svg)](https://crates.io/crates/embedded-hal-nb)
+[![crates.io](https://img.shields.io/crates/v/embedded-hal-nb.svg)](https://crates.io/crates/embedded-hal-nb)
+[![Documentation](https://docs.rs/embedded-hal-nb/badge.svg)](https://docs.rs/embedded-hal-nb)
+![Minimum Supported Rust Version](https://img.shields.io/badge/rustc-1.54+-blue.svg)
+
+# `embedded-hal-nb`
```
eldruin, line 6: Wrong file, this should be about embedded-can

embedded-hal-async/README.md
```diff
@@ -36,20 +36,3 @@ Licensed under either of
 - MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
 
 at your option.
-
-### Contribution
```
eldruin, line 40: I'm fine with removing the CoC notice in the subfolders but it might be good to keep the contribution section to avoid legal ambiguities.
(also below)

embedded-hal-bus/src/spi.rs
```diff
@@ -1,86 +1,78 @@
...
-pub mod blocking {
-    use core::fmt::Debug;
-    use embedded_hal::{
-        digital::blocking::OutputPin,
-        spi::{
-            blocking::{SpiBusFlush, SpiDevice},
-            Error, ErrorKind, ErrorType,
-        },
-    };
+use core::fmt::Debug;
```
eldruin, line 3: I think e-h-bus should keep its `blocking` module so that we can put async sharing facilities here as well. otherwise we will need yet another `embedded-hal-async-bus` crate.
An alternative would be including such stuff in the `e-h-async` crate, but it seems rather unsymmetrical.

embedded-hal-nb/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## License
+
+Licensed under either of
+
+- Apache License, Version 2.0 ([LICENSE-APACHE](LICENSE-APACHE) or
+  http://www.apache.org/licenses/LICENSE-2.0)
+- MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
+
+at your option.
```
eldruin, line 33: TODO: contribution section

embedded-hal-nb/src/serial.rs
```diff
@@ -33,3 +35,21 @@ impl<T: Write<Word>, Word: Copy> Write<Word> for &mut T {
         T::flush(self)
     }
 }
+
+/// Implementation of `core::fmt::Write` for the HAL's `serial::Write`.
+///
+/// TODO write example of usage
```
eldruin, line 41: TODO: example

eldruin (commented):
Thanks! I added a few comments.
If you do not find any alternative, we can remove the changelog check since it does not seem to work with workspace repos.
Note that this includes removing the `blocking` modules from `embedded-hal::X` and puts all the traits there.
This is in line with only supporting one execution model but should be noted as such.
Please add a changelog entry to the e-h crate about it.

embedded-hal-nb/src/serial.rs
Dirbaio: The TODO was already there :sweat_smile:

embedded-hal-bus/src/spi.rs
Dirbaio: I think it's cleaner to put all the structs in the same module, because there are some that can impl both blocking+async at the same time. For example, for ExclusiveDevice:

```rust
pub struct ExclusiveDevice<BUS, CS> {..}
impl<BUS, CS> ErrorType for ExclusiveDevice<BUS, CS> where BUS: ErrorType, ... {}
impl<BUS, CS> embedded_hal::spi::SpiDevice for ExclusiveDevice<BUS, CS> where BUS: embedded_hal::spi::SpiBusFlush, ... {}
impl<BUS, CS> embedded_hal_async::spi::SpiDevice for ExclusiveDevice<BUS, CS> where BUS: embedded_hal_async::spi::SpiBusFlush, ... {}
```

(The ExclusiveDevice in `embedded-hal-async` already does that now) 

(see [discussion](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$sYELtnHYHSEfmanWah3I-J8iauPfXeSTTUCIKxICm1k?via=matrix.org&via=psion.agg.io&via=mozilla.org) in today's meeting.)

embedded-can/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "embedded-can"
+version = "0.4.0"
+edition = "2018"
+
+description = "HAL traits for Controller Area Netowrk (CAN) devices."
```
tjallingt, line 6:
```suggestion
description = "HAL traits for Controller Area Network (CAN) devices."
```

embedded-hal-async/src/digital.rs
```diff
@@ -30,7 +30,7 @@ pub trait Wait: embedded_hal::digital::ErrorType {
     /// # Note for implementers
     /// The pin may have switched back to low before the task was run after
     /// being woken. The future should still resolve in that case.
-    fn wait_for_high<'a>(&'a mut self) -> Self::WaitForHighFuture<'a>;
+    fn wait_for_high(&mut self) -> Self::WaitForHighFuture<'_>;
```
eldruin, line 33: Do these changes need an entry in the changelog?

eldruin (commented):
After a rebase this should be good to go.
In the future let's _please avoid_ such huge patchsets.
It just makes reviewing much more time-intensive and adds delay.
There are several easier-to-review PRs hiding inside this one.

Dirbaio:
Rebased, and split off the clippy stuff to a separate PR (#406)

embedded-hal-async/src/digital.rs
Dirbaio: No, they're equivalent (lifetime elision). Clippy wants the elided version if possible. Anyway this is now split to #406.

eldruin (dismissed):
Thanks!
bors r+

.github/bors.toml
```diff
@@ -2,11 +2,15 @@ block_labels = ["needs-decision"]
...
+  "test (stable, x86_64-unknown-linux-gnu)",
+  "test (stable, thumbv6m-none-eabi)",
+  "test (stable, thumbv7m-none-eabi)",
+  "test (1.54.0, x86_64-unknown-linux-gnu)",
+  "test (1.54.0, thumbv6m-none-eabi)",
+  "test (1.54.0, thumbv7m-none-eabi)",
+  "test (nightly, x86_64-unknown-linux-gnu)",
+  "test (nightly, thumbv6m-none-eabi)",
+  "test (nightly, thumbv7m-none-eabi)",
+  "clippy",
```
eldruin, line 14:
```suggestion
```

eldruin:
bors retry

eldruin (approved):
bors r+

PR #399, opened 2022-08-14
SpiDevice: add guidelines for behaviour of CS pin upon bus failures

newAM:
Currently the `SpiDevice` trait does not define the sequence of actions that occur after a bus error.

I discussed this with @Dirbaio in the matrix chat and the conclusion we reached was that this should be added to the documentation as a guideline to match what already exists for `ExclusiveDevice`.  The rationale for a guideline instead of a requirement is that SPI error recovery is implementation specific.

### DIFF
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -66,6 +66,9 @@ pub trait SpiDevice: ErrorType {
     /// transactions from executing concurrently against the same bus. Examples of implementations are:
     /// critical sections, blocking mutexes, async mutexes, returning an error or panicking if the bus is already busy.
     ///
+    /// On bus errors the implementation should try to deassert CS.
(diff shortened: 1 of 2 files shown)


Dirbaio:
LGTM, could you do the same change to async's docs ?

newAM:
> LGTM, could you do the same change to async's docs ?

Good call.  Made the change & rebased.

Dirbaio (approved):

newAM:
Should I add a changelog entry for this?  I wasn't sure because it doesn't change anyone's impl.

Dirbaio:
IMO it's small enough that no changelog is fine, but feel free to add it if you want.

eldruin (approved):
Looks good to me, thank you!
I agree no changelog entry is necessary especially since we are still in alpha phase.
bors r+

PR #402, opened 2022-08-27
Add I2C `software_reset` method.

reitermarkus:
Add method for software reset via the “General Call” address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Implement `PartialOrd`, `Ord`, `Hash` for `can::StandardId`, `can::ExtendedId` and `can::Id` according to CAN bus arbitration rules
+- `I2c::software_reset` for resetting all (supported) devices connected to the bus.
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


eldruin (commented):
This is an interesting addition which seems fine to me as the traits stand.
However, I wonder how it would look after #392.
Specially there I think there is a non-negligible risk of people thinking it is only going to reset their device and not every device connected (which supports it) since they get access to the bus through an _`I2cDevice`_.
```rust
device.transaction(|bus| {
    bus.software_reset()
});
```
I see that `software_reset_on_every_device_which_supports_it()` seems wonky, though.
I think this would be an interesting point to add to the [comments I added here](https://github.com/rust-embedded/embedded-hal/pull/392#issuecomment-1219960952).

eldruin:
We discussed this in the meeting today and decided against merging it due to a number of reasons:
- It does not seem anybody would need anything else than the default implementation, so it would not build an abstraction that implementations later define, like with the rest of traits.
- It is trivial to do in user code if needed: `bus.write(0x0, &[0x06])`.
- It has potential for unintended foreign device resets.
- Support for I2C General Call Reset command is rare.

Nevertheless, thank you very much for the proposal!

PR #403, opened 2022-08-29
Make the async SpiDevice trait unsafe to implement

GrantM11235:
This is necessary due to the raw pointer workaround.


Dirbaio (approved):
LGTM, thank you!

It's very unfortunate `unsafe trait` is needed, but it's the correct thing to do, so :+1: . Everything else about the raw pointer workaround is unfortuate anyway... :shrug: 

bors r+

PR #404, opened 2022-09-03
async/spi: Add a transaction helper macro

GrantM11235:
Based on #403

This macro allows users to use `SpiDevice::transaction` in a completely safe way, without needing to dereference a raw pointer.

The macro is exported as `embedded_hal_async::spi_transaction_helper` because exported macros always go in the root of the crate.
It is also publicly reexported as `embedded_hal_async::spi::transaction_helper` because I think that most people would expect to find it in the spi module.
It is not possible to apply `doc(hidden)` to the instance in the root of the crate because that would also hide the reexport, see rust-lang/rust#59368.

### DIFF
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -41,14 +41,14 @@ where
 /// # Examples
 ///
 /// ```
-/// use embedded_hal_async::spi::{transaction_helper, SpiBus, SpiBusRead, SpiBusWrite, SpiDevice};
+/// use embedded_hal_async::spi::{transaction, SpiBus, SpiBusRead, SpiBusWrite, SpiDevice};
 ///
 /// pub async fn transaction_example<SPI>(mut device: SPI) -> Result<u32, SPI::Error>
 /// where
 ///     SPI: SpiDevice,
 ///     SPI::Bus: SpiBus,
 /// {
-///     transaction_helper!(&mut device, move |bus| async move {
+///     transaction!(&mut device, move |bus| async move {
 ///         // Unlike `SpiDevice::transaction`, we don't need to
 ///         // manually dereference a pointer in order to use the bus.
 ///         bus.write(&[42]).await?;
@@ -62,7 +62,7 @@ where
 ///
 /// Note that the compiler will prevent you from moving the bus reference outside of the closure
 /// ```compile_fail
-/// # use embedded_hal_async::spi::{transaction_helper, SpiBus, SpiBusRead, SpiBusWrite, SpiDevice};
+/// # use embedded_hal_async::spi::{transaction, SpiBus, SpiBusRead, SpiBusWrite, SpiDevice};
 /// #
 /// # pub async fn smuggle_test<SPI>(mut device: SPI) -> Result<(), SPI::Error>
 /// # where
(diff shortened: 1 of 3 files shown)


embedded-hal-async/src/spi.rs
```diff
@@ -35,13 +35,58 @@ where
...
+///         let mut data = [0; 4];
+///         bus.read(&mut data).await?;
+///         Ok(u32::from_be_bytes(data))
+///     })
+///     .await
+/// }
+/// ```
+macro_rules! spi_transaction_helper {
+    ($device:expr, move |$bus:ident| async move $closure_body:expr) => {
+        $crate::spi::SpiDevice::transaction($device, move |$bus| async move {
```
Dirbaio, line 64: I wouldn't force `move` here, you might want to read to a borrowed buffer captured from the parent function.

embedded-hal-async/src/spi.rs
```diff
@@ -35,13 +35,58 @@ where
...
+            // Safety: Implementers of the `SpiDevice` trait guarantee that the pointer is
+            // valid and dereferencable for the entire duration of the closure.
+            let $bus = unsafe { &mut *$bus };
```
Dirbaio, line 67: it's better to dereference the raw pointer *outside* the `async{}` block, to avoid "future is not Send/Sync because a raw pointer lives through an .await"  issues

embedded-hal-async/src/spi.rs
GrantM11235: Should all of these combinations be supported?

-  `|bus| async {}`
-  `|bus| async move {}`
-  `move |bus| async {}`
-  `move |bus| async move {}`

Dirbaio: I *think* just these would be enough

```
|bus| async {}
|bus| async move {}
```

because if the inner `async{}` captures by move the net result is the whole "thing" captures by move, so there's no point in making the outer closure `move` as well. but I'm not sure :sweat:

GrantM11235: The `transfer` convenience method only works with `move |bus| async move {}`

Dirbaio: I'm still not sure if we need all combinations, but since the macro supports them all now I'm going to mark this as resolved. Shouldn't hurt to support them all.

embedded-hal-async/src/spi.rs
```diff
@@ -35,13 +35,117 @@ where
     Word: Copy + 'static,
 = impl Future<Output = Result<(), T::Error>>;
 
+#[macro_export]
+/// This macro is a workaround for the workaround in [`SpiDevice::transaction`]
```
Dirbaio, line 39: Perhaps a more descriptive docs on why you might want to use it? Something like "Do an SPI transaction on a bus. This is a safe wrapper for [`SpiDevice::transaction`], which handles dereferencing the raw pointer for you."

embedded-hal-async/src/spi.rs
```diff
@@ -35,13 +35,117 @@ where
...
+/// # {
+///     let mut bus_smuggler: Option<&mut SPI::Bus> = None;
+///     transaction_helper!(&mut device, move |bus| async move {
+///         bus_smuggler = Some(bus);
+///         Ok(())
+///     })
+///     .await
+/// # }
+/// ```
+macro_rules! spi_transaction_helper {
```
Dirbaio, line 80: I wonder if the name should be simply `spi_transaction`. I think users will end up using this 100% of the time instead of the unsafe `transaction()` method.

embedded-hal-async/src/spi.rs
```diff
@@ -69,9 +173,13 @@ pub trait SpiDevice: ErrorType {
     /// On bus errors the implementation should try to deassert CS.
     /// If an error occurs while deasserting CS the bus error should take priority as the return value.
     ///
+    /// # Safety
```
Dirbaio, line 176: Mention that the `transaction!` safe wrapper macro exist and that using it if possible is recommended.

Dirbaio (changes requested):
I've updated a driver to use the macro, and it worked very nicely! HUGE improvement in ergonomics. https://github.com/embassy-rs/cyw43/pull/19/files

I've commented a few naming/doc nits, LGTM otherwise!

Dirbaio (approved):
Thank you!

bors r+

PR #405, opened 2022-09-21
Fix doc links after embedded-hal-bus split

GrantM11235:


eldruin (approved):
Nice catch, thanks!
bors r+

PR #406, opened 2022-09-26
Enforce clippy in CI

Dirbaio:
Depends on #394

CI wasn't actually enforcing clippy, because `actions-rs/clippy-check` needs to run as `pull_request_target` to get the token to add the "fancy" comments to the diff. It does nothing when ran as `pull_request`.

`pull_request_target` is troublesome because it uses the yaml from `master` instead of from the PR branch, which means when bumping clippy you don't know whether CI passes until *after* it's merged. I've made it use `cargo clippy` instead, which works on `pull_request`.

### DIFF
--- a/embedded-hal/src/i2c.rs
+++ b/embedded-hal/src/i2c.rs
@@ -259,7 +259,7 @@ impl AddressMode for TenBitAddress {}
 /// Transactional I2C operation.
 ///
 /// Several operations can be combined as part of a transaction.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum Operation<'a> {
     /// Read data into the provided buffer
     Read(&'a mut [u8]),
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -11,5 +11,6 @@ status = [
   "test (nightly, x86_64-unknown-linux-gnu)",
   "test (nightly, thumbv6m-none-eabi)",
   "test (nightly, thumbv7m-none-eabi)",
+  "clippy",
   "fmt",
 ]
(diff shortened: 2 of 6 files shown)


embedded-hal/src/i2c.rs
```diff
@@ -259,7 +259,7 @@ impl AddressMode for TenBitAddress {}
 /// Transactional I2C operation.
 ///
 /// Several operations can be combined as part of a transaction.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
```
eldruin, line 262: This should be noted in the changelog.

eldruin (commented):
I am not a fan of enforcing clippy. Using the nightly clippy is probably even more volatile.
Granted, it is pinned to a version that works fine now, but we will update it regularly and I see us making unnecessary/unrelated/even contestable changes in unrelated crates in order to land PRs.
For me, the failing :x: icon in CI is enough warning when reviewing a PR.

Dirbaio:
If we keep clippy but not enforcing it, the moment we decide to *not* fix a warning because we don't agree with it, all following PRs will have the red :x: forever, due to an issue unrelated to the PR. At that point it stops being a useful flag for code review, and it's also confusing for contributors because a red :x: in a PR usually means you did something wrong.

IMO if we keep it we should enforce it, otherwise we should remove it.

Dirbaio:
> making unnecessary/unrelated/even contestable changes in unrelated crates in order to land PRs.

Random warnings won't appear when landing a feature PR, because the clippy version is pinned. We can update it in dedicated PRs, so the changes for any new warnings don't mix with normal PRs.

eldruin:
Up to you @ryankurte, @therealprof

therealprof (approved):
LGTM, thanks.

bors r+

PR #407, opened 2022-09-26
async: switch to async-fn-in-traits, release v0.2.0-alpha.0

Dirbaio:
Latest Rust nightlies have somewhat usable async-fn-in-trait support already! :tada: 

embassy-nrf updated here https://github.com/embassy-rs/embassy/pull/974

Paprecuts encountered:

- there's this annoying error [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=f04fca1f2a3d643c323fb49c05bd3ed3), workaround is to use the concrete type instead of `Self::Error`. This is a limitation of all `async fn`s, not just in traits, but it hits especially hard within traits, so I dunno if there's plans to improve it.

> `async fn` return type cannot contain a projection or `Self` that references lifetimes from a parent scope

- The SpiDevice trait ICEs, issue filed https://github.com/rust-lang/rust/issues/102310
- default methods don't work, but there's a PR already https://github.com/rust-lang/rust/issues/102308

Due to the last 2 I've left `SpiDevice` alone for now.

### DIFF
--- a/embedded-hal-async/src/lib.rs
+++ b/embedded-hal-async/src/lib.rs
@@ -9,7 +9,8 @@
 
 #![warn(missing_docs)]
 #![no_std]
-#![feature(type_alias_impl_trait)]
+#![allow(incomplete_features)]
+#![feature(async_fn_in_trait, impl_trait_projections)]
 
 pub mod delay;
 pub mod digital;
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -14,7 +14,7 @@ jobs:
           profile: minimal
           # embedded-hal-async needs nightly.
           # Use a pinned version to avoid spontaneous breakages (new clippy lints are added often)
-          toolchain: nightly-2022-09-25
+          toolchain: nightly-2022-11-22
           override: true
           components: clippy
       - run: cargo clippy -- --deny=warnings
--- a/embedded-hal-async/CHANGELOG.md
+++ b/embedded-hal-async/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.2.0-alpha.0] - 2022-11-23
+
+- Switch all traits to use [`async_fn_in_trait`](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html) (AFIT). Requires `nightly-2022-11-22` or newer.
 
 ## [v0.1.0-alpha.3] - 2022-10-26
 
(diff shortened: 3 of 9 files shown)


korken89:
It seems the 2 issues with `SpiDevice` has been fixed in nighly now.

Dirbaio:
converted all traits, `embedded-hal-async` no longer explodes when compiling now!  I encountered another rustc bug https://github.com/rust-lang/rust/issues/103457 but thankfully we don't need the `+ 'a` anymore, so just removing it worked...

I haven't tried updating embassy yet, it's still possible it explodes with actual use.

korken89:
How are we on this, is this ready to merge?
I'm about to start writing some example drivers and would love this :D

If not I'll use a git-dep to this branch.

Dirbaio:
Since AFIT support is very recent in rustc we should check that it works end-to-end first. `embedded-hal-async` itself builds without ICEs now, but I haven't tried implementing it in a HAL and using it from a driver yet, it's still possible these things ICE. I wanted to update embassy but haven't found time for it yet. If you want to help test it out by using it as a git dep please go ahead! :)

When we know it works well we can merge it and release it as `embedded-hal-async 0.2` :)

korken89:
Super, then I know where it stands!
I'll git-dep for now and start hacking :)

Dirbaio:
No longer builds in nightly-2022-10-29 and newer. Filed https://github.com/rust-lang/rust/issues/103850

embedded-hal-async/src/lib.rs
```diff
@@ -10,6 +10,7 @@
 #![warn(missing_docs)]
 #![no_std]
 #![feature(type_alias_impl_trait)]
```
newAM, line 12:
```suggestion
```

minor: TAIT is no longer needed.

Dirbaio: Done, thanks!

Dirbaio:
Also, side note: A few places use `T::Error` instead of `Self::Error` because `Self` is not allowed in `async fn` return if it has lifetimes. Not a blocker since it can always be workarounded by writing the full `Self` type, but it can be annoying if it has lots of generics. Will likely be fixed in https://github.com/rust-lang/rust/pull/103491

Dirbaio:
Status update!

- `nightly-2022-11-22` contains fixes for all the blockers
- Ported embassy-nrf here https://github.com/embassy-rs/embassy/pull/974 . Builds fine, runs fine. The only issue is that nightly also introduces a [TAIT regression](https://github.com/rust-lang/rust/issues/104691), but that can be fixed by porting the affected GA+TAIT traits to AFIT.

So, I'm fairly certain this is ready to go now!

eldruin (approved):
Looks good to me, thank you!
bors r+

PR #416, opened 2022-10-04
Raise MSRV for embedded-hal-async to Rust 1.65.0

nospam2678:
Required for generic_associated_types feature.

### DIFF
--- a/embedded-hal-async/Cargo.toml
+++ b/embedded-hal-async/Cargo.toml
@@ -12,6 +12,7 @@ name = "embedded-hal-async"
 readme = "README.md"
 repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.0-alpha.0"
+rust-version = "1.65.0"
 
 [dependencies]
 embedded-hal = { version = "=1.0.0-alpha.9", path = "../embedded-hal" }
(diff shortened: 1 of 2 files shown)


Dirbaio:
Technically `embedded-hal-async` requires 1.65 because it uses GAT, but it still requires nightly because TAIT. As long as we require nightly, having a MSRV doesn't make much sense.

Could you change it instead to state something along the lines of "requires nightly newer than <date>, but keep in mind nightly can break at any time"?

nospam2678:
Updated with suggested change to the README file.

As I wrote in the chatroom, I **believe** setting `rust-version` might also take a part when cargo performs dependency resolution to determine which crates to use.

One might argue that this PR adds little value, and that's probably true. I didn't realize until now that there are no previous releases, which really makes this PR less valuable than I initially thought catching this in a release candidate would.

For context, the reason for bothering to make it is that the fact that 0.1.0-alpha.2 failed to build while 0.1.0-alpha.1 worked caused a bump in the road for me. The upgrade happened while I was in the process of updating a project with a six+ months outdated embassy. It took a session of trouble shooting that I would want to spare others, if possible.

Dirbaio (approved):
I've updated the required nightly version, which is newer now that we're using AFIT.

Thanks!

bors r+

PR #421, opened 2022-10-22
embedded-hal-async: fix compilation on latest nightly

newAM:
Compiling `embedded-hal-async` on the current nightly, rustc 1.66.0-nightly (5c8bff74b 2022-10-21), will fail.

This fixes compilation.

### DIFF
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -13,28 +13,28 @@ where
     T: SpiDevice + ?Sized + 'a,
     T::Bus: SpiBusRead<Word>,
     Word: Copy + 'static,
-= impl Future<Output = Result<(), T::Error>>;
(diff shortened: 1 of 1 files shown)


ryankurte (approved):
lgtm, and i don't think we have an MSRV for `e-h-a` given it's nightly only atm?

newAM:
> lgtm, and i don't think we have an MSRV for `e-h-a` given it's nightly only atm?

Yup, nightly only at the moment because of `#![feature(type_alias_impl_trait)]`

Dirbaio (approved):
bors r+ 

thsnk!

For the record breakage is due to https://github.com/rust-lang/rust/pull/102417. Fix here still builds with older nightlies, so this is not even a "nightly MSRV" bump (not that it's something we should track anyway)

PR #423, opened 2022-10-26
Release embedded-hal-async v0.1.0

Dirbaio:
Fixes compliation on latest nightlies, so we'd like to get this out soon.

I've named the version `0.1.0` and not `0.1.0-alpha.3` because there's not much point in using the `alpha` if the main version is `0.1` which already indicates "API is WIP". Next breaking change can be released as v0.2 (probably the switch to AFIT), etc.


eldruin:
Another rationale for the `-alpha` version is the dependency on an `-alpha` version of `embedded-hal`.
For this crate I am fine with either, though.
Do you want to go ahead with `0.1.0`?

Dirbaio:
> Another rationale for the -alpha version is the dependency on an -alpha version of embedded-hal.

Oh... I didn't think about that at all. :rofl:  That's not semver-stable at all, indeed.

`-alpha.3` it is then.

eldruin (approved):
Looks good to me, thank you!
bors r+

PR #424, opened 2022-11-01
ci: Fix embedded-hal-async not being tested.

Dirbaio:
Fixing a somewhat embarrasing mistake causing embedded-hal-async to never be tested even if using nightly.


eldruin (approved):
Thank you!
bors r+

PR #429, opened 2022-12-13
Implement From<PinState> for bool

### LINKED ISSUE #427: Implement From<PinState> for bool
`From<bool>` is already implemented for PinState in `digital.rs`, which makes sense for output pins. For input pins, it would make sense to implement the `Into<bool>` trait as well, so a function could take an argument that is either a `PinState` or a `bool`:

```rust
impl Into<bool> for PinState {
    fn into(self) -> bool {
        match self {
            PinState::Low => false,
            PinState::High => true,
        }
    }
}

fn print_state<T: Into<bool>>(state: T) {
    println!("{:?}", state.into());
}

fn main() {
    print_state(false);
    print_state(true);
    print_state(PinState::Low);
    print_state(PinState::High);
}
```

sourcebox:
Closes #427

### DIFF
--- a/embedded-hal/src/digital.rs
+++ b/embedded-hal/src/digital.rs
@@ -55,6 +55,15 @@ impl Not for PinState {
     }
 }
 
+impl From<PinState> for bool {
+    fn from(value: PinState) -> bool {
+        match value {
+            PinState::Low => false,
+            PinState::High => true,
+        }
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
Makes a lot of sense. Thank you!

bors r+

Issue #358, opened 2022-02-13
Tracking issue for PWM traits


eldruin:
The PWM traits available on the 0.2.x versions have been removed ahead of the 1.0.0 release. See: #357 
This issue servers as a tracking issue until we add them back. The open (breaking) problems should be reasonably-well solved and all associated types should have appropriate bounds that enable generic code use.

PWM traits as of the 0.2.7 release:
- [`Pwm` trait docs](https://docs.rs/embedded-hal/0.2.7/embedded_hal/trait.Pwm.html)
- [`PwmPin` trait docs](https://docs.rs/embedded-hal/0.2.7/embedded_hal/trait.PwmPin.html)
- [Source](https://github.com/rust-embedded/embedded-hal/blob/2e91b430c5f3c2e06cb3f12a7812e17cafd2d3bc/src/lib.rs#L800)

Relevant issues/PRs:
- #352 
- #256 
- #226 
- #182 
- Your issue here!

Please feel free to participate and

burrbull:
As `PwmPin` doesn't have `Time` and only `Duty`, I think it is not so big issue. We just need to decide what type to use.
Float is bad idea as timers are discrete. And could always be evaluated as `(get_duty() as f32)/(get_max_duty() as f32)`.
So `u16` or `u32`? I think in most cases `u16` is enough. But 32-bit timers can support both `u16` and `u32`.

What if we move duty from associated type to generic and make `u16` default type (like with I2c address type)?
```rust
pub trait PwmPin<Duty=u16> {
    fn disable(&mut self);
    fn enable(&mut self);
    fn get_duty(&self) -> Duty;
    fn get_max_duty(&self) -> Duty;
    fn set_duty(&mut self, duty: Duty);
}
```
In this case we "say" that `u16` is required and other types are optional.

Alternative: return u32 always (like with Delay):
```rust
pub trait PwmPin {
    fn disable(&mut self);
    fn enable(&mut self);
    fn get_duty(&self) -> u32;
    fn get_max_duty(&self) -> u32;
    fn set_duty(&mut self, duty: u32);
}
```
But in this case we have potential problems with `set_duty` on 16-bit timers.

rursprung:
[...]

PR #430, opened 2022-12-20
Add PWM SetDuty trait.

Dirbaio:
This adds back a version of the 0.2 trait `PwmPin`. 

cc #358

WG meeting [chatlog](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$_4DVnptWkGRjvW7-5Y6qCQZvj8oi9t4tbt8v4hXxJak?via=matrix.org&via=psion.agg.io&via=tchncs.de)

Differences to 0.2:

- `enable()` and `disable()` are gone. I think they were underspecified (for example, is the pin high or low when disabled?), and not very useful in practice. Disabling can be achieved already by setting duty to 0% or 100%. If the HAL cares about power consumption, it can transparently disable the timer and set the pin to fixed high/low when duty is 0% or 100%.
- Duty is no longer an unconstrained associated type. `u16` has been chosen because it gives enough dynamic range without being too big. `u8` might give too little dynamic range for some applications, `u32` might be annoyingly big for smaller archs like AVR/MSP430.
- Range is `0..u16::MAX` instead of `0..get_max_duty()`. This makes the HAL instead of the user responsible for the scaling, which makes using the trait easier. Also, if the HAL wants to optimize for performance, it can set the hardware period to be a power of 2, so scaling is just a bit shift.
- Name is `SetDuty` instead of `PwmPin`, because we might have a `SetFrequency` or similar in the future.
- I haven't included `get_duty()`, because I think in practice most drivers don't need it, and implementing it in HALs is annoying. They either have to read it back from hardware and unscaling it (losing precision), or storing the unscaled value in `self` (wasting RAM). We could add a `GetDuty` or `StatefulSetDuty` in the future if it turns out this is wanted, but I hope it won't be.

### DIFF
--- a/embedded-hal/src/pwm.rs
+++ b/embedded-hal/src/pwm.rs
@@ -59,19 +59,62 @@ impl<T: ErrorType> ErrorType for &mut T {
 
 /// Single PWM channel / pin
 pub trait SetDuty: ErrorType {
-    /// Set the duty cycle.
+    /// Get the maximum duty cycle value.
     ///
-    /// `duty` is the duty cycle. Valid values span the entire `u16` range:
+    /// This value corresponds to a 100% duty cycle.
+    fn get_max_duty(&self) -> u16;
+
+    /// Set the duty cycle to `duty / max_duty`.
     ///
-    /// - `duty = 0` is considered 0% duty, which makes the pin permanently low.
-    /// - `duty = u16::MAX` is considered 100% duty, which makes the pin permanently high.
+    /// The caller is responsible for ensuring that the duty cycle value is less than or equal to the maximum duty cycle value,
+    /// as reported by `get_max_duty`.
+    fn set_duty(&mut self, duty: u16) -> Result<(), Self::Error>;
+
+    /// Set the duty cycle to 0%, or always inactive.
+    fn set_off(&mut self) -> Result<(), Self::Error> {
+        self.set_duty(0)
+    }
+
+    /// Set the duty cycle to 100%, or always active.
+    fn set_on(&mut self) -> Result<(), Self::Error> {
+        self.set_duty(self.get_max_duty())
+    }
+
+    /// Set the duty cycle to `num / denom`.
     ///
-    /// Implementations must scale the duty value linearly to the range required by the hardware.
-    fn set_duty(&mut self, duty: u16) -> Self::Error;
+    /// The caller is responsible for ensuring that `num` is less than or equal to `denom`,
+    /// and that `denom` is not zero.
+    fn set_fraction(&mut self, num: u16, denom: u16) -> Result<(), Self::Error> {
+        let duty = num as u32 * self.get_max_duty() as u32 / denom as u32;
+        self.set_duty(duty as u16)
+    }
+
+    /// Set the duty cycle to `percent / 100`
+    ///
+    /// The caller is responsible for ensuring that `percent` is less than or equal to 100.
+    fn set_percent(&mut self, percent: u8) -> Result<(), Self::Error> {
+        self.set_fraction(percent as u16, 100)
+    }
 }
 
 impl<T: SetDuty> SetDuty for &mut T {
-    fn set_duty(&mut self, duty: u16) -> Self::Error {
+    fn get_max_duty(&self) -> u16 {
+        T::get_max_duty(self)
+    }
+    fn set_duty(&mut self, duty: u16) -> Result<(), Self::Error> {
         T::set_duty(self, duty)
+    }
+
+    fn set_off(&mut self) -> Result<(), Self::Error> {
+        T::set_off(self)
+    }
+    fn set_on(&mut self) -> Result<(), Self::Error> {
+        T::set_on(self)
+    }
+    fn set_fraction(&mut self, num: u16, denom: u16) -> Result<(), Self::Error> {
+        T::set_fraction(self, num, denom)
+    }
+    fn set_percent(&mut self, percent: u8) -> Result<(), Self::Error> {
+        T::set_percent(self, percent)
     }
 }
--- a/embedded-hal/src/pwm.rs
+++ b/embedded-hal/src/pwm.rs
@@ -71,11 +71,13 @@ pub trait SetDuty: ErrorType {
     fn set_duty(&mut self, duty: u16) -> Result<(), Self::Error>;
 
     /// Set the duty cycle to 0%, or always inactive.
+    #[inline]
     fn set_off(&mut self) -> Result<(), Self::Error> {
         self.set_duty(0)
     }
 
     /// Set the duty cycle to 100%, or always active.
+    #[inline]
     fn set_on(&mut self) -> Result<(), Self::Error> {
         self.set_duty(self.get_max_duty())
     }
@@ -84,6 +86,7 @@ pub trait SetDuty: ErrorType {
     ///
     /// The caller is responsible for ensuring that `num` is less than or equal to `denom`,
     /// and that `denom` is not zero.
+    #[inline]
     fn set_fraction(&mut self, num: u16, denom: u16) -> Result<(), Self::Error> {
         let duty = num as u32 * self.get_max_duty() as u32 / denom as u32;
         self.set_duty(duty as u16)
@@ -92,6 +95,7 @@ pub trait SetDuty: ErrorType {
     /// Set the duty cycle to `percent / 100`
     ///
     /// The caller is responsible for ensuring that `percent` is less than or equal to 100.
+    #[inline]
     fn set_percent(&mut self, percent: u8) -> Result<(), Self::Error> {
         self.set_fraction(percent as u16, 100)
     }
--- a/embedded-hal/src/pwm.rs
+++ b/embedded-hal/src/pwm.rs
@@ -1,0 +1,77 @@
+//! Pulse Width Modulation (PWM) traits
+
+/// Error
+pub trait Error: core::fmt::Debug {
+    /// Convert error to a generic error kind
+    ///
+    /// By using this method, errors freely defined by HAL implementations
+    /// can be converted to a set of generic errors upon which generic
+    /// code can act.
+    fn kind(&self) -> ErrorKind;
+}
+
+impl Error for core::convert::Infallible {
+    fn kind(&self) -> ErrorKind {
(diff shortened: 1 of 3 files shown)


embedded-hal/src/pwm.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// Single PWM channel / pin
+pub trait SetDuty: ErrorType {
+    /// Set the duty cycle.
+    ///
+    /// `duty` is the duty cycle. Valid values span the entire `u16` range:
+    ///
+    /// - `duty = 0` is considered 0% duty, which makes the pin permanently low.
+    /// - `duty = u16::MAX` is considered 1000% duty, which makes the pin permanently high.
```
rursprung, line 67:
```suggestion
    /// - `duty = u16::MAX` is considered 100% duty, which makes the pin permanently high.
```

ryankurte (approved):
seems broadly good to me!

> Duty is no longer an unconstrained associated type. u16 has been chosen because it gives enough dynamic range without being too big.

if we do need more (or less) resolution in the future it should be non-breaking to swap to a default associated type + markers (à la vec) and const MAX... i'd probably prefer this from the get-go, (and maybe there's some overlap with timer widths) but am okay either way.

GrantM11235:
This is the alternative that I mentioned during the meeting. It brings back the `get_max_duty` method and adds some handy convenience methods. I think that this makes it easier to use and it also should result in faster math with less rounding errors

```rust
pub trait Pwm {
    // Required methods
    fn get_max_duty(&self) -> u16;
    fn set_duty(&mut self, duty: u16);

    // Provided methods
    fn set_off(&mut self) {
        self.set_duty(0)
    }
    fn set_on(&mut self) {
        self.set_duty(self.get_max_duty())
    }
    fn set_fraction(&mut self, num: u16, denom: u16) {
        let duty = num as u32 * self.get_max_duty() as u32 / denom as u32;
        self.set_duty(duty as u16)
    }
    fn set_percent(&mut self, percent: u8) {
        self.set_fraction(percent as u16, 100)
    }
}
```

rursprung:
i like the `set_percent` feature from @GrantM11235, that'd cover nearly all my existing use-cases.
the only other use-case i have seen is when mapping angles (in degrees) to a PWM signal when driving a servo - obviously going from 0 - 180° (or even 0 - 359°) has a higher resolution than going from 0 - 100% (see [the arduino API for servos](https://www.arduino.cc/reference/en/libraries/servo/write/)). but that would of course be covered by the `set_fraction` method which can be used for this use-case.

burrbull:
@GrantM11235 All autoimplemented methods should be marked with `#[inline]`.

related: https://rust.godbolt.org/z/44bn77Msx

GrantM11235:
Some more notes in no particular order:

- It's important to stress that setting the duty to the max value means "on 100% of the time", not just 99.99%.
- My previous post was a simplified version that didn't contain all the necessary result/error stuff. In the real version, all methods should return a `Result` except for `get_max_duty` which should be infallible.
- If `set_duty` is called with a value greater than the max duty, the implementation is allowed do just about anything, as long as it isn't UB. Including but not limited to:
	- Clamp to max duty
	- Panic
	- Return an error
	- Set the duty to some other value (due to integer wrapping)
	- That's all I can think of
- The same thing applies to calling `set_fraction` with `num` > `denom` and calling `set_percent` with a value > 100.
- Calling `set_fraction` with a `denom` of zero will panic.

rursprung:
> * the implementation is allowed do just about anything, as long as it isn't UB

for drivers this might be an issue if they're passing on input - now they have to do the sanity check of the input values and the HAL might well do it again. i think it'd be good if this were specified.

by the way: would the API profit if it'd use something like [`UpperBounded`](https://docs.rs/num-traits/0.2.15/num_traits/bounds/trait.UpperBounded.html) from [`num-traits`](https://crates.io/crates/num-traits) or would that be overkill (or not even possible in this situation)? i have no personal experience with this crate and didn't think it through yet, but have seen it before (probably in some examples, don't remember) and it has 111 million downloads, so i guess it's somewhat mature? 😅

GrantM11235:
> > * the implementation is allowed do just about anything, as long as it isn't UB
> 
> for drivers this might be an issue if they're passing on input - now they have to do the sanity check of the input values and the HAL might well do it again. i think it'd be good if this were specified.

Some drivers will want to clamp to the max value, some drivers will want to return an error, and some (most?) drivers won't need to do any sanity checking at all if all possible inputs map to a valid duty value. Therefore the most efficient option is for each driver to implement their preferred behavior, and to not require any sanity checking in the HALs

> by the way: would the API profit if it'd use something like [`UpperBounded`](https://docs.rs/num-traits/0.2.15/num_traits/bounds/trait.UpperBounded.html) from [`num-traits`](https://crates.io/crates/num-traits)

That trait returns the maximum value that can be represented by a particular type. It doesn't help us in this case because the max duty value is not represented in the type system. If the max duty value was represented in the type system (for example with const generics), that would prevent dynamic configuration/reconfiguration of the timer period

adamgreig:
It might be nice to not allow panicking in drivers, just because unavoidable panic paths are pretty annoying. If the operation returns a Result anyway, couldn't it just return an error condition (which we could add to ErrorKind) for this situation? We could allow it to silently do something else like clamp or wrap or whatever if it likes.

I expect few implementations will be able to accept all u16 as valid duties, not least because insisting on max being 100% duty is likely to mean HALs will have to special-case that value anyway. I expect it's more likely HALs will do something like accept an 8-bit or 10-bit or whatever value and mask off whatever's supplied as the most-convenient option, or otherwise clamp?

GrantM11235:
I think that a bad duty value is a "garbage in, garbage out" situation. I think that HALs should be encouraged to use a `debug_assert!`, and otherwise not worry about it.

I don't think that there should be an ErrorKind for it because that would imply that all implementations need to check it.

Some food for thought, including:

- A pseudo-implementation for stm32 which shows that no special casing is required
- A moderately complex generic driver that uses a precomputed lookup table while also allowing the user to dynamically reconfigure the pwm period.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=47d88c1da07cfd679f66b6fe0db778ae

embedded-hal/src/pwm.rs
```diff
@@ -0,0 +1,124 @@
...
+        self.set_duty(self.get_max_duty())
+    }
+
+    /// Set the duty cycle to `num / denom`.
+    ///
+    /// The caller is responsible for ensuring that `num` is less than or equal to `denom`,
+    /// and that `denom` is not zero.
+    #[inline]
+    fn set_fraction(&mut self, num: u16, denom: u16) -> Result<(), Self::Error> {
+        let duty = num as u32 * self.get_max_duty() as u32 / denom as u32;
```
chrysn, line 91: Rather than leaving this to prose, why not use the core type? That way, the compiler will know not to handle a zero-division even when it can't const-propagate the denominator.

```suggestion
    /// The caller is responsible for ensuring that `num` is less than or equal to `denom`,
    /// and that `denom` is not zero.
    #[inline]
    fn set_fraction(&mut self, num: u16, denom: core::num::NonZeroU16) -> Result<(), Self::Error> {
        let duty = num as u32 * self.get_max_duty() as u32 / denom.get() as u32;
```

adamgreig: We discussed this at [this week's meeting](https://libera.irclog.whitequark.org/rust-embedded/2023-02-21) but didn't reach a consensus. Using NonZero ensures the function can't panic, but makes for a less ergonomic interface, especially since the majority of applications are likely to specify a constant denominator.

Dirbaio:
Discussed in today's meeting. Last bikeshed item: renamed "duty" to "duty cycle".

I think it should be good to go.

@rust-embedded/hal

burrbull (dismissed):

therealprof (approved):
Fine with me.

usbalbin:
In some cases one PWM timer may be able to adjust frequency on the run and also have multiple output channels with independent duty cycles. In those cases there may be one timer type responsible for the frequency or period and one or more types for the channels. Then the channels does not know what the current period is and thus can not know the max duty value.

With all of the `set_*` methods supposed to take `get_max_duty` into account, I believe this might complicate things when the period may be adjusted from another type potentially from another interrupt/thread/context.

Same thing the other way around, say the above was somehow resolved. If we managed to `ch1.set_percentage(50)`, what would happen the next time `timer.set_period(old_period * 2)` is called? What would that do to the duty? In many cases I would imagine that it would not stay at 50% which I think would be quite surprising 

Please correct me if I am wrong but I, atleast in the above case, I believe `get_max_duty` belongs more with the timer than with the channel. Thus it will have to be up to the user to adjust for that and we are stuck with using raw compare values for setting the duty. While this would put more responsibility on the user, i think it would be less surprising and more straight forward (in that particular case).

Again please let me know if I am missing something :)

Dirbaio:
The main goal of the traits is to support HAL-independent drivers. For example, a "RGB LED driver" would take 3 `impl SetDuty`s, one for each R/G/B channel.

If the traits did model "PWM peripheral" and "PWM output channel", the RGB LED driver would have to be aware of whether the chip configures max_duty per peripheral or per channel, whether different R/G/B channels are in the same PWM peripheral or not... This makes the trait harder/impossible to use for drivers without HAL compat issues. 

(The previous PWM trait in EH0.2 worked similarly to what you mention, it was scrapped due to this reason (and due to unconstrained Time types #324 too, but that's another story))

HALs can still implement the traits for chips where max_duty is per-PWM-peripheral instead of per-channel. There's a few options:
- Simply do not allow changing frequency after creating the PWM.
- Always report max_duty = `u16::MAX`. On `set_duty`, save the unscaled (`0..u16::MAX`) duty, scale it and apply it to the hardware. On freq change, apply max_duty to the hardare, scale all the saved unscaled duties again with the new max_duty, apply them again the hardware. Use Cell/RefCell/Mutex to share that data across the several channels.

The HALs don't necessarily have to use the `embedded-hal` traits as their main API. They can have a "native" API that exactly matches the hardware and has low level control, and then implement the traits in parallel, only to be used with HAL-independent drivers. So it's OK If the trait API needs some more heavy lifting from the HAL, or is less flexible, to ensure compatibility.

PR #431, opened 2022-12-20
gpio: add ErrorKind.

Dirbaio:
This adds an `ErrorKind` to the GPIO `Error` associated type, just like all the other drivers. It has just a single `Other` variant for now. 

The reason to have it is future-proofness: Even if we don't have ideas for ErrorKind variants for now, we might have them in the future.

- If we have `ErrorKind`, we can add more variants in the future backwards-compatibly after the 1.0 launch, thanks to `ErrorKind` being marked `#[non_exhaustive]`.
- If we *don't* have `ErrorKind`, adding it in the future would be a breaking change, because HALs would have to add support for it.

### DIFF
--- a/embedded-hal/CHANGELOG.md
+++ b/embedded-hal/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- gpio: add `ErrorKind` enum for consistency with other traits and for future extensibility. No kinds are defined for now.
 
 ## [v1.0.0-alpha.9] - 2022-09-28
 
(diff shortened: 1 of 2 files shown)


ryankurte (dismissed):
i'm _okay_ with this, though the idea behind error kind on other interfaces was to deal with bus-specific errors (eg. CAN contention, I2C NACK) that a driver / application may need to handle for normal operation, and i am not sure equivalents exist for GPIOs?

eldruin (commented):
From the top of my head, we may need something like a wrong-mode error if we add dynamic-mode GPIO, for example.

@Dirbaio Thanks! Could you add a changelog entry?

Dirbaio:
Done

therealprof (approved):
LGTM, thanks.

bors r+

PR #438, opened 2023-02-28
Bump MSRV to 1.59, switch to Edition 2021. Fix CI.

Dirbaio:
Bump MSRV to 1.59, switch to Edition 2021.

Fixes CI, It broke due to yanking `cortex-m-rt 0.71` due to the stack align bug.

~~I don't think the MSRV should matter for `dev-dependencies`? They're only required for running tests, bumping it doesn't affect crates using `embedded-hal` as a regular dep.~~ of course *out* CI requires it, duh.


jannic (dismissed):

Dirbaio:
> Is it really necessary to set edition = 2021 everywhere? Looks like an unnecessarily big change just to fix CI.

Once we bump MSRV, there's no reason not to do it.

The other option is to *not* bump MSRV, and pin `cortex-m-rt =0.7.0` instead. Though IMO if one of the major WG crates is 1.59+, there's not much point in keeping MSRV lower. Also 1.59 is 1 year old by now.

therealprof (approved):
LGTM, thanks.

bors r+

PR #439, opened 2023-02-28
Use github action dtolnay/rust-toolchain@master instead of actions-rs

jannic:
actions-rs/toolchain@v1 uses deprecated features and has not been updated for years

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -9,12 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Dirbaio:
The base GHA ubuntu image already contains `rustup`. I wonder if we could use that, with just shell commands to set the right rust version. What are the advantages of using a dedicated action?

jannic:
> The base GHA ubuntu image already contains `rustup`. I wonder if we could use that, with just shell commands to set the right rust version. What are the advantages of using a dedicated action?

I think the action applies some optimizations regarding github caching etc., but I don't know any details. That's just the minimal change I found to avoid the deprecation warnings.

Issue #367, opened 2022-02-18
I2C transaction_iter should take an iterator of mutable references


yodaldevoid:
I was recently porting an I2C controller device to `1.0.0-alpha.7` which forced me to implement `transaction` and `transaction_iter` for the first time. I initially tried to implement `transaction` as follows:
```rust
fn transaction<'a>(
    &mut self,
    address: u8,
    operations: &mut [Operation<'a>],
) -> Result<(), Self::Error> {
    self.transaction_iter(address, operations)
}
```
This, of course, did not work as `&mut [Operation<'a>]` produces an iterator with `Item = &mut Operation<'a>` rather than the `Item = Operation<'a>` that `transaction_iter` requires.
```rust
fn transaction_iter<'a, O>(&mut self, address: u8, operations: O) -> Result<(), Self::Error>
where
    O: IntoIterator<Item = Operation<'a>>,
```
---
In my opinion, `transaction_iter`'s signature should be this instead:
```rust
fn transaction_iter<'a, 'b: 'a, O>(&mut self, address: u8, operations: O) -> Result<(), Self::Error>
where
    O: IntoIterator<Item = &'a Operation<'b>>,
```
This signature mainly allows the implementation of `transaction` that I wrote above. By changing this signature almost all functions of `I2C` can be implemented as calls to `transaction_iter` even on systems without `alloc` support, with `write_iter` and `write_iter_read` being the exceptions. This would allow de-duplication of logic, which is rarely a bad thing. Additionally, I see no reason that `transaction_iter` needs an iterator over owned `Operations` in the first place.

Thoughts?

Dirbaio:
If the iterator yields owned Operations, it can generate them on the fly and return them. If it yields references, it has to allocate all the operations upfront and keep them alive throughout the entire `transaction_iter` call, as all the yielded references must live for `'a`.

However, even with the current "yields owned Operations", the same problem applies to the buffers, they must still be all allocated upfront. Given this I'm not sure `transaction_iter` has any benefits at all over `transaction` in the first place... Maybe we should remove it completely.

Also, the whole problem disappears if we change i2c to use a `transaction` closure like in #351...

yodaldevoid:
You make a good point about needing a way of yielding generated operations. You also make a good point that `transaction_iter` doesn't really meet this need at the moment.

I'll look at putting together a PR adding something like a what is seen in #351.

yodaldevoid:
[...]

PR #440, opened 2023-03-07
I2c: simplify, expand docs, document shared bus usage.

Dirbaio:
~Depends on #441 -- check that one out first.~

This does some simplifications to the trait that I think we should do:

- Implement all methods in terms of `transaction`. This way HALs have to implement just that.
- Removed byte-wise-iteration methods: `write_iter` and `write_iter_read`. The reason is that they're quite inefficient, especially with DMA implementations. We've already removed these on other traits, so I think we should do as well here.
- Removed `transaction_iter`. I don't think it's much useful in practice, because the way iterators work all the yielded `Operation`s must have the same lifetime. This means that, even if the user can generate the `Operation`s on the fly, they can't allocate buffers for these on the fly, all buffers must be pre-allocated. So it's not useful for, say, streaming a large transfer by reusing some small buffer repeatedly. See #367 
- Removed useless lifetimes
- Standardized buffer names on `read` and `write`, I think they're clearer.

It also specifies how i2c bus sharing is supposed to work. This is an alternative to #392 . After the discussions there, I don't think we should split I2C into Bus and Device anymore. For SPI it makes sense, because drivers want to enforce that there's a CS pin (`SpiDevice`) or not (`SpiBus`). This is not the case with I2C, the API is exactly the same in the shared and non-shared case. Drivers shouldn't care which case it is.

So all we have to do to "support" bus sharing is docs, This PR does:

- Document that it's allowed for implementations to be either shared or not.
- Document some guidelines for drivers and HALs on how to best use the traits, expand the examples.


embedded-hal/src/i2c.rs
```diff
@@ -14,128 +14,140 @@
...
+//!
+//! The [`embedded-hal-bus`](https://docs.rs/embedded-hal-bus) crate provides several
+//! implementations for sharing I2C buses. You can use them to take an exclusive instance
+//! you've received from a HAL and "split" it into mulitple shared ones, to instantiate
+//! several drivers on the same bus.
+//!
+//! # For driver authors
+//!
+//! Drivers can be generic over `I2c<SevenBitAddress>` or `I2c<TenBitAddress>` depending
+//! on the kind of address supported by the target device. If it can use either, the driver can
```
eldruin, line 44:
```suggestion
//! Drivers can select the adequate address length with `I2c<SevenBitAddress>` or `I2c<TenBitAddress>` depending
//! on the target device. If it can use either, the driver can
```

embedded-hal/src/i2c.rs
```diff
@@ -14,128 +14,140 @@
...
+//! be generic over the address kind as well, though this is rare.
+//!
+//! Drivers should take the `I2c` instance as an argument to `new()`, and store it in their
+//! struct. They **should not** take `&mut I2c`, the trait has a blanket impl for all `&mut T`
+//! so taking just `I2c` ensures the user can still pass a `&mut`, but is not forced to.
+//!
+//! Drivers **should not** try to enable bus sharing by taking `&mut I2c` at every method.
+//! This is much less ergonomic than owning the `I2c`, and allowing the user to pass an
```
eldruin, line 52:
```suggestion
//! This is much less ergonomic than owning the `I2c`, which still allows the user to pass an
```

embedded-hal/src/i2c.rs
```diff
@@ -14,128 +14,140 @@
...
+//! # For HAL authors
+//!
+//! HALs **should not** include bus sharing mechanisms. They should expose a single type representing
+//! exclusive ownership over the bus, and let the user use [`embedded-hal-bus`](https://docs.rs/embedded-hal-bus)
+//! if they want to share it. (One exception is if the underlying platform already
+//! supports sharing, such as Linux or some RTOSs.)
+//!
+//! Here is an example of an embedded-hal implementation of the `I2C` trait
+//! for both modes. All trait methods have have default implementations in terms of `transaction`,
+//! so that's the only one that requires implementing.
```
eldruin, line 115:
```suggestion
//! for both addressing modes. All trait methods have have default implementations in terms of `transaction`.
//! As such, that is the only method that requires implementation in the HAL.
```

eldruin (commented):
I agree with the analysis and looks good to me, thank you!
I just added some documentation improvements.

TODO:
- [x] update changelog

Dirbaio:
Done

eldruin (approved):
Alright, let's get this merged.
Thank you for your work!
bors r+

PR #443, opened 2023-03-14
spi: make SpiDevice transaction take an operation slice instead of a closure.

Dirbaio:
As discussed in [today's WG meeting](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$l3MzRbzKQ0LGOops0ioYfM2kEr6-eT7ODboDj9iiCfY?via=matrix.org&via=psion.agg.io&via=tchncs.de).

- Fixes https://github.com/rust-embedded/linux-embedded-hal/issues/87
- Gets rid of the cursed borrow issues in async.
- Makes the trait more consistent with I2C.

### DIFF
--- a/embedded-hal/src/spi.rs
+++ b/embedded-hal/src/spi.rs
@@ -46,20 +46,19 @@
 //! consists of asserting CS, then doing one or more transfers, then deasserting CS. For the entire duration of the transaction, the [`SpiDevice`]
 //! implementation will ensure no other transaction can be opened on the same bus. This is the key that allows correct sharing of the bus.
 //!
-//! The capabilities of the bus (read-write, read-only or write-only) are determined by which of the [`SpiBus`], [`SpiBusRead`] [`SpiBusWrite`] traits
-//! are implemented for the [`Bus`](SpiDevice::Bus) associated type.
+//! For read-only or write-only SPI devices, the [`SpiDeviceRead`] and [`SpiDeviceWrite`] are available.
 //!
 //! # For driver authors
 //!
 //! When implementing a driver, it's crucial to pick the right trait, to ensure correct operation
 //! with maximum interoperability. Here are some guidelines depending on the device you're implementing a driver for:
 //!
-//! If your device **has a CS pin**, use [`SpiDevice`]. Do not manually manage the CS pin, the [`SpiDevice`] implementation will do it for you.
-//! Add bounds like `where T::Bus: SpiBus`, `where T::Bus: SpiBusRead`, `where T::Bus: SpiBusWrite` to specify the kind of access you need.
+//! If your device **has a CS pin**, use [`SpiDevice`] (or [`SpiDeviceRead`]/[`SpiDeviceWrite`]). Do not manually
+//! manage the CS pin, the [`SpiDevice`] implementation will do it for you.
 //! By using [`SpiDevice`], your driver will cooperate nicely with other drivers for other devices in the same shared SPI bus.
 //!
 //! ```
-//! # use embedded_hal::spi::{SpiBus, SpiBusRead, SpiBusWrite, SpiDevice};
+//! # use embedded_hal::spi::{SpiBus, SpiBusRead, SpiBusWrite, SpiDevice, Operation};
 //! pub struct MyDriver<SPI> {
 //!     spi: SPI,
 //! }
@@ -67,7 +66,6 @@
 //! impl<SPI> MyDriver<SPI>
 //! where
 //!     SPI: SpiDevice,
-//!     SPI::Bus: SpiBus, // or SpiBusRead/SpiBusWrite if you only need to read or only write.
 //! {
 //!     pub fn new(spi: SPI) -> Self {
 //!         Self { spi }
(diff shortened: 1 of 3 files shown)


embedded-hal/src/spi.rs
```diff
@@ -323,71 +376,94 @@ pub trait SpiDevice: ErrorType {
...
-    {
-        self.transaction(|bus| bus.read(buf))
-    }
+    /// On bus errors the implementation should try to deassert CS.
+    /// If an error occurs while deasserting CS the bus error should take priority as the return value.
+    fn transaction(&mut self, operations: &mut [Operation<'_, Word>]) -> Result<(), Self::Error>;
 
     /// Do a transfer within a transaction.
     ///
     /// This is a convenience method equivalent to `device.transaction(|bus| bus.transfer(read, write))`.
```
eldruin, line 419:
```suggestion
    /// This is a convenience method equivalent to `device.transaction(&mut [Operation::Transfer(read, write)]`.
```

embedded-hal/src/spi.rs
```diff
@@ -323,71 +376,94 @@ pub trait SpiDevice: ErrorType {
...
-        Word: Copy,
-    {
-        self.transaction(|bus| bus.transfer(read, write))
+    fn transfer(&mut self, read: &mut [Word], write: &[Word]) -> Result<(), Self::Error> {
+        self.transaction(&mut [Operation::Transfer(read, write)])
     }
 
     /// Do an in-place transfer within a transaction.
     ///
     /// This is a convenience method equivalent to `device.transaction(|bus| bus.transfer_in_place(buf))`.
```
eldruin, line 428:
```suggestion
    /// This is a convenience method equivalent to `device.transaction([Operation::TransferInPlace(buf)]`.
```

eldruin (commented):
I think this is an interesting idea, thank you!
However, I do not have everything discussed in issues like #35, #351, #245 and #299.
Those should be considered for evaluation.

TODO:
- [ ] Update changelog

ryankurte:
hey seems reasonable to me! a couple of questions:

- how do we represent busses with actual hardware CS, pass through a `NullCS` of some sort?
- can we test this with a PR against `linux-embedded-hal` as well as another hal and a driver?

Dirbaio:
> how do we represent busses with actual hardware CS, pass through a NullCS of some sort?

implementations with hardware CS would directly implement `SpiDevice`, not `SpiBus`. You *dont* take a "SpiBus with hardware CS" and then wrap it on an ExclusiveDevice with NullCS.

eldruin (commented):
@Vollbrecht made an interesting comment about `read_transaction` but it seems only a mild HAL-impl-side inconvenience:
```rust
fn read_transaction(&mut self, operations: &mut [&mut [Word]]) -> Result<(), Self::Error>;
```

> Its a lot of duplication here from transaction. The problem is the transaction takes an &mut [Operation<'_, u8>] but this read_transaction is &mut [&mut [u8]]. We would need something that converts the &mut [u8] in an Operation::Read

It seems to me like this should be fine to implement on `linux-embedded-hal` but we should check on it.

ryankurte (approved):
aight thanks for all your work / bearing with me / lgtm!

note that one consequence of the move from closures is that it is no longer possible to represent devices that need to do things _within_ as CS assertion using `SpiDevice`, where this behaviour is required drivers should instead bind `SpiBus` with an `OutputPin` (ie. implement software/userspace CS control themselves)

Dirbaio:
> @eldruin: @Vollbrecht made an interesting comment about read_transaction but it seems only a mild HAL-impl-side inconvenience:

Yes, it's somewhat annoying but I think it's better than taking a `&[Operation]` and then asserting at runtime that it has only reads (or only writes). And I can't think of any other solution...

The duplication also shows up in #444 . It won't affect most HALs though, as they'll likely implement only `SpiBus` and let the user use `embedded-hal-bus` to get an `SpiDevice`. The only reasons a HAL would implement SpiDevice is for hardware CS (which is very annoying and the perf benefits are questionable in MCUs), or when the kernel handles CS for you (like in esp-idf or Linux).

> It seems to me like this should be fine to implement on linux-embedded-hal but we should check on it.

Yeah I think so too. It has to convert from either `&[Operation]` or `&[&[u8]]` to the linux ioctl struct anyway...

eldruin (approved):
We discussed this in this week's meeting.
As @ryankurte also mentioned, there are some things that are not possible anymore.
@GrantM11235 Came up with two interesting examples of these:
```rust
// with the slices approach we would need to create copies of the data
fn send_data(&mut self, data: &[u16]) {
    self.spi.transaction( |bus|
        for chunk in data.chunks(32) {
            let mut buffer = [0_u16; 32];
            let mut buffer = &mut buffer[0..chunk.len()];
            buffer.copy_from_slice(chunk);

            for word in buffer {
                *word = word.to_be();
            }

            let buffer = transmute_u16_to_u8(buffer);

            bus.send(buffer);
        }
    )
}
```
```rust
// this version is more efficient than what would be possible with slices
fn send_packet(&mut self, packet: &[u8]) {
    let mut control_flow = Continue;
    while control_flow == Continue {
        self.spi.transaction( |bus| {
            let mut first_byte = [COMMAND_SEND];
            bus.transfer_inplace(&mut first_byte);
            let [status_byte] = first_byte;

            if check_can_send(status_byte) {
                bus.write(packet);
                control_flow = Break;
            }
        });
        // maybe add a delay here
    }
}
```
The MAX3421E driver might benefit from the closure-based approach.

On balance, it is clear that the slice-based approach is simpler to understand, to implement and to use while probably covering > 90% of the use cases as well as resolving the problem with the linux implementation.
As such, let's take the slices-based approach for 1.0 and let's discuss re-adding a closure-based approach in a future release.

Thank you everybody for your work, insights and the fruitful discussions.

bors r+

PR #444, opened 2023-03-14
bus/spi: add RefCell, CriticalSection and Mutex shared bus implementations.

Dirbaio:
Requires #443 

This adds a few bus sharing implementations, with varying tradeoffs:
- `RefCellDevice`: single thread only
- `CriticalSectionDevice`: thread-safe, coarse locking, nostd.
- `MutexDevice`: thread-safe, fine-grained locking, std only.


eldruin:
I hit the button too quickly in #443. Could you add entries to the changelog for the changes in #443 here or add a new PR?

Dirbaio:
@eldruin done

eldruin (approved):
Thank you!
bors r+

PR #445, opened 2023-03-14
bus/i2c: add RefCell, CriticalSection and Mutex shared bus implementations.

Dirbaio:
Requires #440 

Same as #443 but for I2C.

This adds a few bus sharing implementations, with varying tradeoffs:
- `RefCellDevice`: single thread only
- `CriticalSectionDevice`: thread-safe, coarse locking, nostd.
- `MutexDevice`: thread-safe, fine-grained locking, std only.

### DIFF
--- a/embedded-hal-bus/src/i2c/critical_section.rs
+++ b/embedded-hal-bus/src/i2c/critical_section.rs
@@ -1,0 +1,70 @@
+use core::cell::RefCell;
+use critical_section::Mutex;
+use embedded_hal::i2c::{ErrorType, I2c};
+
+/// `critical-section`-based shared bus [`I2c`] implementation.
+///
+/// Sharing is implemented with a `critical-section` [`Mutex`](critical_section::Mutex). A critical section is taken for
+/// the entire duration of a transaction. This allows sharing a single bus across multiple threads (interrupt priority levels).
+/// The downside is critical sections typically require globally disabling interrupts, so `CriticalSectionDevice` will likely
+/// negatively impact real-time properties, such as interrupt latency. If you can, prefer using
+/// [`RefCellDevice`](super::RefCellDevice) instead, which does not require taking critical sections.
+pub struct CriticalSectionDevice<'a, T> {
+    bus: &'a Mutex<RefCell<T>>,
+}
+
+impl<'a, T> CriticalSectionDevice<'a, T> {
+    /// Create a new `CriticalSectionDevice`
+    pub fn new(bus: &'a Mutex<RefCell<T>>) -> Self {
+        Self { bus }
+    }
+}
+
+impl<'a, T> ErrorType for CriticalSectionDevice<'a, T>
+where
+    T: I2c,
+{
+    type Error = T::Error;
+}
+
+impl<'a, T> I2c for CriticalSectionDevice<'a, T>
+where
+    T: I2c,
+{
+    fn read(&mut self, address: u8, read: &mut [u8]) -> Result<(), Self::Error> {
+        critical_section::with(|cs| {
+            let bus = &mut *self.bus.borrow_ref_mut(cs);
+            bus.read(address, read)
+        })
+    }
+
+    fn write(&mut self, address: u8, write: &[u8]) -> Result<(), Self::Error> {
+        critical_section::with(|cs| {
+            let bus = &mut *self.bus.borrow_ref_mut(cs);
+            bus.write(address, write)
+        })
+    }
+
+    fn write_read(
+        &mut self,
+        address: u8,
+        write: &[u8],
+        read: &mut [u8],
+    ) -> Result<(), Self::Error> {
+        critical_section::with(|cs| {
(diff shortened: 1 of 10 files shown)


embedded-hal-bus/src/i2c/critical_section.rs
```diff
@@ -0,0 +1,70 @@
...
+        address: u8,
+        write: &[u8],
+        read: &mut [u8],
+    ) -> Result<(), Self::Error> {
+        critical_section::with(|cs| {
+            let bus = &mut *self.bus.borrow_ref_mut(cs);
+            bus.write_read(address, write, read)
+        })
+    }
+
```
eldruin, line 59:
```suggestion
```
We do not need any of these, right?

embedded-hal-bus/src/i2c/mutex.rs
```diff
@@ -0,0 +1,59 @@
+use embedded_hal::i2c::{ErrorType, I2c};
+use std::sync::Mutex;
+
+/// `std` `Mutex`-based shared bus [`I2c`] implementation.
+///
+/// Sharing is implemented with a `std` [`Mutex`](std::sync::Mutex). It allows a single bus across multiple threads,
```
eldruin, line 6:
```suggestion
/// Sharing is implemented with an `std` [`Mutex`](std::sync::Mutex). It allows a single bus across multiple threads,
```
I think

embedded-hal-bus/src/i2c/mutex.rs
```diff
@@ -0,0 +1,59 @@
...
+/// with finer-grained locking than [`CriticalSectionDevice`](super::CriticalSectionDevice). The downside is
```
eldruin, line 7:
```suggestion
/// with finer-grained locking than [`CriticalSectionDevice`](super::CriticalSectionDevice). The downside is that
```
A bit clearer

embedded-hal-bus/src/i2c/mutex.rs
```diff
@@ -0,0 +1,59 @@
...
+/// it is only available in `std` targets.
+pub struct MutexDevice<'a, T> {
+    bus: &'a Mutex<T>,
+}
+
+impl<'a, T> MutexDevice<'a, T> {
+    /// Create a new Mutexdevice
```
eldruin, line 14:
```suggestion
    /// Create a new `MutexDevice`
```

embedded-hal-bus/src/i2c/mutex.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn write_read(
+        &mut self,
+        address: u8,
+        write: &[u8],
+        read: &mut [u8],
+    ) -> Result<(), Self::Error> {
+        let bus = &mut *self.bus.lock().unwrap();
+        bus.write_read(address, write, read)
+    }
+
```
eldruin, line 50:
```suggestion
```
Again, we should not need any of these, right?

embedded-hal-bus/src/i2c/refcell.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn write_read(
+        &mut self,
+        address: u8,
+        write: &[u8],
+        read: &mut [u8],
+    ) -> Result<(), Self::Error> {
+        let bus = &mut *self.bus.borrow_mut();
+        bus.write_read(address, write, read)
+    }
+
```
eldruin, line 50:
```suggestion
```
Same

embedded-hal-bus/src/i2c/critical_section.rs
```diff
@@ -0,0 +1,70 @@
...
+/// the entire duration of a transaction. This allows sharing a single bus across multiple threads (interrupt priority levels).
+/// The downside is critical sections typically require globally disabling interrupts, so `CriticalSectionDevice` will likely
+/// negatively impact real-time properties, such as interrupt latency. If you can, prefer using
+/// [`RefCellDevice`](super::RefCellDevice) instead, which does not require taking critical sections.
+pub struct CriticalSectionDevice<'a, T> {
+    bus: &'a Mutex<RefCell<T>>,
+}
+
+impl<'a, T> CriticalSectionDevice<'a, T> {
+    /// Create a new Mutexdevice
```
eldruin, line 17:
```suggestion
    /// Create a new `CriticalSectionDevice`
```

embedded-hal-bus/src/i2c/critical_section.rs
Dirbaio: They could all go through the `transaction` default impl, yes. However, some HAL might implement `read/write/write_read` with some optimizations. Forwarding all methods keeps them going through `transaction` would lose them.

eldruin: I see, ok.

eldruin (approved):
Looks good to me, thank you!
bors r+

PR #448, opened 2023-03-28
Misc doc fixes.

Dirbaio:
- Switch from GAT to AFIT has already happened.
- We're no longer planning to integrate EHA into EH
- Fix all broken links.


eldruin (approved):
Looks good to me, thank you!
bors r+

PR #456, opened 2023-05-02
spi: clarify flushing/blocking behavior.

### LINKED ISSUE #455: Clarify flushing/"blocking" behavior
The `SpiDevice` trait's documentation is now very explicit that operations may return before they are finished. That brings up some questions:
1. What happens with adjacent operations? E.g. if there are two back-to-back `write()` calls, is it legal for the second call to return `Err(Busy)`, or is it expected that this call will block until it can enqueue the operation?
2. In the same vein, is calling `write(); write();` acceptable for drivers, or does it have to be `write(); flush(); write()` to not risk an error?
3. What about other traits like I2C? They do not mention that returning early is acceptable. Does that imply that they block until the bus is idle? If not, why don't they have a `flush()` operation. If so, shouldn't that be explicitly documented?

Dirbaio:


eldruin:
I like this addition, thank you!
However, I do not think it completely fixes #455 as that also refers to a similar situation for I2C.

eldruin (approved):
Thank you!
bors r+

PR #458, opened 2023-05-03
bus: Add a doc example for i2c::RefCellDevice

tomgilligan:
`embedded-hal-bus` is turning out to be very useful for something I'm working on.  It wasn't too hard to work out how to use it but it feels like an example would still be helpful here.  Not sure:

- is this the best place to put such an example?
- how much of the example should be hidden with `#`?


Dirbaio:
The example looks nice! :+1:. I'd put it on the struct instead of on `new()`, so that it's more visible.

Dirbaio:
Could you check the CI failure?

Dirbaio (approved):
bors r+

PR #461, opened 2023-05-10
spi: remove write-only and read-only traits.

Dirbaio:
When introducing the Bus/Device split in #351, I kept the ability to represent "read-only" and "write-only" buses, with separate `SpiBusRead`, `SpiBusWrite` buses. This didn't have much discussion, as it was the logical consequence of keeping the separation in the already existing traits (`Read`, `Write`, `Transfer`, ...).

Later, in #443, when switching from the closure-based API to the operation-slice-based API, this required adding `SpiDeviceRead`, `SpiDeviceWrite` as well, because you can no longer put a bound on the underlying bus with the new API.

This means we now have *seven* traits, which we can reduce to *two* if we drop the distinction. So, is the distinction worth it?

I've always been on the fence about it, now I'm sort of leaning towards no.

First, using write-only or read-only SPI is rare. 
- write-only SPI: for SPI displays you don't want to read from, or to abuse it to bitbang stuff like ws2812b,
- read-only SPI: some ADCs that can't be configured at all, you just clock out bits. Or even weirder abuses, like to build a logic analyzer

Second, for it to be useful HALs have to track "read-onlyness / write-onlyness" in the type signature, so a read-only SPI really only implements `SpiBusRead` and not `SpiBus`. HALs already have enough generics. For example, Embassy HALs don't implement the distinction (you can create MOSI-only or MISO-only SPIs, but they all impl the full `SpiBus`, because I didn't want to make the types carry pin information).

Third, it makes the API easier to use. Simpler, users only have to import one trait, docs have all the methods in one place. Much less boilerplate to impl the traits (look at how shorter `embedded-hal-bus` gets!).

So I think we should remove them.

### DIFF
--- a/embedded-hal-async/CHANGELOG.md
+++ b/embedded-hal-async/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Removed
+- spi: removed read-only and write-only traits.
+
 ## [v0.2.0-alpha.1] - 2023-04-04
 
 ### Added
--- a/embedded-hal-async/src/spi.rs
+++ b/embedded-hal-async/src/spi.rs
@@ -8,96 +8,19 @@ pub use embedded_hal::spi::{
     Error, ErrorKind, ErrorType, Mode, Operation, Phase, Polarity, MODE_0, MODE_1, MODE_2, MODE_3,
 };
 
-/// SPI read-only device trait
-///
-/// `SpiDeviceRead` represents ownership over a single SPI device on a (possibly shared) bus, selected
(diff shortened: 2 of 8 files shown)


almindor (approved):
LGTM

ryankurte (approved):
defs a fan of simplification, seems good to me.

Vollbrecht:
spi.rs https://github.com/esp-rs/esp-idf-hal/blob/master/src/spi.rs#L1337 well i can live with reducing that line count :smiley_cat:

eldruin (approved):
I agree with the assessment, thank you!

Dirbaio:
let's merge then! :D

bors r+

PR #462, opened 2023-05-10
spi: add Operation::DelayUs(u32).

Dirbaio:
depends on #461

There are SPI devices out there that need delays between transfers within a transaction. *Not* dummy bytes, SCK must be idle during the delay. And it must be a single transaction, CS must not be deasserted.

This is not doable with the new operation-list API (it was doable with the previous closure API).

Examples:

- ADS1255/56 needs a 7 micro second delay between the command to read a register and reading the corresponding data over SPI. [datasheet](https://www.ti.com/lit/ds/symlink/ads1256.pdf?ts=1683244329972&ref_url=https%253A%252F%252Fwww.ti.com%252Fproduct%252Fes-mx%252FADS1256) page 34. [matrix discussion](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$b97m_oLuXJs7vzzmV395FEyq4gUr1pO1Via9cshnFQ0?via=matrix.org&via=psion.agg.io&via=tchncs.de).
-  CAM312 Central Tracking Unit [datasheet](https://www.cambridgeic.com/resources/downloads/39-033-0068-cam312/file), Section 9.3. [matrix discussion](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$Cv2muTA878kdwufyM3ZovTFOw65g5dmk1qgiH-1wcJQ?via=matrix.org&via=psion.agg.io&via=tchncs.de)

To support these, this adds a new `Operation::DelayUs(u32)`.

Linux spidev can implement this fine, it [supports](https://github.com/torvalds/linux/blob/master/include/uapi/linux/spi/spidev.h#L42) delaying after each transfer.

The downside is that `SpiDevice` impls now need a time source. [Discussed a bit](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$DQFvuFyeU2T4ikBj_uB16wYoLFDvk5wes7gZrGKUDmU?via=matrix.org&via=psion.agg.io&via=tchncs.de) in yesterday's WG meeting.

I don't think separating the traits by capability is worth it, we end up with the same problem as with the read-only/write-only traits (#461)

### DIFF
--- a/embedded-hal-bus/src/spi/mutex.rs
+++ b/embedded-hal-bus/src/spi/mutex.rs
@@ -1,3 +1,4 @@
+use embedded_hal::delay::DelayUs;
 use embedded_hal::digital::OutputPin;
 use embedded_hal::spi::{ErrorType, Operation, SpiBus, SpiDevice};
 use std::sync::Mutex;
@@ -12,19 +13,36 @@ use super::DeviceError;
 /// Sharing is implemented with a `std` [`Mutex`](std::sync::Mutex). It allows a single bus across multiple threads,
 /// with finer-grained locking than [`CriticalSectionDevice`](super::CriticalSectionDevice). The downside is
 /// it is only available in `std` targets.
-pub struct MutexDevice<'a, BUS, CS> {
+pub struct MutexDevice<'a, BUS, CS, D> {
     bus: &'a Mutex<BUS>,
     cs: CS,
+    delay: D,
 }
 
-impl<'a, BUS, CS> MutexDevice<'a, BUS, CS> {
+impl<'a, BUS, CS, D> MutexDevice<'a, BUS, CS, D> {
     /// Create a new ExclusiveDevice
-    pub fn new(bus: &'a Mutex<BUS>, cs: CS) -> Self {
-        Self { bus, cs }
+    pub fn new(bus: &'a Mutex<BUS>, cs: CS, delay: D) -> Self {
+        Self { bus, cs, delay }
     }
 }
 
-impl<'a, BUS, CS> ErrorType for MutexDevice<'a, BUS, CS>
+impl<'a, BUS, CS> MutexDevice<'a, BUS, CS, super::NoDelay> {
+    /// Create a new MutexDevice without support for in-transaction delays.
+    ///
+    /// # Panics
+    ///
+    /// The returned device will panic if you try to execute a transaction
+    /// that contains any operations of type `Operation::DelayUs`.
+    pub fn new_no_delay(bus: &'a Mutex<BUS>, cs: CS) -> Self {
+        Self {
+            bus,
+            cs,
+            delay: super::NoDelay,
+        }
+    }
+}
+
+impl<'a, BUS, CS, D> ErrorType for MutexDevice<'a, BUS, CS, D>
 where
     BUS: ErrorType,
     CS: OutputPin,
@@ -32,10 +50,11 @@ where
     type Error = DeviceError<BUS::Error, CS::Error>;
 }
 
-impl<'a, Word: Copy + 'static, BUS, CS> SpiDevice<Word> for MutexDevice<'a, BUS, CS>
+impl<'a, Word: Copy + 'static, BUS, CS, D> SpiDevice<Word> for MutexDevice<'a, BUS, CS, D>
 where
     BUS: SpiBus<Word>,
     CS: OutputPin,
+    D: DelayUs,
 {
     fn transaction(&mut self, operations: &mut [Operation<'_, Word>]) -> Result<(), Self::Error> {
         let bus = &mut *self.bus.lock().unwrap();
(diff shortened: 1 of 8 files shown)


ryankurte:
hmm, is this a similar problem to the need to be able to do weird acknowledgement things within a transaction?  iirc for those cases we'd recommend using the bus instead of the device abstraction (which works for most cases with the usual caveats related to manual CS with multiple apps sharing buses under linux)

jannic:
Could there be SpiDevice implementations that can't support the delay operation? Perhaps there should be some documented best practice on how to react in that situation.

For example:
- drivers should avoid the delay operation if possible, to improve compatibility
- SpiDevice implementations should support the delay operation if possible, for the same reason
- in case an SpiDevice can't support the delay operation, it should reject a transaction containing such an operation. (It could return an error with a newly defined ErrorKind `OperationNotSupported`?) It should return that error immediately, even if `Operation::DelayUs` is not the first operation

With an `OperationNonSupported` error kind, one could even make `Operation` `non_exhaustive` to allow for later extensions. Who knows, perhaps there are devices which need varying SPI clocks within a single transaction?

embedded-hal-bus/src/spi/refcell.rs
```diff
@@ -12,30 +13,48 @@ use super::DeviceError;
...
 }
 
-impl<'a, BUS, CS> ErrorType for RefCellDevice<'a, BUS, CS>
+impl<'a, BUS, CS> RefCellDevice<'a, BUS, CS, super::NoDelay> {
+    /// Create a new RefCellDevice without support for in-transaction delays.
+    ///
+    /// # Panics
+    ///
+    /// The returned device will panic if you try to execute a transaction
+    /// that contains one `Operation::Delay`.
```
adamgreig, line 35:
```suggestion
    /// that contains `Operation::DelayUs`.
```

embedded-hal-bus/src/spi/mutex.rs
```diff
@@ -12,30 +13,48 @@ use super::DeviceError;
...
 }
 
-impl<'a, BUS, CS> ErrorType for MutexDevice<'a, BUS, CS>
+impl<'a, BUS, CS> MutexDevice<'a, BUS, CS, super::NoDelay> {
+    /// Create a new MutexDevice without support for in-transaction delays.
+    ///
+    /// # Panics
+    ///
+    /// The returned device will panic if you try to execute a transaction
+    /// that contains one `Operation::Delay`.
```
adamgreig, line 35:
```suggestion
    /// that contains `Operation::DelayUs`.
```

Dirbaio: Just removing the "one" sounds a bit weird to me. Changed it to "that contains at least one Operation::DelayUs", hopefully that's fine?

adamgreig: Saying "at least" sounds like a minimum _requirement_ to me, for example I'd be happy with "succeeds if it contains at least one Delay", but not "fails if it contains at least one Delay". I'd maybe say "that contains one or more Delay operations", but needing it to be plural makes it a bit awkward, so my first call would just be "that contains Delay".

eldruin (commented):
I think this might make `embedded-hal-bus` unusable for `linux-embedded-hal` (or any implementation which provides meaningful buffering of the SPI transfers).

While this would work for `SpiDevice` implementations provided by `l-e-h`, I think it would not work for its `SpiBus` implementation, since delays there are [appropriately marked transfers](https://github.com/torvalds/linux/blob/master/include/uapi/linux/spi/spidev.h#L42) to the kernel and not something handled externally without `SpiBus` knowing about it.

This means that  when using `e-h-b` on a `l-e-h::SpiBus`, we are delaying in "program-defined time", not in "bus-transfer-defined time".
Depending on the internal queue fill-up status of the SPI handler in the kernel, we might delay for a _shorter_ time than requested to, down to no delay at all.

However, I do not know how the linux kernel buffers SPI transfers, and if there is a queue at all, so somebody with knowledge of the matter should clarify.
Nevertheless, the same issue applies to any HAL implementation which provides some kind of buffering due to the fundamental time difference.

Dirbaio:
Good point about buffering/queueing. I've added `bus.flush()` calls before the delays. This ensures the queued operations are fully finished before delaying, so the delay now becomes bus time, not just program time.

(Note that supporting `flush()` was already mandatory for `SpiBus`, and is also needed for correct software CS. So we're not ruling out any impls by doing this. Also I'm pretty sure linux spidev doesn't buffer transfers, or if it does then there must be a way to flush it for sure)

Dirbaio:
friendly ping @eldruin @ryankurte @therealprof

eldruin (approved):
Alright, I checked a bit over at `l-e-h` and this seems fine although we would not be able to use the kernel API for the delay when using `e-h-b`, but for that we would need to make the `e-h::SpiBus` know about delays and that would further the impact of this special case, without a clear benefit. When absolutely needed, it would be possible to define a different implementation for `e-h-b`. Let's get it merged.
Thank you!

bors r+

PR #463, opened 2023-05-31
embedded-hal: digital: add #[inline] hints for PinState functions

luojia65:
Add inline hints for:

- `<PinState as Not>::not`
- `<PinState as From<bool>>::from`
- `<bool as From<PinState>>::from`

Those hints would allow further optimizations if PinState structure functions is used on embedded-hal projects.


Dirbaio (approved):
Thank you!

bors r+

PR #470, opened 2023-07-12
Add `+ ?Sized` to all blanket impls.

Dirbaio:
It doesn't make much of a difference in practice because unsized types are rare, but it's best practice to do so, so why not.


eldruin (approved):
Seems fine to me, thanks!

PR #471, opened 2023-07-12
io: add error kinds from std.

Dirbaio:
depends on #466 

This copies all stable variants from `std::io::ErrorKind`, except:

- `WouldBlock` is removed, since `embedded-io` traits are always blocking. See the [crate-level documentation](crate) for details.
- `WriteZero` is removed, since it is a separate variant in [`WriteAllError`] and [`WriteFmtError`].


eldruin (approved):
Seems fine to me, thanks!

PR #473, opened 2023-07-12
Add bus access methods to `ExclusiveDevice`

### LINKED ISSUE #467: SPI `ExclusiveDevice`: access the bus object
Sometimes it might be necessary/useful to access the SPI bus object that is held by an `ExclusiveDevice`. One such case might be adjusting SCLK speed based on some data read from the SPI device.

bugadani:
Closes #467

I've chosen `bus` and `bus_mut` instead of `inner{_mut}` because ExclusiveDevice also wraps CS and Delay, so "inner" isn't very descriptive here. I've also opted to not implement `into_inner` for the same reason, and because a destructuring conversion would need to return both the bus, the chip select and the delay, too. So in summary, this PR is the minimal, hopefully non-controversial subset of additions.

### DIFF
--- a/embedded-hal-async/CHANGELOG.md
+++ b/embedded-hal-async/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
LGTM, thank you!

PR #474, opened 2023-07-15
spi: move async ExclusiveDevice to embedded-hal-bus.

Dirbaio:
!! REQUIRES #472 (for the MSRV bump and CI fixes) !!

Same logic as to why the blocking bus impls are in a separate crate: if we ever want to stabilize `embedded-hal-async` 1.0 we can't make breaking changes to ExclusiveDevice, so it's best to keep it in a separate crate so we can.


eldruin (approved):
Good idea, thank you!

bugadani:
To be quite honest, `ExclusiveDevice` is both the sync and the async implementor, so including that under the `asynch` module is a bit weird. It was equally weird to have it in `embedded-hal-async` but that's the past. I feel like the type could be moved up from `asynch::spi` to `spi`.

PR #475, opened 2023-07-15
Missing std io error impls for `ReadExactError` & `WriteAllError` 

MabezDev:
Currently depends on changes from https://github.com/rust-embedded/embedded-hal/pull/472.

I don't think the move of `to_std_error` to embedded-io will cause any stabilization concerns, it seems unlikely that that signature would change. If it is, I can make this hidden within embedded-io and we can duplicate the function for the adapters crate.


Dirbaio:
with the From impls I added, `to_std_error` becomes 1 line: `std::io::Error::new(err.kind().into(), format!("{:?}", err))` . With that I think it's less critical to share it, I'd rather not expose it as public API.

MabezDev:
Ah, good point! Changed :).

Dirbaio:
Incorporated into #472. Thank you! :)

PR #477, opened 2023-07-18
Merge spi `ExclusiveDevice` implementations

bugadani:
As a somewhat counter-proposal to #474 this PR merges the two different ExclusiveDevice implementations. I don't think there is any need in keeping them separate, especially when the async version implements the blocking SpiDevice trait. The separation also meant that bus access was implemented for the async version, but not the blocking-only one. This has now been fixed.


Dirbaio (approved):
I agree it makes more sense this way, yes.

My reasoning to make a separate `asynch` module was that we'll likely add async versions of the shared bus adapters. For those it's harder to make the same adapter implement both blocking and async (Tokio can thanks to std threads, but embassy-sync, rtic-sync can't). However there's no "standard" async mutex currently so perhaps that's best to ecosystem crates (like embassy-embedded-hal for example)

PR #486, opened 2023-08-17
Fix typo in embedded-io changelog

daniel-larsen:
Looks like a typo to me as ErrorKind is not a trait.


Dirbaio (approved):
nice catch, thanks!

PR #488, opened 2023-08-23
lib: amend documents to add periods (.), add inline hints when necessary

luojia65:
In documentation style of the Rust standard library, first sentence of all modules, types, and functions documentation has a period. We follow Rust standard library style to make it easier for users to read.

Most of functions in embedded-hal{,-async,-bus,-nb} (especially type conversations, function fowarding calls) are now marked #[inline] to allow further optimizations.

Those changes affects embedded-hal, embedded-hal-async, embedded-hal-bus and embedded-hal-nb.

Note: there are multiple commits for each commit changes an amount of files of certain module. If that's not the case, please reply and I'll squash the commits.

### DIFF
--- a/embedded-hal-async/CHANGELOG.md
+++ b/embedded-hal-async/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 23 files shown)


Dirbaio (approved):
Thank you!

PR #490, opened 2023-08-26
Use rust-version field

Sh3Rm4n:
I've noticed that this project does not use the [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) field yet, while maintaining a MSRV policy. Add this field to make
this policy machine-readable and leverage nicer cargo feedback in case of an older rust version.


Dirbaio:
`rust-version` actively forbids using the crate in lower version. I'm not sure if this is something we want to do. 

The 1.60 MSRV is due to the use of `dep:` in Cargo.toml. Only `embedded-hal-bus` and `embedded-io-adapters` actually use it, so in theory other crates like `embedded-hal` could work on 1.59 or older. So, adding `rust-version` is overly strict. 

Options:

- Set `rust-version` to 1.60 anyway even if it's too strict. Perhaps it's fine, 1.60 is almost 1.5 years old by now.
- Track MSRV for each crate individually, set `rust-version` to that.
- Keep tracking a single MSRV for the entire repo (i.e 1.60) but not add `rust-version`. So crates that don't actually need 1.60 still work on older Rusts, but it's not officially supported by the `embedded-hal` project.

I'm not sure what's the best solution here.

rursprung:
FYI: if you want to identify the MSRV for each crate (to set it now initially) you can use [cargo-msrv](https://crates.io/crates/cargo-msrv) to do that. i think with that it's then also no hassle to update it for the individual crates when needed

Sh3Rm4n:
I've reduced the MSRV for all crates, which do not use `dep:` to 1.56, which is the 2021 edition release.

rursprung:
i think this should land before v1.0 is released - it'd help with cleaner error messages in the future when the MSRV for a crate is increased. the rust/crates teams also encourages setting this field (see various posts in their forums & issues)

Dirbaio:
Doing option 2 ("Track MSRV for each crate individually, set rust-version to that.") sounds good to me then. We can discuss it next weekly meeting and merge if there's no objections.

Dirbaio (approved):
Discussed in today's meeting, conclusion is track MSRV per crate separately.

Async stuff will have to be `rust-version = 1.75`.

I'll add CI to test the different MSRVs in a follow-up PR.

PR #491, opened 2023-08-28
Implement from custom write errors for Infallible

rmja:
This PR will allow io middleware to constrain their `Write` implementation to a restricted error type that can convert from the new `WriteAllError`/`WriteFmtError` types. One example is the proposal over here: https://github.com/rmja/buffered-io/blob/constrained-write-error/src/asynch/write.rs#L28-L30

Specifically, it allows for easy use of e.g. Vec in tests which has the `Infallible` error type.

For reference, see the discussions here:
* https://github.com/drogue-iot/reqwless/pull/39
* https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$ghiVYOEf6MqDoMNFUiOQmcW82cHPjVeKsVBPGRIGoZ0?via=matrix.org&via=grusbv.com&via=tchncs.de

And, please, let me know if there is anything obvious that I am missing that could easy the transition to handling these two new error types.

cc @lulf @Dirbaio @MabezDev

### DIFF
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -3,7 +3,7 @@
 #![warn(missing_docs)]
 #![doc = include_str!("../README.md")]
 
-use core::fmt;
+use core::{convert::Infallible, fmt};
 
 // needed to prevent defmt macros from breaking, since they emit code that does `defmt::blahblah`.
 #[cfg(feature = "defmt-03")]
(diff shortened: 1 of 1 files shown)


eldruin (commented):
Seems reasonable to me, thanks.
Would need an entry to the Changelog, though.

What do you think @Dirbaio ?

Dirbaio:
an `impl Write<Error=Infallible>` can still return 0 from `write()`, which would cause a `WriteAllError::WriteZero`. So the panics here are not unreachable.

I think it's not idiomatic for `From` impls to panic. `TryFrom` would be appropriate here, though that doesn't work with `?` which I guess is your goal.

rmja:
Given the discussion on the previous wg meeting where I believe the consensus was to revert the introduction of the `WriteAllError` type in favor of disallowing `Ok(0)` from `write()`, this PR gets superfluous.

PR #500, opened 2023-09-05
Make embedded-io MSRV 1.48.0

RCasatta:
Would it be acceptable to make `embedded-io` crate MSRV 1.48.0 (without features) ?
It seems minimal changes are needed.

The reason to propose this is that rust-bitcoin crate is looking for a replacement for core2 crate, but at the moment has this lower MSRV required for dependencies.

Thanks

### DIFF
--- a/embedded-io/Cargo.toml
+++ b/embedded-io/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "embedded-io"
 version = "0.5.0"
-edition = "2021"
+edition = "2018"
 description = "Embedded IO traits"
 repository = "https://github.com/rust-embedded/embedded-hal"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


Dirbaio:
Rust 1.48 is *ancient*, almost 3 years old. This requires down

bugadani:
Not sure how much my opinion matters here, but the docs issue may be improved by introducing a `lower-my-msrv-pls` feature that disables the attribute with `include_str!`, but it's an inelegant hack at best.

jcdickinson:
You can use git, and hence a fork, for dependencies in Rust. For example:

```toml
embedded-hal = { git = "https://github.com/rust-bitcoin/embedded-hal.git", rev = "abc123..." }
```

Issue #501, opened 2023-09-05
embedded_io's WriteAllError


adamgreig:
Currently, `embedded_io::Write::write()` returns `Ok(0)` to indicate "did not write any bytes and cannot write any more", for example due to a closed socket or EOF or other situation that is not "an error" but nevertheless means no more data can ever be written.

The implementation of `Write::write_all()` must detect this situation and return early, otherwise it would block forever trying to write more data. To signal this condition, `write_all()` returns `Result<(), WriteAllError<E>>`, where `WriteAllError<E>` is an enum that can indicate either that the sink returned `Ok(0)` or that it returned some other error `E`, specific to the writer type.

This makes it awkward to write middleware which needs to call `write_all()`, as each middleware layer then needs to implement their own wrapping error type rather than passing up the error type used by the underlying writer.

We discussed this in the [wg meeting on 2023-08-29](https://libera.irclog.whitequark.org/rust-embedded/2023-08-29#1693333889-1693337191;) and concluded the best option is to change the documentation for `write()` to prohibit returning `Ok(0)` and instead signal the "no more data can be written" condition as an error instead. This allows `write_all()` to return `Result<(), E>` directly.

The current behaviour matches `std::io`, but in `std` the returned error is Boxed which avoids the wrapping problem we encounter. By changing `write()`'s behaviour we diverge from `std`, but it is still possible to maintain the implementations of `embedded_io` for `std` traits.

If anyone has any further thoughts on the change, please discuss here; otherwise if anyone would like to prepare a PR implementing this change please go ahead.

Dominaezzz:
> This makes it awkward to write middleware which needs to call `write_all()`, as each middleware layer then needs to implement their own wrapping error type rather than passing up the error type used by the underlying writer.

Initially I wasn't sure what you meant by needing to make a wrapping error type, since `write_all` calling an underlying `write_all` can just forward the error as is, which is what I have done in some middleware I've written as well.
However after reading the chat/meeting I now see you're trying to call `write_all` in `flush`, which make the problem more obvious.

For reference this is the implementation of `flush` for the `BufferedWrite` struct. 
```
    async fn flush(&mut self) -> Result<(), Self::Error> {
        if self.pos > 0 {
            self.inner.write_all(&self.buf[..self.pos]).await?;
            self.pos = 0;
        }

        self.inner.flush().await
    }
```

[...]

PR #505, opened 2023-09-13
Clarify write() may only return Ok(0) if data is empty

adamgreig:
Provisional fix for #501.

I'm not totally sure it makes sense to re-add `WriteZero` to `ErrorKind` since we're now using it for something a bit different to `std`, but since we do ask implementations to return an error instead of `Ok(0)`, I guess that error should have a type.

### DIFF
--- a/embedded-io/src/impls/slice_mut.rs
+++ b/embedded-io/src/impls/slice_mut.rs
@@ -1,9 +1,40 @@
-use crate::{ErrorType, Write};
+use crate::{Error, ErrorKind, ErrorType, Write};
 use core::mem;
 
-impl ErrorType for &mut [u8] {
-    type Error = core::convert::Infallible;
+// needed to prevent defmt macros from breaking, since they emit code that does `defmt::blahblah`.
+#[cfg(feature = "defmt-03")]
+use defmt_03 as defmt;
+
+/// Errors that could be returned by `Write` on `&mut [u8]`.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt-03", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SliceWriteError {
+    /// The target slice was full and so could not receive any new data.
+    Full,
 }
+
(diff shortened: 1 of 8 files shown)


embedded-io/src/lib.rs
```diff
@@ -401,11 +361,12 @@ pub trait Write: ErrorType {
     /// implementation to write an amount of bytes less than `buf.len()` while the writer continues to be
     /// ready to accept more bytes immediately.
     ///
-    /// Implementations should never return `Ok(0)` when `buf.len() != 0`. Situations where the writer is not
-    /// able to accept more bytes and likely never will are better indicated with errors.
+    /// Implementations must never return `Ok(0)` when `buf.is_empty()`. Situations where the writer is not
```
Dirbaio, line 364:
```suggestion
    /// Implementations must never return `Ok(0)` when `!buf.is_empty()`. Situations where the writer is not
```

embedded-io/src/impls/slice_mut.rs
```diff
@@ -1,23 +1,56 @@
...
+// needed to prevent defmt macros from breaking, since they emit code that does `defmt::blahblah`.
+#[cfg(feature = "defmt-03")]
+use defmt_03 as defmt;
+
+/// Errors that could be returned by `Write` on `&mut [u8]`.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt-03", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SliceWriteError {
+    /// The target slice was full and so could receive any new data.
```
Dirbaio, line 13:
```suggestion
    /// The target slice was full and so couldn't receive any new data.
```

Dirbaio:
LGTM. We should also change the adapters to convert `Ok(0)` to an error.

adamgreig:
Thanks catching those points! I've updated the first commit and added a new one that updates the adaptors.

Dirbaio (approved):
LGTM!

PR #507, opened 2023-09-18
Document `write_fmt` length trick for slice

MabezDev:
I put this on the documentation for the `Write::write_fmt` itself, instead of the slice impl, for more exposure.

### DIFF
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -401,6 +401,17 @@ pub trait Write: ErrorType {
     /// If you are using [`WriteReady`] to avoid blocking, you should not use this function.
     /// `WriteReady::write_ready()` returning true only guarantees the first call to `write()` will
     /// not block, so this function may still block in subsequent calls.
+    ///
+    /// Unlike [`Write::write`], the number of bytes written is not returned. However, in the case of
+    /// writing to an `&mut [u8]` its possible to calculate the number of bytes written by subtracting
+    /// the length of the slice after the write, from the initial length of the slice.
+    ///
+    /// ```rust
+    /// # use embedded_io::Write;
+    /// let mut buf: &mut [u8] = &mut [0u8; 256];
+    /// let start = buf.len();
+    /// let len = write!(buf, "{}", "Test").and_then(|_| Ok(start - buf.len()));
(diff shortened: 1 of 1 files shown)


embedded-io/src/lib.rs
```diff
@@ -401,6 +401,15 @@ pub trait Write: ErrorType {
     /// If you are using [`WriteReady`] to avoid blocking, you should not use this function.
     /// `WriteReady::write_ready()` returning true only guarantees the first call to `write()` will
     /// not block, so this function may still block in subsequent calls.
+    ///
+    /// Unlike [`Write::write`], the number of bytes written is not returned. However, in the case of
+    /// writing to an `&mut [u8]` its possible to calculate the number of bytes written by subtracting
+    /// the length of the slice after the write, from the initial length of the slice.
+    ///
+    /// ```rust,ignore
+    /// let start = buf.len();
```
Dirbaio, line 410: for this to work you need to explicitly ensure `buf` is a slice: `let buf: &mut [u8] = &mut [0u8; 256];` can you add that to the snippet? 

not adding it creates a bit of a footgun: `let mut buf = [0u8; 256]` will compile but not work because the `write!` macro creates a temporary slice, so `buf.len()` afterwards will refer to the array len (so, 256), not to the shortened slice len.

embedded-io/src/lib.rs
```diff
@@ -401,6 +401,15 @@ pub trait Write: ErrorType {
     /// If you are using [`WriteReady`] to avoid blocking, you should not use this function.
     /// `WriteReady::write_ready()` returning true only guarantees the first call to `write()` will
     /// not block, so this function may still block in subsequent calls.
+    ///
+    /// Unlike [`Write::write`], the number of bytes written is not returned. However, in the case of
+    /// writing to an `&mut [u8]` its possible to calculate the number of bytes written by subtracting
+    /// the length of the slice after the write, from the initial length of the slice.
+    ///
+    /// ```rust,ignore
```
Dirbaio, line 409: can you remove the ignore so that the example is tested, so it won't bitrot?

Dirbaio (approved):
thanks!

PR #511, opened 2023-10-02
Release embedded-io{,-async,-adapters} v0.6.0

### LINKED ISSUE #510: Release embedded-io(-async) 0.6
We need a backwards-incompatible release because of [this change](https://github.com/rust-embedded/embedded-hal/pull/505) (which I absolutely like though!).

Context: I'm updating a bunch of crates from e-io 0.3 and 0.4 to 0.5 and their release is imminent. But given the upcoming breaking change in eio - if I can rebase my work on top of a freshly-released 0.6 that would be ideal. Otherwise, I need to do - relatively soon - another backwards incompatible release of these crates.

Is there any additional scope we need so as to branch a new release or shall we just do it now?

Dirbaio:
closes #510

### DIFF
--- a/embedded-hal-bus/src/lib.rs
+++ b/embedded-hal-bus/src/lib.rs
@@ -2,6 +2,11 @@
 #![warn(missing_docs)]
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+// disable warning for already-stabilized features.
+// Needed to pass CI, because we deny warnings.
+// We don't immediately remove them to not immediately break older nightlies.
+// When all features are stable, we'll remove them.
+#![cfg_attr(feature = "async", allow(stable_features))]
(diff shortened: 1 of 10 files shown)


embedded-hal-bus/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(feature = "async", feature(async_fn_in_trait, impl_trait_projections))]
+#![allow(stable_features)]
```
MabezDev, line 6: It might be worth documenting somewhere why we need this and when we can remove it.

MabezDev (dismissed):
LGTM, just one small nitpick about documenting `#![allow(stable_features)]`, feel free to ignore it.

embedded-hal-bus/src/lib.rs
Dirbaio: done

MabezDev (approved):

PR #519, opened 2023-11-06
Handle reading from Tokio with empty buf as a special case

rmja:


Dirbaio:
nice find, thanks! Could you add it to the changelog?

rmja:
@Dirbaio absolutely, done!

PR #522, opened 2023-11-16
async require rust 1.75 (beta)

burrbull:


burrbull:
cc @eldruin @Dirbaio should we remove doc_cfg?
Now it's only present in `embedded-hal-bus` and `embedded-hal-io`.

burrbull:
Closing. See #534

PR #534, opened 2023-11-28
Use `feature()` on nightly toolchains only.

Dirbaio:
`feature()` is only allowed on Nightly, it's completely disallowed on stable and beta
even for already-stabilized features. So, we autodetect whether the user is using nightly
and conditionally use `feature()`. This allows the crates to Just Work on current 1.75 beta
and will also Just Work when 1.75 stable is out.

Keeping `feature()` is desirable to keep support for:

- Espressif's xtensa rustc fork. (they build from the stable branch but enable use of `feature()`, so latest xtensa rustc still requires `feature()`)
- Users of older nightlies

Once xtensa rust 1.75 is out, we can remove this (upstream nightlies that require `feature()` will be quite old by then, so dropping support for them should be OK).

I decided to not use already-made crates like `rustversion` to do this because they're quite big and do way more than what we need, so I felt badd adding another dep. The code is inspired from `rustversion`'s build.rs.


MabezDev (approved):
Thank you for thinking of Xtensa too :blush:, LGTM!

PR #537, opened 2023-11-28
Rename `DelayUs` to `BasicDelay`, add `delay_ns`.

### LINKED ISSUE #521: `embedded_hal::spi::Operation::DelayUs` is not granular enough
Hi,

When porting drivers to use the new HAL traits we've found an issue that was "implementation specific" before, but now is part of the traits, and that is how to insert delays due to chip requirements.

Usually this means for example (these come from the DW1000 chip):

- CS must be low for 50 ns before a transaction starts 
- CS must stay low for 200 ns after a transaction finishes
- CS must stay high for 250 ns between two transactions

With the current API there is no way to represent these kinds of delays in the `Operation` API as it is microsecond based.
As it is now, either we need to pay up to 20x more time than required by selecting a 1 us delay, or more as it is in practice due to implementation details of timers, one will get between 1 and 2 us of actual delay.
So with the requirements stacked, one get 3-6 us of delay (!) instead of 500 ns in total. 

I propose an `Operation::DelayNs` is added to cover this use-case.

For the requirement between two transactions, I think this might be out of scope of this API?
But if there are any ideas here I'd love to hear them. 
I would argue that if it cannot be added to this trait then it's not possible to achieve generic drivers over chips with CS requirements. This because you must always sidestep this trait then to actually implement support for a chip in a robust manner.
Generally though chips "work" even when these times are not upheld, but I do not think we should build the SPI traits on hope that it will work when requirements cannot be specified.

If I was to shoot from my hip I'd say that the trait could have a constant that is

```rust
const TIME_BETWEEN_TRANSACTIONS_NS: u32 = 0; // 0 for the current behavior
```

and that would use the same nanosecond API required above.

Looking forward to your feedback!

BR Emil

### LINKED ISSUE #535: async DelayUs should have a default DelayMs impl.
https://github.com/rust-embedded/embedded-hal/blob/master/embedded-hal-async/src/delay.rs#L11

Dirbaio:
As discussed in today's meeting, we should add nanosecond precision to the delay trait.

Fixes #521
Fixes #535

Though as discussed in the meeting as well, we should add configuration settings to the SpiDevice impls in `embedded-hal-bus` to do the delays, and document drivers are discouraged from using `Operation::DelayNs` for CS-to-clocks delays. Opened #537 to track this

### DIFF
--- a/embedded-hal-async/CHANGELOG.md
+++ b/embedded-hal-async/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - delay: Rename `DelayUs` to `DelayNs`
 - delay: Add `DelayNs::delay_ns()`
 - delay: Add default impls of `delay_ms` and `delay_us` based on `delay_ns`.
+- spi: Rename `Operation::DelayUs` to `Operation::DelayNs`, with nanosecond precision.
 
 ## [v1.0.0-rc.1] - 2023-08-15
 
--- a/embedded-hal-async/CHANGELOG.md
+++ b/embedded-hal-async/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Minor document fixes.
 - Add #[inline] hints to most of `embedded-hal-async` functions.
+- delay: Rename `DelayUs` to `DelayNs`
+- delay: Add `DelayNs::delay_ns()`
+- delay: Add default impls of `delay_ms` and `delay_us` based on `delay_ns`.
 
 ## [v1.0.0-rc.1] - 2023-08-15
 
--- a/embedded-hal/src/delay.rs
+++ b/embedded-hal/src/delay.rs
@@ -1,25 +1,42 @@
 //! Delays.
 
-/// Microsecond delay.
-pub trait DelayUs {
+/// Delay with up to nanosecond precision.
+pub trait DelayNs {
+    /// Pauses execution for at minimum `ns` nanoseconds. Pause can be longer
+    /// if the implementation requires it due to precision/timing issues.
+    fn delay_ns(&mut self, ns: u32);
+
     /// Pauses execution for at minimum `us` microseconds. Pause can be longer
     /// if the implementation requires it due to precision/timing issues.
-    fn delay_us(&mut self, us: u32);
+    fn delay_us(&mut self, mut us: u32) {
+        while us > 4_294_967 {
+            us -= 4_294_967;
+            self.delay_ns(4_294_967_000);
+        }
+        self.delay_ns(us * 1_000);
+    }
 
     /// Pauses execution for at minimum `ms` milliseconds. Pause can be longer
     /// if the implementation requires it due to precision/timing issues.
     #[inline]
-    fn delay_ms(&mut self, ms: u32) {
-        for _ in 0..ms {
-            self.delay_us(1000);
+    fn delay_ms(&mut self, mut ms: u32) {
+        while ms > 4294 {
+            ms -= 4294;
+            self.delay_ns(4_294_000_000);
         }
+        self.delay_ns(ms * 1_000_000);
     }
 }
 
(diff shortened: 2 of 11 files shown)


MabezDev (dismissed):
Looks good to me, thanks!

adamgreig:
What makes it "Basic"? I still prefer DelayNs in line with having DelayUs before, assuming we save Delay for some future trait, but maybe it's too confusing since it provides all three methods?

I think of it as "delay with ns resolution", plus methods to delay for longer units which can trivially be done with more nanoseconds.

MabezDev:
> What makes it "Basic"? I still prefer DelayNs in line with having DelayUs before, assuming we save Delay for some future trait, but maybe it's too confusing since it provides all three methods?

I also thought about making this point, but I guess as an end user with no knowledge of these traits I wouldn't assume `DelayNs` would be able to do other resolution delays. It's a bit subjective, to be honest. Maybe we can trial it in an rc and see what implementors think?

MabezDev (approved):

Dirbaio:
it's "basic" as "not using the fancy hypothetical future `Duration` type that will magically solve all our problems"... :see_no_evil: 

I think `BasicDelay` makes more sense than `DelayNs`, because `DelayNs` implies we also have `DelayUs`, `DelayMs`, which we don't. Also it might throw off users looking for microsecond or millisecond delays? 

No strong opinion from me, I'm fine with `DelayNs` too. If people think `DelayNs` is better I'll change it.

MabezDev:
Just to throw some other names into the mix:

- `DelayU32` - Rationale: All trait methods use a u32 instead of `core::time::Duration` which I guess more accurately describes why it's not just called `Delay`.
- `ImpreciseDelay` or `LowResolutionDelay` - Depending on the implementor, there can be different behaviours, particularly for `delay_ns`.

I have no strong opinions on any of the suggestions so far, I'd be happy with any of them.

embedded-hal-async/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Minor document fixes.
 - Add #[inline] hints to most of `embedded-hal-async` functions.
+- delay: Rename `DelayUs` to `BasicDelay`
+- delay: Add `BasicDelay::delay_ns()`
+- delay: Add default impls of `delay_ms` and `delay_us` based on `delay_ns`.
+- spi: Rename `Operation::DelayUs` to `Operation::DelayNs`, with microsecond precision.
```
adamgreig, line 15:
```suggestion
- spi: Rename `Operation::DelayUs` to `Operation::DelayNs`, with nanosecond precision.
```

Dirbaio:
> `DelayU32`

same as `DelayNs`, it implies we have other traits in the set like `DelayU16` but we don't.

> ImpreciseDelay or LowResolutionDelay

hmm I'm not sure, there's nothing inherently "imprecise" about it, it just rounds up if needed. The `Delay` trait with the magic rainbow unicorn `Duration` type will also likely be spec'd to round up if the hardware can't reach the requested precision, it won't be more precise.

embedded-hal-async/CHANGELOG.md
Dirbaio: oops fixed.

embedded-hal/src/delay.rs
```diff
@@ -1,25 +1,42 @@
 //! Delays.
 
-/// Microsecond delay.
-pub trait DelayUs {
+/// Basic delay.
```
MabezDev, line 3: Based on the confusion around this trait, do we want to add a bit more to the docs. Might be worthwhile discussing the holy-grail Duration trait and why this trait exists instead?

Dirbaio: I wouldn't, it's not actionable info for the end user, and there's the risk it might never materialize even :sweat_smile: .

I forgot the "basic" on the rename tho, I've replaced "basic delay" with "delay with up to nanosecond precision".

MabezDev (approved):

PR #542, opened 2023-11-30
Use named constants to clarify delay-stepdown behavior

jamesmunns:
Add some named constants to clarify behavior/intent, which confused at least one person already.

The compiler almost certainly should be able to optimize away `N / M * M`, but if we'd like to be extra sure, I can add another const.


Dirbaio (approved):
thank you!

PR #544, opened 2023-12-08
Add an explanation of how 7-bit addresses should be aligned within a u8

jamesmunns:


jamesmunns:
As a note, this seems to match existing drivers and HALs that I spot-checked. We may want to also:

* add a similar note for 10-bit addresses (I have never used them)
* potentially go further and say that `address` values exceeding 0x7F should be treated as an error

eldruin (approved):
Looks good to me, thank you!
Please feel free to send another PR with those refinements if you want.

PR #545, opened 2023-12-13
ci: simplify, test MSRV.

### LINKED ISSUE #538: ci: test MSRVs separately.
in #490 we decided to have different MSRVs per crate

- Adjust CI to test each crate down to their actual MSRV
- Update README.

Dirbaio:
- Do two MSRV tests, for 1.60 and 1.75
- 1.75 uses beta for now, will change after it's released.
- Use separate jobs instead of `matrix`. The Github Actions matrix is very cursed, especially if you use `include`. We don't have that many combinations, IMO doing separate jobs is way more straightforward and only slightly more verbose.
- `embedded-hal` needs `rust-version = 1.60` because it uses `dep:`.

Fixes #538


Dirbaio:
needs updating repo settings, will do after PR approval.

eldruin (approved):
Great, thank you!

eldruin:
I have already updated the branch protection rules as these needed to be updated before the merge. Otherwise some required builds were missing.

eldruin:
Now we are missing a build _without_ features, though. That might be a good addition.

Dirbaio:
> Now we are missing a build without features, though. That might be a good addition.

the `test` job is doing a test without features, which should already build all crates without features. (or is there some scenario where test succeeds but build can fail?)

eldruin:
Ah, I missed that. Thank you!

PR #550, opened 2023-12-19
gpio: remove `ToggleableOutputPin`, move `toggle()` to `StatefulOutputPin`.

### LINKED ISSUE #549: `ToggleableOutputPin` feels redundant
What are the practical use cases for a pin that you can _only_ toggle, and is there hardware out there with pins that you can _only_ toggle but not read the current state of?

I can't think of a use case for the trait beyond blinking an LED or something and even then it shouldn't need to be a separate trait. I think it should rather be a method with a default impl on `StatefulOutputPin` to avoid having to implement it separately in every HAL:

```
pub trait StatefulOutputPin: OutputPin {
    // ...

    fn toggle(&mut self) -> Result<(), Self::Error> {
        let new_state = self.is_set_low()?.into();
        self.set_state(new_state)
    }
}
```

The default implementation could of course be overridden on hardware where there is a more efficient way to toggle a pin.

Dirbaio:
fixes #549

### DIFF
--- a/embedded-hal/CHANGELOG.md
+++ b/embedded-hal/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-No unreleased changes yet
+- gpio: remove `ToggleableOutputPin`, move `toggle()` to `StatefulOutputPin`.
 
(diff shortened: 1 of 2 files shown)


eldruin (approved):
I have read the discussion and agree with this change.
Thank you!

PR #555, opened 2023-12-26
Release v1.0.0

Dirbaio:
Cleaned up readmes, prepared v1.0 release.

To be merged the 28th.

- Move migration guide to docs dir.
- docs/migration: add advice to impl both versions, add e-h-compat
- docs: update readmes for v1.0
- Release embedded-hal{,-async,-nb} v1.0.0, embedded-hal-bus v0.1.0


embedded-hal-nb/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 No unreleased changes
 
+## [v1.0.0] - 2023-12-28
+
+- Updated `embedded-hal` to version `1.0.9`.
```
RicoGit, line 14: why `1.0.9` not `1.0.0`

Dirbaio: typo, thanks!

README.md
```diff
@@ -4,6 +4,8 @@
 
 This project is developed and maintained by the [HAL team](https://github.com/rust-embedded/wg#the-hal-team).
 
+📣 `embedded-hal` v1.0 is now released! Check out the [announcement blog post](https://blog.rust-embedded.org/embedded-hal-v1/), the [API documentation](https://docs.rs/embedded-hal) and the [migration guide](docs/migrating-from-0.2-to-1.0.md).
```
rursprung, line 7: maybe you want to make this more useful using the new note/alert syntax from GH? (see [their blog post for more information](https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/))

docs/migrating-from-0.2-to-1.0.md
```diff
@@ -372,3 +374,59 @@ The `embedded-hal` project now spans several crates, where some functionality ha
...
+
+It is strongly recommended that HAL implementation crates provide implementations for both the `embedded-hal` v0.2 and v1.0 traits.
+This allows users to use drivers using either version seamlessly.
+
+The way you do it is adding a dependency on both versions in `Cargo.toml` like this:
+
+```toml
+[dependencies]
+embedded-hal-02 = { package = "embedded-hal", version = "0.2.7", features = ["unproven"] }
+embedded-hal-1 = { package = "embedded-hal", version = "1.0" }
```
rursprung, line 387: why do you want to rename `embedded-hal` to `embedded-hal-v1`? presuming that support for v0 will be dropped in the future v1 should be the only one remaining and then there's no need to refer to it as anything other than `embedded-hal`

docs/migrating-from-0.2-to-1.0.md
```diff
@@ -372,3 +374,59 @@ The `embedded-hal` project now spans several crates, where some functionality ha
...
+
+impl embedded_hal_1::digital::InputPin for Input {
+    fn is_high(&mut self) -> Result<bool, Self::Error> {
+        ...
+    }
+
+    fn is_low(&mut self) -> Result<bool, Self::Error> {
+        ...
+    }
+}
```
RicoGit, line 423: May be better propose v1.0 as main implementation and v.0.2 as dependent.
```Rust
      /// Implement the v1.0 traits on the struct.
    impl embedded_hal_1::digital::ErrorType for Input {
        type Error = Infallible;
    }

    impl embedded_hal_1::digital::InputPin for Input {
        fn is_high(&mut self) -> Result<bool, Self::Error> {
            ...
        }

        fn is_low(&mut self) -> Result<bool, Self::Error> {
            ...
        }
    }

    /// Implement the v0.2 traits on the struct.
    impl<P: embedded_hal_1::digital::InputPin> embedded_hal_02::digital::v2::InputPin for P {
        type Error = <Self as embedded_hal_1::digital::ErrorType>::Error;

        fn is_high(&self) -> Result<bool, Self::Error> {
            self.is_high()
        }

        fn is_low(&self) -> Result<bool, Self::Error> {
            self.is_low()
        }
    }
```
v0.2 also might be under the some feature

Dirbaio: > May be better propose v1.0 as main implementation and v.0.2 as dependent.

i'd rather leave that up to the HAL, there's a few ways of doing it. For example, some HALs have inherent methods so the user can call them without importing any traits (and so the API can be different, for example for GPIO it's nice to make it infallible). In that case the trait methods would call into the inherent methods.

(In this example I just wanted to show how to add the two versions as dependencies, as it's relatively unknown that it's possible. I've seen people entirely remove 0.2 just because they assumed it's not)

> v0.2 also might be under the some feature

Same, it's better left up to each HAL.

docs/migrating-from-0.2-to-1.0.md
```diff
@@ -41,7 +43,7 @@ For `embedded-hal` 1.0, we decided to drop the first goal, targeting only the se
 - The second goal delivers much more value. Being able to use any driver together with any HAL crate, out of the box, and across the entire Rust Embedded ecosystem, is just plain awesome.
 
 This refocusing on drivers is the root cause of many of the changes between `embedded-hal` 0.2 and 1.0:
-- [Associated type compatibiilty](#removed-traits)
+- [Associated type compatibilitycompatibiilty](#removed-traits)
```
rursprung, line 46:
```suggestion
- [Associated type compatibility](#removed-traits)
```

docs/migrating-from-0.2-to-1.0.md
```diff
@@ -363,12 +365,68 @@ The `embedded-hal` project now spans several crates, where some functionality ha
 
 | Crate | crates.io | Docs | |
```
rursprung, line 366: maybe it makes sense to also add a comment explicitly stating that these crates have individual release cycles and not all of them are hitting v1.0 at the same time?

README.md
Dirbaio: huh, nice. Styling looks nice, but it seems you can't remove the header saying "Note" or similar, so maybe it'd look a bit weird?

docs/migrating-from-0.2-to-1.0.md
Dirbaio: I did it like this in Embassy because I thought it's more explicit. When the HAL drops 0.2 support, renaming is trivial with a global find-and-replace.

README.md
rursprung: i think labeling this as "important" or also "note" would be fine, including with the header?

docs/migrating-from-0.2-to-1.0.md
Dirbaio: done

embedded-hal/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+No unreleased changes yet.
+
+## [v1.0.0] - 2023-12-28
+
 - gpio: remove `ToggleableOutputPin`, move `toggle()` to `StatefulOutputPin`.
 
 ## [v1.0.0-rc.3] - 2023-12-14
```
eldruin, line 16: Could you incorporate the changes from 0.2.4 - 0.2.7 in this changelog as well?

eldruin (commented):
Great, thank you!

crespum:
Minor typo, [this line](https://github.com/rust-embedded/embedded-hal/pull/555/files#diff-98242bdbe25001f862872c9e26459860b67a95ab0bb7d5b675c29a7e7e4afd72R96) is missing a period.

MabezDev (approved):
itshappening

PR #557, opened 2023-12-27
i2c: disallow returning before transfer is finished.

### LINKED ISSUE #455: Clarify flushing/"blocking" behavior

Dirbaio:
Fixes #455.

### DIFF
--- a/embedded-hal/src/i2c.rs
+++ b/embedded-hal/src/i2c.rs
@@ -38,6 +38,15 @@
 //! you've received from a HAL and "split" it into multiple shared ones, to instantiate
 //! several drivers on the same bus.
 //!
+//! # Flushing
+//!
+//! Implementations must flush the transfer, ensuring the bus has returned to an idle state before returning.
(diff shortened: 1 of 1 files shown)


eldruin (commented):
Looks good to me. Could you fix the formatting?

eldruin (approved):
Thanks!

PR #558, opened 2023-12-27
bus/spi: add semver risk warning to `new_no_delay`.

### LINKED ISSUE #552: Document SemVer hazards of `spi::Operation::DelayNs`
Imagine this timeline:
- A driver author releases cool-driver 1.0.0 which uses `SpiDevice` but doesn't use `DelayNs`.
- A user uses a `SpiDevice` impl from embedded_hal_bus with `NoDelay`. The user tests it and it works fine.
- The driver author notices a timing problem and adds a `DelayNs` and releases it as cool-driver 1.0.1
- The user updates cool-driver (maybe even by accident). It still compiles with no warnings.
- The program panics at runtime.

Aside from making actual changes to the `SpiDevice` trait, I can only think of two possible "solutions":

1. Tell the author of cool-driver that by adding a `DelayNs` they have made a breaking change and they must release it as version 2.0.0 instead.
2. Tell the user that they shouldn't really use `NoDelay`, but if they do they must pin the version of any dependency that they use it with.

I don't really like either of these options, but I don't have any better ideas.

Dirbaio:
Fixes #552.

### DIFF
--- a/embedded-hal-bus/src/spi/critical_section.rs
+++ b/embedded-hal-bus/src/spi/critical_section.rs
@@ -34,6 +34,16 @@ impl<'a, BUS, CS, D> CriticalSectionDevice<'a, BUS, CS, D> {
 impl<'a, BUS, CS> CriticalSectionDevice<'a, BUS, CS, super::NoDelay> {
     /// Create a new [`CriticalSectionDevice`] without support for in-transaction delays.
     ///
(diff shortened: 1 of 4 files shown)


eldruin (approved):
Sounds good to me, thanks!

MabezDev (approved):
Thanks!

PR #568, opened 2024-01-17
Re-export `digital` error types in `embedded-hal-async`.

reitermarkus:
Align with other modules, which also re-export these `embedded_hal` types.


Dirbaio (approved):
thanks!

Issue #377, opened 2022-04-13
Tracking issue for ADC trait


eldruin:
The `ADC` traits (`adc::Channel` and `adc::OneShot`) available on the 0.2.x versions have been removed ahead of the 1.0.0 release. See: #376 
This issue servers as a tracking issue until we add them back/reject them.

The reasons were:
- The traits are quite complicated to use in generic code.
- It seems users would be better served by inherent methods.
- Also, these traits are `nb`-only.

Timer traits as of the 0.2.7 release:
- [Docs](https://docs.rs/embedded-hal/0.2.7/embedded_hal/trait.Adc.html)
- [Source](https://github.com/rust-embedded/embedded-hal/blob/2e91b430c5f3c2e06cb3f12a7812e17cafd2d3bc/src/adc.rs#L89)

Relevant issues/PRs:
- #10 
- #123 
- #101 
- #110
- Your issue here!

Please feel free to participate, highlight your current use cases, problems and provide solutions.

ryankurte:
the updated `spi` traits have me wondering if a similar approach would suit for ADCs... something like a shared `AdcDevice` that is split into separate channels implementing `AdcChannel` which drivers then take in the same manner as digital pins or any other injected type.

ADC peripherals would seem to have very similar constraints as sharing as a bus (one ADC peripheral with multiple channels that may be used in different places, and a need to be able to take ownership of the peripheral for more complex behaviors like interleaving or DMA)

Dirbaio:
Maybe having a trait for an "ADC device" is not really needed? We could just have this

```
trait AdcChannel
    fn read(&mut self) -> Result<u32, Error>;
}
```

If the hardware has a single multi-channel ADC it's up to the implementors to do the "sharing" for it. Via RefCell, Mutex, or whatever. Or it could be even a noop for MCU peripheral ADCs: hit the registers directly from `read()` and ensure the channels are `!Send` so that the user can't hit the ADC concurrently from multiple threads (irq priority levels).

> need to be able to take ownership of the peripheral for more complex behaviors like interleaving or DMA.

I don't think making traits for these is feasible, they're incredibly hardware dependent...

ryankurte:
> Maybe having a trait for an "ADC device" is not really needed? We could just have this

what i'm mostly wondering is whether this is a consistent problem with spi / i2c / adc / etc. that we can extract in a consistent way, but yeah seems to me like `AdcChannel` could be a good start.

[...]

PR #569, opened 2024-01-17
WIP: Add `AdcChannel` trait.

reitermarkus:
This adds an `AdcChannel` trait as mentioned in https://github.com/rust-embedded/embedded-hal/issues/377.

Still needs to be tested by implementing an actual driver. I feel there is something still missing with regard to how the return value can actually be used further. Is it even practical to have this trait without having a corresponding unit and a lower/upper bound for converting the raw value?


onkoe:
That's a good point! Also, it may be challenging to take something implementing `AdcChannel` in a library if there's no way to tell what the voltage 'could' be. 

However, how do you ask every implementation to know the voltage range of an arbitrary device? Would having an `Option<Voltage>` suffice? (i.e., we 'ask' the implementation to tell us if it knows)

My apologies if I've missed your point, though! 😄

reitermarkus:
I have changed the API here now to measure voltage instead of the raw value.

> the voltage range of an arbitrary device

I think you may have indeed missed my point. 😄 I am talking about the voltage range of the ADC itself. So it's not an arbitrary device's voltage, but the voltage the ADC is actually measuring. For example, with a 16-bit ADC with a range of ±5V, this means `i16::MIN` corresponds to -5V and `i16::MAX` to +5V.

Then you can swap out ADCs, as long as they support the range you need. And the rest of the code only depends on the measured voltage. As a simple example, I have a water depth sensor with a range of 0-5V corresponding to 0-5m, which would be a 1:1 mapping from mV to mm. So as long as I have any ADC that can measure between 0 and 5V, I can implement e.g. a `WaterDepthSensor` using the trait.

onkoe:
Got it - I thought I was missing something! Thanks for the clarification ☺️

reitermarkus:
Given there can be ADCs that measure voltage or current, I have now split the `AdcChannel` trait into `Voltmeter` and `Ammeter` traits.

kurtjd:
I would be interested in seeing this gain more traction. My particular use case is for developing generic drivers for analog temperature sensors. Would be nice for the drivers to not have to depend on any particular ADC implementation.

michalfita:
> How would a trait look like without voltage/current? You cannot implement a driver against a raw value without unit. If the voltage is board-specific or calibration-specific, you will need to write a specific Voltmeter implementation for getting the voltage in any case.

How about sound samples then? PCM writes a word per sample, but it's not voltage. Consumer grade equipment uses nominal level of 0.316V (-10 dBV) while professional uses 1.228V (+4 dBu). Peak amplitude (saturated sample) are 0.894V and 3.472V respectively.

If the driver converts directly to voltage first, then application back to PCM sample, that's inefficient.

But I may be wrong.

PR #574, opened 2024-01-23
e-h-bus/spi: Require infallible chipselect pins

### LINKED ISSUE #573: The SPI sharing utilities are broken for fallible chipselect pins
This affects all the spi sharing utilities in some way, but I would like to highlight the worst case scenario:

* You are using the mutex or critical section impl
* In thread 1 you do a transaction to device A, but the cs pin fails to deassert
* Before thread 1 even has a chance to react to the error, it is preempted by thread 2
* Thread 2 does a transaction to device B while the cs for device A is still active
* This leads to data corruption at best, and physical damage at worst

---

## Option 1: Infallible only

`CS: OutputPin<Error = Infallible>`

This is the simplest option. In fact it is even simpler than the existing impls because it allows us to remove the `DeviceError` enum.

You can still use fallible chipselect pins if you wrap them in some sort of adapter that panics on failure. We could even provide such an adapter.

## Option 2: track "poisoned" chipselects

I have some idea how to implement this, it would involve adding a shared flag to each bus, as well as another flag for each `SpiDevice`. The impl would refuse to do any spi transactions until the offending chipselect is fixed.

This would be a lot of extra overhead and complexity that is completely unnecessary for 99.9% of users. I don't think the compiler will be able to remove the unnecessary overhead.

## Option 3: do nothing

Document the broken error handling, how to work around it, and where it is impossible to work around.

If we want to be thorough, the amount of warnings and disclaimers will probably outweigh the amount of actual code: 

<details>
<summary>expand me!</summary>

- how to detect a failed chipselect
	- `DeviceError::Cs`, obviously, but also
	- `DeviceError::Spi`, because it can "hide" chipselect errors 
	- basically any `SpiDevice` error
	- downstream errors such as `DisplayError::BusWriteError` from `display_interface` (but ironically not `DisplayError::CSError`)
- how to properly recover from a chipselect error
	- **before** using any device on the bus:
		- get access to the `SpiDevice` by destroying whichever driver was using it
		- get access to the chipselect by destroying the `SpiDevice`
		- do whatever you need to do to fix and deassert the chipselect pin
	- in multitheaded/concurrent code, it is impossible to ensure that this is done before another thread uses a device on the bus. It is not possible to correctly handle chipselect failure in this case
- what happens if you use a device on the bus before fixing the chipselect error
	- this is even a minor problem for `ExclusiveDevice` because the framing will be messed up
	- if two chipselects are active at once
		- the other device will receive garbage data
		- both devices will fight to drive MISO, causing data corruption and possibly even physical damage

</details>

---

## Conclusions

If you only have infallible chipselect pins, or if you want to panic on chipselect failure, option 1 is the best.

If you actually want to try to gracefully recover from chipselect errors, option 2 is the only real option.

Option 3 isn't good for either group. It is slightly inconvenient for the first group, and it is almost impossible to use correctly for the second group

My recommendation is option 1 because it is the best option for most people. If someone really needs option 2, they can write it for themself, it doesn't need to be in our crate.

GrantM11235:
Fixes #573

- [x] Require infallible chipselect pins
- [x] Add an unwrapping adapter
	- [ ] Think of a good name for the adapter
- [ ] Add documentation
- [ ] Update changelog

### DIFF
--- a/embedded-hal-bus/src/spi/exclusive.rs
+++ b/embedded-hal-bus/src/spi/exclusive.rs
@@ -1,5 +1,7 @@
 //! SPI bus sharing mechanisms.
 
+use core::convert::Infallible;
+
 use embedded_hal::delay::DelayNs;
 use embedded_hal::digital::OutputPin;
 use embedded_hal::spi::{ErrorType, Operation, SpiBus, SpiDevice};
@@ -10,7 +12,6 @@ use embedded_hal_async::{
 };
 
 use super::shared::transaction;
-use super::DeviceError;
 
 /// [`SpiDevice`] implementation with exclusive access to the bus (not shared).
 ///
@@ -72,15 +73,15 @@ impl<BUS, CS> ExclusiveDevice<BUS, CS, super::NoDelay> {
 impl<BUS, CS, D> ErrorType for ExclusiveDevice<BUS, CS, D>
 where
     BUS: ErrorType,
-    CS: OutputPin,
+    CS: OutputPin<Error = Infallible>,
 {
-    type Error = DeviceError<BUS::Error, CS::Error>;
+    type Error = BUS::Error;
 }
 
 impl<Word: Copy + 'static, BUS, CS, D> SpiDevice<Word> for ExclusiveDevice<BUS, CS, D>
 where
     BUS: SpiBus<Word>,
-    CS: OutputPin,
+    CS: OutputPin<Error = Infallible>,
     D: DelayNs,
 {
     #[inline]
@@ -94,7 +95,7 @@ where
 impl<Word: Copy + 'static, BUS, CS, D> AsyncSpiDevice<Word> for ExclusiveDevice<BUS, CS, D>
 where
     BUS: AsyncSpiBus<Word>,
-    CS: OutputPin,
+    CS: OutputPin<Error = Infallible>,
     D: AsyncDelayNs,
 {
     #[inline]
@@ -102,7 +103,9 @@ where
         &mut self,
         operations: &mut [Operation<'_, Word>],
     ) -> Result<(), Self::Error> {
-        self.cs.set_low().map_err(DeviceError::Cs)?;
+        use crate::spi::shared::into_ok;
+
+        into_ok(self.cs.set_low());
 
         let op_res = 'ops: {
             for op in operations {
@@ -128,12 +131,8 @@ where
 
         // On failure, it's important to still flush and deassert CS.
         let flush_res = self.bus.flush().await;
-        let cs_res = self.cs.set_high();
+        into_ok(self.cs.set_high());
 
-        op_res.map_err(DeviceError::Spi)?;
-        flush_res.map_err(DeviceError::Spi)?;
-        cs_res.map_err(DeviceError::Cs)?;
-
-        Ok(())
+        op_res.and(flush_res)
     }
 }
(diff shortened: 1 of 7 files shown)


embedded-hal-bus/src/spi/exclusive.rs
```diff
@@ -72,15 +73,15 @@ impl<BUS, CS> ExclusiveDevice<BUS, CS, super::NoDelay> {
...
 {
-    type Error = DeviceError<BUS::Error, CS::Error>;
+    type Error = BUS::Error;
 }
 
 impl<Word: Copy + 'static, BUS, CS, D> SpiDevice<Word> for ExclusiveDevice<BUS, CS, D>
 where
     BUS: SpiBus<Word>,
-    CS: OutputPin,
+    CS: OutputPin<Error = Infallible>,
```
jannic, line 84: Is this change really needed for ExclusiveDevice? The scenario described in #573 doesn't apply here, as the bus isn't shared in this case.

GrantM11235: The issue is less severe with ExclusiveDevice, but there is still a problem. The `SpiDevice` trait says that all transactions must start by asserting the chipselect (a falling edge), which isn't possible if it is already asserted. This is important for devices that use chipselect for framing. In practice it is possible to work around the issue in this case (always de-assert cs again if you get any kind of DeviceError) but it is a pretty big footgun

GrantM11235:
The docs and changelog are done. I am open to suggestions for the name of `UnwrappingAdapter` if anyone has any better ideas, but I think it is okay as it is.

This PR was brought up at the meeting today, but no one had much to say about it. I assume that means there are no strong objections.

Dirbaio:
i'm personally 50/50 on this. The tradeoffs are subtle and i'm not sure which side of them is the best. So I won't merge this myself but I'm OK with merging if other people in the HAL team are in favor.

MabezDev (approved):
I did mean to leave a review a couple of weeks ago, but forgot, sorry! 

Imo, given that this is not in ehal proper, and if required the users who need poisoning can write their own bus wrapper (and maybe PR it back here?), this LGTM. 

ehal bus is not ehal and we can be a bit more selective on what level of support we want to provide out of the box :D.

Dirbaio (changes requested):
After [discussion in today's meeting](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$S_AH1hCVwFg84CY5UVLkVedIjVXJCYgCaU4tHKsDMYg?via=matrix.org&via=catircservices.org&via=tchncs.de) I now believe we should not require infallible pins, we shold instead keep the current implementation, document how it handles errors and recommend the user use another implementation if it doesn't suit their requirements.

the TLDR is

- This doesn't fix the full problem: there's a similar case where `flush` fails so the SPI keeps clocking out buffered bits, we deassert CS then unlock the bus, another driver asserts their CS and now the buffered bits go to the wrong device. Therefore we should require infallible SPI too, but that's not practical
   - @GrantM11235 proposed an alternative: specify `flush` must not return before SPI is stopped in ALL cases, even on failure, but I don't believe this is generally feasible and I don't think it's worth complicating the SpiBus contract to accomodate for SpiDevice implementation technicalities.
- Therefore we should apply the same "policy" to SPI and CS errors. Not doing so would be inconsistent. However most SPIs out there are fallible, so this would be quite burdensome for the user.
- the 3rd alternative (poisoning) is too complex, it requires wrapping the SPI bus into an "inner" struct for all users even if they're not affected by this problem. It's also not zero cost.

PR #579, opened 2024-03-07
e-h-bus: Adding `AtomicDevice` for I2C and SPI bus sharing in multiple interrupt contexts

ryan-summers:
This PR adds a primitive for sharing an I2C or SPI bus across multiple threads where the synchronization scheme is managed externally (i.e. RTIC sync resources).

The user is returned an `Error::Busy` if they ever attempt to pre-empt usage of the bus.

CC @Dirbaio as we were talking about this in the Matrix channel yesterday.

### DIFF
--- a/embedded-hal-bus/src/i2c/atomic.rs
+++ b/embedded-hal-bus/src/i2c/atomic.rs
@@ -92,7 +92,7 @@ impl<T: Error> Error for AtomicError<T> {
     }
 }
 
-unsafe impl<'a, T> Sync for AtomicDevice<'a, T> {}
+unsafe impl<'a, T> Send for AtomicDevice<'a, T> {}
 
 impl<'a, T> AtomicDevice<'a, T>
 where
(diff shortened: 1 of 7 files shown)


embedded-hal-bus/src/i2c/atomic.rs
```diff
@@ -0,0 +1,170 @@
...
+impl<T: Error> Error for AtomicError<T> {
+    fn kind(&self) -> ErrorKind {
+        match self {
+            AtomicError::Other(e) => e.kind(),
+            _ => ErrorKind::Other,
+        }
+    }
+}
+
+unsafe impl<'a, T> Send for AtomicDevice<'a, T> {}
```
ryan-summers, line 95: I would expect this to be `impl Sync`, but for some reason if I only do Sync, my test application complains that the `AtomicDevice` cannot be shared across threads safely.

ryan-summers:
Thanks for taking this over and pushing it through! :)

 I didn't see the maintainer allow button, my bad.

PR #589, opened 2024-04-16
Add license file to the top-level directory.

### LINKED ISSUE #588: README: Links to LICENSE-APACHE and LICENSE-MIT are not found
The readme have links to files called `LICENSE-APACHE ` and `LICENSE-MIT`

see:
image

However when `LICENSE-APACHE ` or `LICENSE-MIT are clicked an error appears (the long URL versions work)

see:
image

I assume these files were deleted without having the readme updated.

jannic:
These files are referenced from README.md, so they should be present.

The same files are already available in the individual crates, so another possibility would be linking to one of the subdirectories.

I think it's preferable to have them at the top-level for better visibility.

Fixes #588

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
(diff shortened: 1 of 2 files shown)


jannic:
Don't worry (too much) about the CI failures.
They are caused by a new warning (and we set `-D warnings`), but it looks like that warning caused too many issues so it'll probably be reverted. See https://github.com/rust-lang/rust/issues/121708 for more information. The latest comments suggest that the warning has already been disabled in beta.

Dirbaio (approved):

PR #590, opened 2024-04-16
Fix CI errors caused by changes in rust beta

jannic:
Use stable / 1.75.0, instead.


Dirbaio (approved):
thanks for taking care of this!

PR #594, opened 2024-04-23
bus: make constructors set CS high.

Dirbaio:
This avoids a footgun where a HAL makes OutputPins low by default (or the user sets
them to low and doesn't notice).

Discussed in today's WG meeting.


eldruin (approved):
Thank you!
I do not like that a constructor has side-effects, even worse, potentially fallible ones.
However, I agree with the overall assessment in the meeting.

PR #595, opened 2024-04-23
Release embedded-hal-bus v0.2.0

Dirbaio:
Depends on #594.

Needs to be v0.2 because #594 is breaking.

- **bus: make constructors set CS high.**
- **Release embedded-hal-bus v0.2.0**


eldruin:
Could you rebase this now?

Dirbaio:
@eldruin done

PR #607, opened 2024-06-08
Add embedded-hal-bus atomic-device features

### LINKED ISSUE #598: embedded-hal-bus no longer builds with target thumbv6m-none-eabi
Added `embedded-hal-bus = "0.2.0"` to my project and it failed to build.
Using target `thumbv7m-none-eabi` works, using target `thumbv6m-none-eabi` fails. Version 0.1.0 works.

Trying to build the latest `master` branch:
```
./embedded-hal-bus> cargo build --target thumbv6m-none-eabi
   Compiling embedded-hal v1.0.0 (/home/cyber/Documents/VSCode/embedded-hal/embedded-hal)
   Compiling portable-atomic v1.6.0
   Compiling embedded-io-adapters v0.6.1 (/home/cyber/Documents/VSCode/embedded-hal/embedded-io-adapters)
   Compiling embedded-io-async v0.6.1 (/home/cyber/Documents/VSCode/embedded-hal/embedded-io-async)
   Compiling embedded-hal-async v1.0.0 (/home/cyber/Documents/VSCode/embedded-hal/embedded-hal-async)
   Compiling nb v1.1.0
   Compiling embedded-io v0.6.1 (/home/cyber/Documents/VSCode/embedded-hal/embedded-io)
   Compiling embedded-hal-bus v0.2.0 (/home/cyber/Documents/VSCode/embedded-hal/embedded-hal-bus)
   Compiling critical-section v1.1.2
   Compiling embedded-can v0.4.1 (/home/cyber/Documents/VSCode/embedded-hal/embedded-can)
   Compiling embedded-hal-nb v1.0.0 (/home/cyber/Documents/VSCode/embedded-hal/embedded-hal-nb)
error[E0599]: no method named `compare_exchange` found for struct `portable_atomic::AtomicBool` in the current scope
   --> embedded-hal-bus/src/i2c/atomic.rs:122:14
    |
120 | /         self.bus
121 | |             .busy
122 | |             .compare_exchange(
    | |             -^^^^^^^^^^^^^^^^ method not found in `AtomicBool`
    | |_____________|
    | 

error[E0599]: no method named `compare_exchange` found for struct `portable_atomic::AtomicBool` in the current scope
   --> embedded-hal-bus/src/spi/atomic.rs:127:14
    |
125 | /         self.bus
126 | |             .busy
127 | |             .compare_exchange(
    | |             -^^^^^^^^^^^^^^^^ method not found in `AtomicBool`
    | |_____________|
    | 

For more information about this error, try `rustc --explain E0599`.
error: could not compile `embedded-hal-bus` (lib) due to 2 previous errors
```

Seems to be caused by #593 

https://github.com/rust-embedded/embedded-hal/blob/a0ccb6579669804e9be723912ed645a16362c8a6/embedded-hal-bus/src/spi/atomic.rs#L125-L133

https://github.com/rust-embedded/embedded-hal/blob/a0ccb6579669804e9be723912ed645a16362c8a6/embedded-hal-bus/src/i2c/atomic.rs#L120-L128

ARMv6-M does not have atomic CAS, so `portable_atomic` does not implement `AtomicBool::compare_exchange` on this target.

9names:
Add atomic-device and portable-atomic features as described in [this comment](https://github.com/rust-embedded/embedded-hal/issues/598#issuecomment-2086463791) 
The tasks from the comment are copied below for convenience
>
- Add a new atomic-device feature, default off
- Only enable AtomicDevice when either the atomic-device feature is enabled or cfg(target_has_atomic...) is set, so it's by default available on platforms with CAS
- Add new portable-atomic-critical-section and portable-atomic-unsafe-assume-single-core features which proxy to the corresponding portable-atomic features (so users don't have to depend on portable-atomic directly) and enable our atomic-device feature
- Document these new features
- Release as 0.3

Note that the entire util.rs module is only ever used by AtomicDevice and the i2c/spi atomic impls, so we could sensibly rename it to atomic_util.rs and gate that that file instead of everything defined inside it.

Note: untested at this point. Will add comment once testing is done.

Resolves #598


9names:
CI errors, because `cargo clippy --all-features` is definitely not compatible with these new, mutually incompatible features

Dirbaio:
perhaps we can instead do the following:

- Add a feature `portable-atomic`
- Document that if you enable it, you have to add a dep on `portable-atomic` and enable one of the two polyfill mode features on it.
- Enable AtomicDevice if `any(feature = "portable-atomic", target_has_atomic = "8")`

Advantages
- Generalizes better if we add more things requiring atomics in the future (the `portable-atomic` feature would enable all "things" requiring atomics, if we name the feature `atomic-device` we'd need one per "thing")
- Generalizes better if `portable-atomic` adds more polyfill modes in the future.
- Avoids non-additive features in `embedded-hal-bus`.

9names:
Certainly looks cleaner @Dirbaio. I've also added `require-cas` to the enabled features of `portable-atomic` as per the  docs, so that users that fail to add an appropriate polyfill feature on `portable-atomic` will get a link error:
image

embedded-hal-bus/Cargo.toml
```diff
@@ -26,7 +36,7 @@ embedded-hal = { version = "1.0.0", path = "../embedded-hal" }
 embedded-hal-async = { version = "1.0.0", path = "../embedded-hal-async", optional = true }
 critical-section = { version = "1.0" }
 defmt-03 = { package = "defmt", version = "0.3", optional = true }
-portable-atomic = {version = "1", default-features = false}
+portable-atomic = {version = "1.3", default-features = false, features = ["require-cas"]}
```
Dirbaio, line 39: doesn't this need `optional = true`? otherwise thumbv6 build will fail due to `require-cas` being enabled if the user doesn't enable any of the features for `portable-atomic`

9names: You are correct.
I've now made portable-atomic optional and done the cleanup required to handle building without it enabled.

I've tested the following on a thumbv6 target:
- feature disabled - build success, AtomicCell, {spi,i2c}::AtomicDevice not present in docs
- feature enabled, no polyfill set - compile error
- feature enabled, portable-atomic/unsafe-assume-single-core - build success, AtomicCell, {spi,i2c}::AtomicDevice present in docs

Dirbaio (approved):
thank you!

PR #609, opened 2024-06-18
embedded-io-adapters: Add adapters for embedded-storage and embedded-storage-async

aurelj:


aurelj:
Note that this PR currently depends on unreleased git version of `embedded-storage`.

Also note that the failure in Rustdoc check / clippy is due to the CI using an old nightly rust to build the `embedded-storage-async` dependency which requires rust-version = "1.75".

Dirbaio:
`embedded-storage` is not maintained by the WG. Perhaps this makes more sense to add in the `embedded-storage` repo instead?

PR #613, opened 2024-07-05
Add a Rc<RefCell<Bus>>-based implementation of SpiDevice and I2C

adri326:
The current implementation of `RefCellDevice` for sharing an SPI bus using `RefCell` works fine, but restricts users of the library into patterns where the compiler is able to guarantee that the bus reference will outlive all usage of its devices.

In my current work, this isn't really the case, as I would like to hide the `SpiBus`/`OutputPin` details behind some structures a shared trait, so that my code can interface with foreign microcontrollers as it would interface with its local peripherals.
This means that I either need to manually manage the memory used for the `SpiBus`, or I need to use something like `Rc` to safely manage it for me.

Right now, I can create an ad-hoc `ExclusiveDevice` with the reference from my `Rc<RefCell<SpiBus>>` whenever I need to do a transaction, and drop it immediately after, but this is a bit of a hacky solution.

This PR proposes a more ellegant approach, which simply offers an alternative to `RefCellDevice` that uses reference-counting for its SPI bus.

I hesitated on adding a utility method for acquiring a copy of the `Rc<RefCell<SpiBus>>`, but I decided against it, since this is a rather niche need and one could just keep a `Weak<RefCell<SpiBus>>` around and not occur any penalty.

I'm unsure if I did things well regarding adding a new feature, `alloc`, and making sure that it renders correctly on docs.rs

**Edit:** also added an implementation for I2C, for consistency. I noticed that the I2C implementations could benefit from `Clone`, but that's out of the scope of this PR.

### DIFF
--- a/embedded-hal-bus/Cargo.toml
+++ b/embedded-hal-bus/Cargo.toml
@@ -20,8 +20,6 @@ async = ["dep:embedded-hal-async"]
 defmt-03 = ["dep:defmt-03", "embedded-hal/defmt-03", "embedded-hal-async?/defmt-03"]
 # Enables additional utilities requiring a global allocator.
 alloc = []
-# TODO: remove this
-default = ["alloc"]
 
 [dependencies]
 embedded-hal = { version = "1.0.0", path = "../embedded-hal" }
--- a/embedded-hal-bus/Cargo.toml
+++ b/embedded-hal-bus/Cargo.toml
@@ -15,9 +15,11 @@ repository = "https://github.com/rust-embedded/embedded-hal"
 version = "0.2.0"
 
 [features]
-std = []
+std = ["alloc"]
 async = ["dep:embedded-hal-async"]
 defmt-03 = ["dep:defmt-03", "embedded-hal/defmt-03", "embedded-hal-async?/defmt-03"]
+# Enables additional utilities requiring a global allocator.
+alloc = []
 
 [dependencies]
 embedded-hal = { version = "1.0.0", path = "../embedded-hal" }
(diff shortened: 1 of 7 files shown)


Dirbaio:
LGTM! can you add it to I2C too for consistency?

adri326:
I can, yeah; do you want it in a separate PR?

Dirbaio:
either separate or this one works, up to you.

adri326:
Might as well then :)

embedded-hal-bus/Cargo.toml
```diff
@@ -15,9 +15,13 @@ repository = "https://github.com/rust-embedded/embedded-hal"
...
 
 [features]
-std = []
+std = ["alloc"]
 async = ["dep:embedded-hal-async"]
 defmt-03 = ["dep:defmt-03", "embedded-hal/defmt-03", "embedded-hal-async?/defmt-03"]
+# Enables additional utilities requiring a global allocator.
+alloc = []
+# TODO: remove this
+default = ["alloc"]
```
eldruin, line 24:
```suggestion
```

eldruin (commented):
Thank you! Looks good to me.
Could you add some documentation about the alloc feature to the readme and also add an entry to the changelog?

embedded-hal-bus/Cargo.toml
adri326: Oh great I managed to forget to remove it :')

Dirbaio (approved):
thanks!

PR #625, opened 2024-08-24
Clarify Read trait blocking behavior

ivmarkov:
This PR is an effort to clarify the blocking behavior of the `Read` trait with regards to the length of the user-supplied buffer.

It is a follow up on a recent discussion in the ["Rust Embedded" Matrix chat](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$KPz0XkHW0cQt2bAG2-8pGvzZNMSeZlTJlVTBlVVVEvY?via=matrix.org&via=catircservices.org&via=tchncs.de) where the agreement was that we could probably be more explicit about that.

For now, I've changed only the documentation of the blocking `Read` trait. Once/if we arrive at a text which seems reasonable to everybody, I'll extend the PR with a similar change to `embedded_io_async::Read`.

Ideally, we should arrive at a piece of documentation that we could also justify to be included upstream [in the Rust STD Read trait as well](https://github.com/rust-lang/rust/blob/edbc000fa4413f9ff92f4e5492f8ff497716ce48/library/std/src/io/mod.rs#L595), and possibly in its `async` variants in the `futures` and `tokio` crates.

### DIFF
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -309,8 +309,10 @@ pub trait Read: ErrorType {
     /// If bytes are available to read:
     /// - A non-zero amount of bytes is read to the beginning of `buf`, and the amount is returned immediately,
     ///   *without blocking and waiting for more bytes to become available*;
-    /// - It is not guaranteed that *all* available bytes are returned, it is possible for the implementation to
-    ///   read an amount of bytes less than `buf.len()` while there are more bytes immediately available.
+    ///
+    /// Note that once some bytes are available to read, it is *not* guaranteed that all available bytes are returned.
+    /// It is possible for the implementation to read an amount of bytes less than `buf.len()` while there are more
+    /// bytes immediately available.
     ///
     /// This blocking behavior is important for the cases where `Read` represents the "read" leg of a pipe-like
     /// protocol (a socket, a pipe, a serial line etc.). The semantics is that the caller - by passing a non-empty
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -300,16 +300,23 @@ impl<E: fmt::Debug> std::error::Error for WriteFmtError<E> {}
 pub trait Read: ErrorType {
     /// Read some bytes from this source into the specified buffer, returning how many bytes were read.
     ///
-    /// If no bytes are currently available to read, this function blocks until at least one byte is available.
+    /// If no bytes are currently available to read:
+    /// - The method blocks until at least one byte becomes available;
+    /// - Once at least one (or more) bytes become available, a non-zero amount of those is read to the
+    ///   beginning of `buf`, and the amount is returned, *without waiting or blocking any further for
+    ///   more bytes to become available*.
     ///
-    /// If bytes are available, a non-zero amount of bytes is read to the beginning of `buf`, and the amount
-    /// is returned. It is not guaranteed that *all* available bytes are returned, it is possible for the
-    /// implementation to read an amount of bytes less than `buf.len()` while there are more bytes immediately
-    /// available.
+    /// If bytes are available to read:
+    /// - A non-zero amount of bytes is read to the beginning of `buf`, and the amount is returned immediately,
+    ///   *without blocking and waiting for more bytes to become available*;
+    /// - It is not guaranteed that *all* available bytes are returned, it is possible for the implementation to
+    ///   read an amount of bytes less than `buf.len()` while there are more bytes immediately available.
     ///
-    /// If the reader is at end-of-file (EOF), `Ok(0)` is returned. There is no guarantee that a reader at EOF
-    /// will always be so in the future, for example a reader can stop being at EOF if another process appends
-    /// more bytes to the underlying file.
+    /// This blocking behavior is important for the cases where `Read` represents the "read" leg of a pipe-like
+    /// protocol (a socket, a pipe, a serial line etc.). The semantics is that the caller - by passing a non-empty
+    /// buffer - does expect _some_ data (one or more bytes) - but _not necessarily `buf.len()` or more bytes_ -
+    /// to become available, before the peer represented by `Read` would stop sending bytes due to
+    /// application-specific reasons (as in the peer waiting for a response to the data it had sent so far).
     ///
     /// If `buf.len() == 0`, `read` returns without blocking, with either `Ok(0)` or an error.
     /// The `Ok(0)` doesn't indicate EOF, unlike when called with a non-empty buffer.
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -318,6 +318,10 @@ pub trait Read: ErrorType {
     /// to become available, before the peer represented by `Read` would stop sending bytes due to
     /// application-specific reasons (as in the peer waiting for a response to the data it had sent so far).
     ///
+    /// If the reader is at end-of-file (EOF), `Ok(0)` is returned. There is no guarantee that a reader at EOF
+    /// will always be so in the future, for example a reader can stop being at EOF if another process appends
+    /// more bytes to the underlying file.
+    ///
     /// If `buf.len() == 0`, `read` returns without blocking, with either `Ok(0)` or an error.
     /// The `Ok(0)` doesn't indicate EOF, unlike when called with a non-empty buffer.
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error>;
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -302,7 +302,7 @@ pub trait Read: ErrorType {
     ///
     /// If no bytes are currently available to read:
     /// - The method blocks until at least one byte becomes available;
-    /// - Once at least one (or more) bytes become available, a non-zero amount of those is read to the
+    /// - Once at least one (or more) bytes become available, a non-zero amount of those is copied to the
     ///   beginning of `buf`, and the amount is returned, *without waiting or blocking any further for
     ///   more bytes to become available*.
     ///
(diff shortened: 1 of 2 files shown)


ivmarkov:
@Dirbaio @jamesmunns Pinging you here as you were actively participating in the Rust Embedded discussion on the subject.

embedded-io/src/lib.rs
```diff
@@ -300,16 +300,23 @@ impl<E: fmt::Debug> std::error::Error for WriteFmtError<E> {}
 pub trait Read: ErrorType {
     /// Read some bytes from this source into the specified buffer, returning how many bytes were read.
     ///
-    /// If no bytes are currently available to read, this function blocks until at least one byte is available.
+    /// If no bytes are currently available to read:
+    /// - The method blocks until at least one byte becomes available;
+    /// - Once at least one (or more) bytes become available, a non-zero amount of those is read to the
```
jamesmunns, line 305:
```suggestion
    /// - Once at least one (or more) bytes become available, a non-zero amount of those is copied to the
```

jamesmunns (commented):
Overall I'm find with this addition, I think it is very clearly written!

My only questions are whether we say "SHOULD" or "is recommended", to make clear that this isn't mandatory, at least until we have matching language in `std::io::Read`, or whether we note "this is how the `recv` syscall works in posix, and we're aiming to match that behavior which is generally consistent in all major desktop operating systems".

ivmarkov:
> Overall I'm find with this addition, I think it is very clearly written!
> 
> My only questions are whether we say "SHOULD" or "is recommended", to make clear that this isn't mandatory, at least until we have matching language in `std::io::Read`, or whether we note "this is how the `recv` syscall works in posix, and we're aiming to match that behavior which is generally consistent in all major desktop operating systems".

I'm very certain it cannot be "SHOULD" or "recommended". It is a "MUST", so to say, or else any protocol implementation on top of `Read` that assumes this behavior is simply buggy and needs to be rewritten to read only byte-by-byte.

That includes [the "read_until" readers in STD](https://github.com/rust-lang/rust/blob/edbc000fa4413f9ff92f4e5492f8ff497716ce48/library/std/src/io/mod.rs#L2087) and the [`fill_buf`](https://github.com/rust-lang/rust/blob/edbc000fa4413f9ff92f4e5492f8ff497716ce48/library/std/src/io/buffered/bufreader/buffer.rs#L136) implementation.

embedded-io/src/lib.rs
```diff
@@ -300,16 +300,23 @@ impl<E: fmt::Debug> std::error::Error for WriteFmtError<E> {}
...
-    /// is returned. It is not guaranteed that *all* available bytes are returned, it is possible for the
-    /// implementation to read an amount of bytes less than `buf.len()` while there are more bytes immediately
-    /// available.
+    /// If bytes are available to read:
+    /// - A non-zero amount of bytes is read to the beginning of `buf`, and the amount is returned immediately,
+    ///   *without blocking and waiting for more bytes to become available*;
+    /// - It is not guaranteed that *all* available bytes are returned, it is possible for the implementation to
+    ///   read an amount of bytes less than `buf.len()` while there are more bytes immediately available.
     ///
-    /// If the reader is at end-of-file (EOF), `Ok(0)` is returned. There is no guarantee that a reader at EOF
```
jannic, line 310: Did you intentionally remove this paragraph?

While there may be no literal end-of-file for embedded-io, there are similar concepts like a closed pipe, where it doesn't make sense to continue blocking, and there's no data available to return.

ivmarkov: Nope - sorry  - this was completely unintentional. Re-added and thanks for spotting!

MabezDev:
> I'm very certain it cannot be "SHOULD" or "recommended". It is a "MUST", so to say, or else any protocol implementation on top of Read that assumes this behavior is simply buggy and needs to be rewritten to read only byte-by-byte.

I would agree with this, otherwise we end up in the same place where one implementor may choose not to follow this "recommendation" and we end up with differing impls.

This LGTM from my perspective once we've applied the same to the async docs.

embedded-io/src/lib.rs
```diff
@@ -300,12 +300,23 @@ impl<E: fmt::Debug> std::error::Error for WriteFmtError<E> {}
...
+    /// If no bytes are currently available to read:
+    /// - The method blocks until at least one byte becomes available;
+    /// - Once at least one (or more) bytes become available, a non-zero amount of those is copied to the
+    ///   beginning of `buf`, and the amount is returned, *without waiting or blocking any further for
+    ///   more bytes to become available*.
+    ///
+    /// If bytes are available to read:
+    /// - A non-zero amount of bytes is read to the beginning of `buf`, and the amount is returned immediately,
+    ///   *without blocking and waiting for more bytes to become available*;
+    /// - It is not guaranteed that *all* available bytes are returned, it is possible for the implementation to
```
Dirbaio, line 312: minor nit: this line looks like it applies only to the "if there are bytes available to read" case, while it should apply to both. Perhaps split it to a new paragraph and remove the `-`.

Dirbaio:
LGTM. Thanks for taking care of this :+1: 

(I too think it should be MUST, for the already-stated reasons)

embedded-io/src/lib.rs
ivmarkov: Hmmm, but it _is_ applicable only to the "if there are bytes available to read" case, as it talks about - if there are bytes available to read _in the first place_ - how many of those would be returned - i.e. not necessarily all of them (because (a) the buffer might actually be shorter than what is available (b) because the implementation just decides so, for whatever other reasons).

ivmarkov:
I've now added identical changes to the `async` version of the `Read` trait, where the only changes I did is to use "waiting" at the places where the blocking version uses either "blocking" or "blocking and waiting".

With that said, shall we delay the merge a bit until I open a similar PR or issue for the `std::io::Read` trait? *If* they accept a similar change, they might have more feedback to the text, that we might want to incorporate in our `Read` trait versions too?

Dirbaio:
Up to you. Getting feedback or merges to rust-lang PRs is not exactly fast though :)

ivmarkov:
> Up to you. Getting feedback or merges to rust-lang PRs is not exactly fast though :)

I want to mitigate the danger of making the described behavior a MUST in our traits, and get resistance from the STD team to do it for the `Read` trait. 
- If they just return feedback that it is "obvious" - fine... 
- ...but if they say that it is unspecified i.e. "Option B" (what to do with the line-readers and stuff is then another topic), that would be quite a problem...

Dirbaio:
IMO even if std explicitly decides to make it Option B we should go for Option A.

Option B is just insane. as you point out many things like ReadBuf / read_line become broken unless you read byte by byte.

adamgreig:
We briefly discussed this in the meeting today as well, it seems like merging this now seems the best option. We can always wait a bit before releasing it to see what happens on rust-lang.

ivmarkov:
> We briefly discussed this in the meeting today as well, it seems like merging this now seems the best option. We can always wait a bit before releasing it to see what happens on rust-lang.

OK - then be it. I've anyway not even started the PR/issue in STD upstream.

So if you could look at the `embedded_io_async::Read` trait wording as well, and if no additional feedback, I guess we can merge.

ivmarkov:
If no further feedback, can somebody with committer permissions approve this?

embedded-io/src/lib.rs
jannic: This applies to the first case as well, doesn't it? In "Once at least one (or more) bytes become available", one could also ask if it's guaranteed that all available bytes are returned. The answer is "it is not guaranteed", but it's not completely obvious from the description as it is written now.

But as Dirbaio wrote, it is a "minor nit". Personally, I would not expect to have such a guarantee when reading the text you proposed.

jannic (approved):

embedded-io/src/lib.rs
ivmarkov: Not sure how come the first case is unclear that only a subset of the just-read bytes are returned, given that it says (emphasis mine): 
"
- Once at least one (or more) bytes become available, **a non-zero amount of those** is copied to the beginning of `buf`, and the amount is returned...

"

"a non-zero amount of those". Do we really need to be more explicit than that?

ivmarkov: Ah be it. If two folks find it unclear then it is probably unclear. :(
I'll re-arrange so that the problematic text is a separate paragraph.

ivmarkov: @jannic I've re-arranged it now so that the problematic text is a separate paragraph. If you can take a quick look - thanks.

jannic: > Not sure how come the first case is unclear that only a subset of the just-read bytes are returned, given that it says (emphasis mine): "
> 
>     * Once at least one (or more) bytes become available, **a non-zero amount of those** is copied to the beginning of `buf`, and the amount is returned...
> 
> 
> "
> 
> "a non-zero amount of those". Do we really need to be more explicit than that?

IMHO the only possible source of confusion was the asymmetry between the two cases. Yes, in general I agree that the wording "a non-zero amount of those" was sufficiently clear. But then the second case, in addition to that, explicitly stated that it's allowed to return less than available. For the first case, that sentence was missing. Readers may wonder why? Is there a difference in guarantees?

Again, this was a _very_ minor detail. But if we can avoid this possible confusion by just putting the last sentence in a separate paragraph, as you now did, we should do it.

jannic (approved):

embedded-io/src/lib.rs
ivmarkov: Argh, don't want to over-complicate it, but _now_ - with the problematic text being a separate paragraph - the _next_ paragraph that starts with "This waiting behavior is important for the cases where `Read`..." becomes unclear... in that readers might be confused that it only refers to the newly-born "Note..." paragraph, while it refers to _everything_ mentioned above.

Any brilliant ideas? Or is it just my mind that finds the new arrangement also problematic?

MabezDev (approved):
LGTM, thanks for taking care of this!

PR #626, opened 2024-08-26
Update the I2C documentation to use "controller/target" instead of "master/slave"

emshotton:
This brings the language in-line with [Rev. 7.0 of the 2C-bus specification and user manual ](https://www.nxp.com/docs/en/user-guide/UM10204.pdf).

### DIFF
--- a/embedded-hal-async/src/i2c.rs
+++ b/embedded-hal-async/src/i2c.rs
@@ -23,23 +23,23 @@ pub use embedded_hal::i2c::{
 
 /// Async I2c.
 pub trait I2c<A: AddressMode = SevenBitAddress>: ErrorType {
-    /// Reads enough bytes from slave with `address` to fill `buffer`.
+    /// Reads enough bytes from target with `address` to fill `buffer`.
     ///
     /// # I2C Events (contract)
     ///
     /// ``` text
-    /// Master: ST SAD+R        MAK    MAK ...    NMAK SP
-    /// Slave:           SAK B0     B1     ... BN
+    /// Controller: ST TAD+R        CAK    CAK ...    NCAK SP
+    /// Target:              TAK B0     B1     ... BN
     /// ```
     ///
     /// Where
     ///
     /// - `ST` = start condition
-    /// - `SAD+R` = slave address followed by bit 1 to indicate reading
-    /// - `SAK` = slave acknowledge
+    /// - `TAD+R` = target address followed by bit 1 to indicate reading
+    /// - `TAK` = target acknowledge
     /// - `Bi` = ith byte of data
-    /// - `MAK` = master acknowledge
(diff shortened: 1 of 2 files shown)


Dirbaio:
I'm not convinced we should do this. As unfortunate as the master/slave names are, thetoday the vast majority of microcontroller and i2c slave/target chips datasheets still use the master/slave terms. IMO the main goal of documentation is explaining to the user how to use the API, and making the user "mentally translate" between two sets of names when reading datasheets and our docs doesn't help that.

Also the controller/target names aren't universally agreed on either. There's also controller/peripheral ([recommended by the OSHWA](https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names/) for SPI, but i've also seen it [used for i2c](https://docs.rs/rp2040-hal/latest/rp2040_hal/i2c/index.html)), and I think I've also seen initiator/target somewhere too.

This was also discussed a bit in the weekly meeting when https://github.com/rust-embedded/embedded-hal/pull/396 was proposed, the conclusion was to keep master/slave names, since as long as it's in just docs (not trait/function/type name) we can easily change them to other names if/when other names become more widely used.

emshotton:
Thank you for the thoughtful reply and I want to acknowledge that naming things is a hard problem.

Interestingly the `rp2040_hal` uses a mix of language. It appears to use controller/peripheral when referring to the physical devices, and [master/slave](https://docs.rs/rp2040-hal/latest/rp2040_hal/i2c/struct.I2C.html#method.read-3) when referring to the devices addresses. The the `rp2040_pac` uses [master/slave](https://github.com/rp-rs/rp2040-pac/blob/89e78ff078d88cfbbad91bd5aa4eb4ac79a57dab/src/i2c0.rs#L70) when refering to the registers of the internal I2C device, but then the `DW_apb_i2c` IP block from Synopsys which appears to be the hardware implementation of i2c the rp2040 uses internally uses [controller/target](https://www.synopsys.com/dw/ipdir.php?ds=amba_apb_advanced) in it's public descriptions. 

We are currently in a situation where there is already a range of competing terminology and confusion, and given that situation I'd argue that moving towards the language in the specification documentation that is being adopted by the hardware IP providers is a sensible move.

Personally I think it's important to move towards standardizing around this new terminology and the place individuals have agency to effect this change is in the open source world. I expect that the terminology used by IC manufactures will lag furthest behind, but will ultimately follow once changes to the spec and software libraries are made.

This change dosent include changes to the SPI terminology as I'm unfamiliar with the current state of naming conventions for that.

amanjeev:
> Personally I think it's important to move towards standardizing around this new terminology and the place individuals have agency to effect this change is in open source libraries such as this.

I second this. It will be nice to lead in this effort.

PR #630, opened 2024-09-09
implement `core::error::Error`

rursprung:
this trait has been stabilised in Rust 1.81.0.

the existing custom `Error` types cannot be removed / replaced as that'd be a breaking change. for the same reason it's not possible for them to require `core::error::Error` being implemented.

however, we can already implement the new trait for all cases where the custom trait has been implemented so far.

existing `std` feature-gated implementations of `std::error::Error` have also been moved to `core::error::Error` and the feature gate removed.

this raises the MSRV to 1.81.0 for most crates, but based on the MSRV policy this should not be an issue.


rursprung:
see #629 for an alternative approach which however doesn't seem to work (but would be much more concise if it did!)

MathiasKoch:
How does the error trait fit with defmt? Is it possible to use it properly while still leveraging the greatness of defmt?

MathiasKoch:
Yeah, I get that. The question was more on whether there is a way to utilize the error trait in a defmt application, or if that would always end up being all device side formatting?

Dirbaio:
2 weeks have passed and no one from the HAL team has raised concerns, let's merge.

Anders429:
Any chance we can get a release with these changes in them? This would really help me be able to interoperate with error types in `embedded-io`.

PR #634, opened 2024-10-07
i2c: AddressMode: add bound Copy

spikespaz:
Since the trait `AddressMode` is `Sealed`, and only implemented on `Copy` types (`u8` and `u16`), there isn't any reason for the trait not to be bound also as `Copy`. This makes multiple-consumption much easier, and does not require calling code to add extra bounds.

### DIFF
--- a/embedded-hal/src/i2c.rs
+++ b/embedded-hal/src/i2c.rs
@@ -276,7 +276,7 @@ impl<T: ErrorType + ?Sized> ErrorType for &mut T {
 /// Address mode (7-bit / 10-bit).
 ///
 /// Note: This trait is sealed and should not be implemented outside of this crate.
-pub trait AddressMode: private::Sealed + 'static {}
+pub trait AddressMode: Copy + 'static + private::Sealed {}
 
 /// 7-bit address mode type.
 ///


jannic:
We had some discussion about this on Monday in the matrix room: https://libera.irclog.whitequark.org/rust-embedded/2024-10-07#1728335635-1728335961;

Adding `Copy` is, at least technically, a breaking change. Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=705af2100f5461aa539eaddb2321bfe4

However we didn't find any _realistic_ situation where one would write code that would be affected. 

There was no conclusion if this should be merged (ignoring the breaking change), or deferred for the next major version.

spikespaz:
We agree that this is a breaking change and not a major one. It might fall under the "additive" category. Isn't there a tool that is used to check how many crates would be broken by small changes like this?

Dirbaio (approved):
LGTM. 

As commented above and discussed in last's week meeting, this is technically breaking but is unlikely to be hit by anyone, since it affects `dyn AddressMode` and there was no way to do anything useful with it anyway. 

I'll wait until meeting on next Tuesday to merge in case another HAL team member wants to raise a concern.

therealprof (approved):
LGTM, too

PR #640, opened 2024-11-19
Don't recommend supporting eh0.2

MabezDev:
This PR changes the recommendation to add 0.2 impls into HALs.

It keeps the instructions but modifies them slightly to put it behind a feature so that we don't recommend third party HALs violate [C-STABLE](https://rust-lang.github.io/api-guidelines/necessities.html#public-dependencies-of-a-stable-crate-are-stable-c-stable). How important C-STABLE is for trait implementations is debatable but I still don't think we, as a rust working group, should be recommending use of a 0.2 version of a crate when a 1.0 exists.

## Alternatives

We remove this section completely and leave it to HAL authors

### DIFF
--- a/docs/migrating-from-0.2-to-1.0.md
+++ b/docs/migrating-from-0.2-to-1.0.md
@@ -470,14 +470,15 @@ Here is the full listing of crates:
 
 ## Supporting both 0.2 and 1.0 in the same HAL
 
-It is strongly recommended that HAL implementation crates provide implementations for both the `embedded-hal` v0.2 and v1.0 traits.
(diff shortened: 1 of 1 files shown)


Rahix:
I think it's also worth bringing <https://github.com/ryankurte/embedded-hal-compat> into this.

MabezDev:
> I think it's also worth bringing [ryankurte/embedded-hal-compat](https://github.com/ryankurte/embedded-hal-compat) into this.

It's mentioned in the next section, unless you meant to include it in this section as well?

Dirbaio (approved):

PR #643, opened 2025-01-11
Splittable trait

shimunn:
Adds splittable trait, signifying that the implementor can be split into an read and write instance


MabezDev:
Do you have a particular use case in mind for this trait? Would it not be better to structure your code in a way to accept the read and write halves separately instead of splitting once the full driver has been passed?

PR #645, opened 2025-01-25
Show features required for AtomicDevice in docs

jannic:
This show the features required for AtomicDevice in the docs. 

However, because `portable-atomic or target_has_atomic="8"` is such a long string, it looks a bit ugly, see attached screenshot. Is there a better alternative?

image


Dirbaio (approved):
yeah I agree it's a bit long but what else can we do? it's the truth. Anything else would be confusing :) 

thanks!

PR #649, opened 2025-02-12
Proposal: add async digital traits

### LINKED ISSUE #647: Add async variants of the traits in `embedded_hal::digital`
Currently all the traits in `embedded_hal::digital` are blocking GPIO APIs. For most microcontroller firmware this is normally fine since you are just reading some register in the address space to access of set state which is effectively immediate.

However this API falls apart when IO expanders enter the equation. Something like the MCP23017 may need to perform an I2C transaction to get the state of GPIO. Blocking during an I2C transaction can take some time and blocking may be undesirable.

I also am working some firmware that would let you turn an MCU into an I/O expander using postcard. Setting an output pin to a high level would need to transmit a packet over some transport to the target MCU. A workaround would be to queue the operation for sending asynchronously later and return `Ok`. However the workaround can lead to race conditions if the transport could lose packets in transit. To me it seems less error prone to be able to `await` a async form of `toggle`.

`embedded_hal_async::digital` does exist, but this only contains the `Wait` trait. This does let you listen for events, but does not let you set output pin state or ask for the current pin state asynchronously. The best way I can think of to make IO expanders act correctly would be to add traits to `embedded_hal_async` which are async equivalents of the traits under the `digital` module.

Another concern I have is crates using embedded-hal already depending on types like `OutputPin`. I don't think it's possible to express a trait bound like implement one of `embedded_hal::digital::OutputPin` or `embedded_hal_async::digital::OutputPin`. Even though this could be added without a semver breakage, I imagine the new traits might ripple into a lot of broken downstream crates.

i509VCB:
As discussed in #647 and on the rust-embedded matrix, the digial traits mean that IO expanders currently need to block. 

Also I am working on a library for turning an MCU into an IO expander (no public code yet, still very WIP), and accessing the IO behind that MCU would need to block. The amount of blocking could be significant given that my library will be usable from over SPI to Ethernet.

Currently the proposal just takes the traits in `embedded_hal::digital`, copies them and adds the `async` keyword in front of each function.

There are also some comments relating to what `Poll::Ready` means. The documented semantics are that when `Poll::Ready` is returned, whatever operation is involved has completed. So if a pin is set to a high level, when `Poll::Ready` is returned, the pin will be driving a high level.

The 2nd commit changes the docs of the blocking `digital` module in `embedded_hal` to mention that the traits are blocking (always have been).

I also hacked together a quick test with `postcard-rpc` to blink an LED over USB to show an implementation of this for at least `OutputPin`. There is some extra stuff I did throw in but it is unused: https://github.com/i509VCB/postcard-gpio-test/tree/master

Resolves #647

### DIFF
--- a/embedded-hal-async/src/digital.rs
+++ b/embedded-hal-async/src/digital.rs
@@ -1,5 +1,12 @@
 //! Asynchronous digital I/O.
 //!
+//! The [`OutputPin`], [`StatefulOutputPin`] and [`InputPin`] traits are `async` variants
+//! of the [blocking traits](embedded_hal::digital). These traits are useful for when
+//! digital I/O may block execution, such as access through an I/O expander or over some
+//! other transport.
+//!
+//! The [`Wait`] trait allows asynchronously waiting for a change in pin level.
+//!
 //! # Example
 //!
 //! ```rust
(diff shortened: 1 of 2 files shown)


Dirbaio (approved):
lgtm!

PR #650, opened 2025-02-15
removed chip select pin from ExclusiveDevice in embedded-hal-bus

uliano:
### Summary
- Removed the chip select (CS) pin requirement from `ExclusiveDevice` in `embedded-hal-bus`.
- Introduced `NoPin` as a dummy pin for hardware-managed CS scenarios.

### Why?
`ExclusiveDevice` was contradictory in its design—when a device exclusively owns the SPI bus, the CS pin is typically **wired low by hardware** and does not need to be controlled in software.  
However, the existing implementation **required** an `OutputPin`, which was unnecessary and effectively wasted.


Rahix (commented):
Can we be sure there are no use-cases for an exclusive device which still expects a proper chip-select signal?  I vaguely remember having seen such things before...

Beyond that, removing the CS from `ExclusiveDevice` now is a breaking change.  I think it's a better option to leave it in and give users the option to opt-out by providing a second constructor:

```rust
impl<BUS, D> ExclusiveDevice<BUS, NoPin, D> {
    pub fn without_cs(bus: BUS, delay: D) -> Self {
        ExclusiveDevice::new(bus, NoPin, delay).unwrap_infallible()
    }
}
```

Dirbaio:
there's *many* SPI chips that need CS toggles for each transaction, so they know where a command begins/ends, and they don't work if you tie CS permanently low.

uliano:
> there's _many_ SPI chips that need CS toggles for each transaction, so they know where a command begins/ends, and they don't work if you tie CS permanently low.

I see, then @Rahix is the only choice. 

I see what I can do along that approach

uliano:
@Rahix @Dirbaio. I refactored the PR along the lines of @Rahix suggestions above. Now the `NoPin` behavior is opt-in.

embedded-hal-bus/src/spi/exclusive.rs
```diff
@@ -49,6 +49,15 @@ impl<BUS, CS, D> ExclusiveDevice<BUS, CS, D> {
     }
 }
 
+use unwrap_infallible::UnwrapInfallible;
+
+impl<BUS, D> ExclusiveDevice<BUS, NoPin, D> {
+    /// Create a new [`ExclusiveDevice`] without a Chip Select (CS) pin.
+    pub fn new_no_cs(bus: BUS, delay: D) -> Self {
+        ExclusiveDevice::new(bus, NoPin, delay).unwrap_infallible()
```
Rahix, line 57: To not pull another dependency, you can also do

```rust
pub fn new_no_cs(bus: BUS, delay: D) -> Self {
    match ExclusiveDevice::new(bus, NoPin, delay) {
        Ok(this) => this,
        // no Err() branch
    }
}
```

Less readable, but avoids dependency-bloat...  Not sure what the official stance of e-h on that topic is...

embedded-hal/src/digital.rs
```diff
@@ -222,3 +222,27 @@ impl<T: InputPin + ?Sized> InputPin for &mut T {
...
+// Implement `OutputPin`
+impl OutputPin for NoPin {
+    fn set_low(&mut self) -> Result<(), Infallible> {
+        Ok(())
+    }
+
+    fn set_high(&mut self) -> Result<(), Infallible> {
+        Ok(())
+    }
+}
```
Rahix, line 248: I guess also implement all the other `digital` traits for this type?

embedded-hal-bus/src/spi/exclusive.rs
Dirbaio: yeah avoiding deps is always nice.

embedded-hal/src/digital.rs
uliano: @Rahix what the use case for a non existent InputPin? what is the value is supposed to return?

PR #652, opened 2025-03-06
`embedded_io_adapters::unblock::Unblock`:  adapter from embedded_io::{Read, Write} to embedded_io_async::{Read, Write}

alsuren:
Funny story: I burned a bunch of time writing this because I couldn't get embassy's examples/std/src/bin/serial.rs working. I assumed that there was some problem with their serial port adapter or something because it was never returning any data. It turns out my USB UART adapter has a developed broken RX, so this was a bit pointless 😢. I thought I'd share it anyway.

I'm not sure how useful it will be to people:
* It doesn't currently implement Seek.
* it's kind-of inefficient.
* the ergonomics are a bit naff, because it can't just assume embedded_io::ErrorType::Error = std::io::Error (there might be a way around this: I haven't thought about it too hard)

If anything, its main usefulness would probably be to point people towards `async_io::Async`, and give people a good chuckle at my expense. Feel free to close this PR if you don't think that it is useful. I'm also happy to re-work it, if you think it could be useful but just has the wrong shape.

### DIFF
--- a/embedded-io-adapters/Cargo.toml
+++ b/embedded-io-adapters/Cargo.toml
@@ -16,8 +16,10 @@ categories = [
 std = ["embedded-io/std"]
 tokio-1 = ["std", "dep:tokio", "dep:embedded-io-async", "embedded-io-async?/std"]
 futures-03 = ["std", "dep:futures", "dep:embedded-io-async", "embedded-io-async?/std"]
+unblock = ["std", "dep:embedded-io-async", "dep:blocking"]
 
 [dependencies]
+blocking = { version = "1.6.0", optional = true }
 embedded-io = { version = "0.6", path = "../embedded-io" }
 embedded-io-async = { version = "0.6.1", path = "../embedded-io-async", optional = true }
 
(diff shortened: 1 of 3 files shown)


alsuren:
Turns out this can also be done with `blocking::Unblock::new()`. I can't get any of these approaches to read from stdin when porting my embassy project to work on desktop though. Probably best to close this until I have something that actually works.

dess83-maker:
when

alsuren:
@dess83-maker `when` is not a useful comment. Assuming that you are not a spam bot, please expand on your comment.

For anyone who came here tryng to turn a blocking stdin reader into an async reader in a std+embassy environment:

I think that a cleaner approach is to just write a blocking parser and then spawn a thread to dump the messages on a channel. See this code as an example:

https://github.com/hoverkite/hoverkite/pull/245/files#diff-2009529c3fa33e0843368f5f842d3ca1845ba36c7aedcd6751d3fe7121742730R124

It means that you end up with an awkward mix of blocking and async parsers dotted around your codebase, but I think it's fine.

PR #660, opened 2025-04-14
nb: do not swallow errors in fmt::Write.

Dirbaio:
Errors were silently swallowed before.

Also fixes clippy lint due to using `map` for the side-effect.

Closes #655


MabezDev (approved):
Thanks!

eldruin:
@Dirbaio Thank you for this! could you also add an entry to the changelog about it?

Issue #657, opened 2025-04-09
Question about CAN abstractions


robamu:
I have a few question and possibly improvement suggestions for the CAN trait.

1. Is there a special reasons why the CAN frame trait contains constructors? The HAL implementors can just constrain the CAN frame type.
    For device drivers, the transmit function already assumes a constructed frame, and for reception.. Wouldn't an API just expect the user to   
    somehow take care of polling CAN frames and pass them to the driver (again, would mean that the frame is already constructed, and 
    passed by reference) ? Maybe it would make more sense to have something like CanTx which only provided transmit functions, and 
    possibly a specialized CanRx?
2. Any reason not to provide a default implementation for `Frame::is_extended`?

   ```rust
    /// Returns true if this frame is an extended frame.
    fn is_extended(&self) -> bool {
        match self.id() {
            Id::Standard(_) => false,
            Id::Extended(_) => true,
        }
    }
    ```

PR #663, opened 2025-04-29
add `is_extended` default impl for CAN frame

robamu:


robamu:
as mentioned in https://github.com/rust-embedded/embedded-hal/issues/657

Issue #566, opened 2024-01-12
embedded-io: API stability and 1.0.0 release


elast0ny:
Hello, I'm working on a project that is considering using `embedded-io`'s traits for `no_std` environments and had some concerns about the stability of the API.

For example, the popular `postcard` serde format supports `embedded-io` traits but only from version `0.4.0`. This means that I either have to commit to an old version to use `postcard` or make a fork of `postcard` (and upstream fixes ideally).
I assume this is a decision that anyone in the `no_std` ecosystem has to make if they want to be generic over `Read/Write`.

Is there any chance that `embedded-io` could take a stance on backwards compatibility of its `Read/Write` traits (as a statement in a `README` perhaps) and make a release of major version `1` ? 

I understand that the crate may not be in such a state to be stable "forever" but if so, it would provide much more confidence in committing to using it.

chrysn:
We'll have at least one more breaking change when Error becomes a subtrait of core::Error. I'm also not sure WriteReady are ready (ha!) for prime time.

Maintainers, we use this issue as a tracking issue to list what has to be done before -io 1.0?

chrysn:
The core::Error step has been done in https://github.com/rust-embedded/embedded-hal/pull/630 -- so the next embedded-io release is breaking.

But is there anything else to do? I think not. (The Error change was also the only effectful change since 0.6.1 of late 2023.)

chrysn:
Right, as you say: that PR added core::error::Error impls, but no requirements on the associated types or the embedded-io trait.

*Should* we do one more braking change before (or as) we ge to 1.0, to add that requirement? After all, from a post-1.81 perspective, the current requirements to have an own error trait that does *not* impl core::error::Error is a bit odd.

I'm softly in favor of doing one more breaking change -- but we should get to a decision *if* we want to change anything any more at all (and not changing the Error requirements sets a pretty high bar for any other changes).

rursprung:
as a breaking change for v1.0 we should bump `defmt` to v1 and (probably?) rename the `defmt-03` feature to just `defmt` (since `defmt` is not supposed to ever hit v2)

chrysn:
With #672 merged, we're now on the "there'll be one breaking release at least before 1.0" track (eg. a 0.7 which then hopefully semver-tricks up to 1.0).

#673 is in the review pipeline, those are so far the only changes that came up for a 0.7.

[...]

PR #673, opened 2025-08-03
io: Update defmt to 1.0, rename feature from "defmt_03" to "defmt"

chrysn:
We *could* do the defmt-1.0 update a bit more compatibly by just going to 0.3.100 (and possibly keeping the defmt_03 feature as a deprecated feature name), but given the discussion in https://github.com/rust-embedded/embedded-hal/issues/566#issuecomment-3148142408, it may make sense to spool this for a single breaking >>0.6 bump, and use that to get rid of anything that'd be viewed as legacy in a 1.0 version.

<del>One alternative to the hard cut would be to to keep a defmt_03 feature that merely pulls in 0.3.100, as a deprecated feature. If >>0.6 becomes a 0.7 that does the semver trick to 1.0, that feature could be removed as part of the 1.0 transition.</del>

An extra deprecated defmt-03 feature is left in place mainly for embedded-io-async, which optionally pulls it in through the path-or-version dependency.

### DIFF
--- a/embedded-io-async/Cargo.toml
+++ b/embedded-io-async/Cargo.toml
@@ -15,11 +15,11 @@ categories = [
 [features]
 std = ["alloc", "embedded-io/std"]
 alloc = ["embedded-io/alloc"]
-defmt-03 = ["dep:defmt-03", "embedded-io/defmt-03"]
+defmt = ["dep:defmt", "embedded-io/defmt"]
 
 [dependencies]
 embedded-io = { version = "0.6.1", path = "../embedded-io" }
-defmt-03 = { package = "defmt", version = "0.3", optional = true }
+defmt = { package = "defmt", version = "1", optional = true }
 
 [package.metadata.docs.rs]
 features = ["std"]
--- a/embedded-io/Cargo.toml
+++ b/embedded-io/Cargo.toml
@@ -15,10 +15,10 @@ categories = [
 [features]
 std = ["alloc"]
 alloc = []
-defmt-03 = ["dep:defmt-03"]
+defmt = ["dep:defmt"]
 
 [dependencies]
-defmt-03 = { package = "defmt", version = "0.3", optional = true }
+defmt = { version = "1", optional = true }
 
 [package.metadata.docs.rs]
 features = ["std"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
           cargo build
           --workspace
           --target thumbv7m-none-eabi
-          --features async,defmt-03
+          --features async,defmt-03,embedded-io/defmt,embedded-io-async/defmt
 
   msrv-1-81:
     runs-on: ubuntu-latest
--- a/embedded-io-async/README.md
+++ b/embedded-io-async/README.md
@@ -14,7 +14,7 @@ This project is developed and maintained by the [HAL team](https://github.com/ru
 
 - **`std`**: Adds `From` impls to convert to/from `std::io` structs.
 - **`alloc`**: Adds blanket impls for `Box`, adds `Write` impl to `Vec`.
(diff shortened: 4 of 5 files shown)


embedded-io/Cargo.toml
```diff
@@ -15,10 +15,15 @@ categories = [
...
+defmt = ["dep:defmt"]
+defmt-03 = ["defmt", "dep:defmt_03100"]
 
 [dependencies]
-defmt-03 = { package = "defmt", version = "0.3", optional = true }
+defmt = { version = "1.0.0", optional = true }
+# This feature is deprecated; it is left in place while dependencies such as
+# the worspace's embedded-io-async work with either 0.6 and 0.7, but will be
+# removed going to a 1.0 version.
+defmt_03100 = { package = "defmt", version = "0.3.100", optional = true }
```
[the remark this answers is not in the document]
chrysn: Still having the embedded-io 0.3 feature serves two purposes:
* Right now, embedded-io-async is in the same workspace, has its own defmt-03 feature and depends on the not-yet-version-bumped embedded-io created in this PR. To not have to update -async just yet, this provides a way to keep the build process happy. (This is my primary motivation here).
* For any revdeps that want to use this with defmt 0.3, they might themselves only depend on (say) 0.3.5, and if anything else keeps back defmt from updating to 0.3.100, their dependency might not advance far enough on its own to make the semver trick work.

But I'm not particularly happy about having this workaround still in place; the only other fix that comes to mind to keep CI happy is to do something to embedded-io-async -- either disentangle it from the workspace's version (remove the path= until we're approaching the 0.7/1.0 release) or start co-evolving it. Any preference?

embedded-io/Cargo.toml
```diff
@@ -15,10 +15,10 @@ categories = [
 [features]
 std = ["alloc"]
 alloc = []
-defmt-03 = ["dep:defmt-03"]
+defmt = ["dep:defmt"]
 
 [dependencies]
-defmt-03 = { package = "defmt", version = "0.3", optional = true }
+defmt = { version = "1.0.0", optional = true }
```
rursprung, line 21:
```suggestion
defmt = { version = "1", optional = true }
```

embedded-io-async/Cargo.toml
```diff
@@ -15,11 +15,11 @@ categories = [
 [features]
 std = ["alloc", "embedded-io/std"]
 alloc = ["embedded-io/alloc"]
-defmt-03 = ["dep:defmt-03", "embedded-io/defmt-03"]
+defmt = ["dep:defmt", "embedded-io/defmt"]
 
 [dependencies]
 embedded-io = { version = "0.6.1", path = "../embedded-io" }
-defmt-03 = { package = "defmt", version = "0.3", optional = true }
+defmt = { package = "defmt", version = "1.0", optional = true }
```
rursprung, line 22:
```suggestion
defmt = { package = "defmt", version = "1", optional = true }
```

rursprung (commented):
i'd just mention the major version and skip minor&patch - there are no specific requirements here (that'd only make sense if you'd e.g. depend on 1.2.3 and really _need_ something which was introduced only with that version; but 1 = 1.0 = 1.0.0).

(but that's just my opinion; i'm no maintainer)

otherwise this LGTM - thanks!

chrysn:
Thanks for reviewing; your comments all being applied, I've taken the liberty to squash and rebase them into a sensible history for the maintainer team's review.

rursprung (approved):
perfect, thanks!

in the nit-picking territory: just note that with the two commits i'd usually expect that i could only build the first without the second but that wouldn't be the case here as then `e-io-async` would depend on a non-existent feature in `e-io`. so you might want to squash the two together completely or already use the `defmt` feature flag in the `e-io` dependency in `e-io-async` in the first commit.

PR #674, opened 2025-08-06
implement Write for uninitialized slices

master-hax:
# background
similar to how we implement the `Write` traits on `&mut [u8]`, we can also add support for `&mut [MaybeUninit<u8>]` without using `unsafe`. could be especially useful with [core::io::BorrowedBuf](https://doc.rust-lang.org/core/io/struct.BorrowedBuf.html).

# changes
* implement `ErrorType`, `embedded_io::Write`, & `embedded_io_async::Write` for `&mut [MaybeUninit<u8>]`
* replace 2 existing usages of `&buf[..amt]` with `buf.split_at(amt).0` for consistency (and potential const-context compatibility)
* lil organization in embedded-io/src/impls/slice_mut.rs

### DIFF
--- a/embedded-io-async/src/impls/slice_mut.rs
+++ b/embedded-io-async/src/impls/slice_mut.rs
@@ -43,13 +43,9 @@ impl Write for &mut [MaybeUninit<u8>] {
             return Err(SliceWriteError::Full);
         }
         let (a, b) = mem::take(self).split_at_mut(amt);
-        buf.split_at(amt)
-            .0
-            .iter()
-            .enumerate()
-            .for_each(|(index, byte)| {
-                a[index].write(*byte);
-            });
+        unsafe {
+            core::ptr::copy_nonoverlapping(buf.as_ptr(), a.as_mut_ptr() as *mut u8, amt);
+        }
         *self = b;
         Ok(amt)
     }
--- a/embedded-io/src/impls/slice_mut.rs
+++ b/embedded-io/src/impls/slice_mut.rs
@@ -75,13 +75,9 @@ impl Write for &mut [MaybeUninit<u8>] {
             return Err(SliceWriteError::Full);
         }
         let (a, b) = mem::take(self).split_at_mut(amt);
-        buf.split_at(amt)
-            .0
-            .iter()
-            .enumerate()
-            .for_each(|(index, byte)| {
-                a[index].write(*byte);
-            });
+        unsafe {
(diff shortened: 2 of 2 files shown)


master-hax:
cc @MabezDev @Dirbaio ptal when you get a chance, thanks!

embedded-io-async/src/impls/slice_mut.rs
```diff
@@ -20,7 +20,36 @@ impl Write for &mut [u8] {
...
+/// kind `ErrorKind::WriteZero`.
+impl Write for &mut [MaybeUninit<u8>] {
+    #[inline]
+    async fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
+        let amt = core::cmp::min(buf.len(), self.len());
+        if !buf.is_empty() && amt == 0 {
+            return Err(SliceWriteError::Full);
+        }
+        let (a, b) = mem::take(self).split_at_mut(amt);
+        buf.split_at(amt)
```
Dirbaio, line 46: maybe core::ptr::copy_nonoverlapping has a better chance at getting optimized than iterator loops. Or maybe copy_from_slice but i'm not sure if it's possible to use soundly.

Dirbaio: also `buf.split_at(amt).0` can be written more readably as `buf[..amt]`

Dirbaio: > but that requires unsafe code

that's not a problem

> they behave differently though - .split_at() splits the lifetime and also works in const contexts, so i think it is preferred over slicing with []

the lifetime is not an issue with readonly `&[u8]` because aliasing is allowed, and slicing is also allowed in const.

master-hax: from what i can tell, slicing is not allowed in const contexts.

this code snippet compiles just fine:
```
const HELLO_WORLD_BYTES: &[u8] = b"hello world";
const HELLO_BYTES_SPLIT_AT: &[u8] = &HELLO_WORLD_BYTES.split_at(4).0;
fn main() { panic!() }
```

whereas this code snippet fails to compile:
```
const HELLO_WORLD_BYTES: &[u8] = b"hello world";
const HELLO_BYTES_SLICE_OPERATOR: &[u8] = &HELLO_WORLD_BYTES[..4];
fn main() { panic!() }
```

compiler error:
```
error[E0015]: cannot call non-const operator in constants
calls in constants are limited to constant functions, tuple structs and tuple variants
```

PR #678, opened 2025-08-09
add adapters for digest trait

master-hax:
there is an existing no_std trait `digest::Update` used in very popular crates like `sha2` - this trait is essentially equivalent to `ErrorType<Error=Infallible> + Write`

this PR adds adapters between the two traits.

example usage:
```
use embedded_io::Write;
use embedded_io_adapters::digest::FromDigest;
use sha2::{Digest, Sha256};

fn main() {
    let mut hasher = FromDigest::<Sha256>::default();
    hasher.write_all(b"hello world").unwrap();
    println!("{}", hex::encode_upper(hasher.into_inner().finalize()))
}
```

### DIFF
--- a/embedded-io-adapters/Cargo.toml
+++ b/embedded-io-adapters/Cargo.toml
@@ -13,11 +13,13 @@ categories = [
 ]
 
 [features]
+default = ["digest"]
 std = ["embedded-io/std"]
 tokio-1 = ["std", "dep:tokio", "dep:embedded-io-async", "embedded-io-async?/std"]
 futures-03 = ["std", "dep:futures", "dep:embedded-io-async", "embedded-io-async?/std"]
 
 [dependencies]
+digest = { version = "0.10.7", default-features = false, optional = true }
 embedded-io = { version = "0.6", path = "../embedded-io" }
 embedded-io-async = { version = "0.6.1", path = "../embedded-io-async", optional = true }
 
--- a/embedded-io-adapters/src/digest.rs
+++ b/embedded-io-adapters/src/digest.rs
@@ -1,0 +1,98 @@
+//! Adapters to/from `digest::Digest` traits e.g. sha2::Sha256
+
+use core::convert::Infallible;
+use digest::Update;
+use embedded_io::{ErrorType, Write};
+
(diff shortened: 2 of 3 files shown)


embedded-io-adapters/Cargo.toml
```diff
@@ -13,11 +13,13 @@ categories = [
 ]
 
 [features]
+default = ["digest"]
```
master-hax, line 16: should be fine to enable by default since this crate is also `no_std`

embedded-io-adapters/src/digest.rs
```diff
@@ -0,0 +1,108 @@
...
+            inner: T::default(),
+        }
+    }
+}
+
+impl<T: ErrorType<Error = Infallible> + Write> Update for ToDigest<T> {
+    fn update(&mut self, data: &[u8]) {
+        match self.inner.write_all(data) {
+            Ok(()) => {}
+            Err(_) => unreachable!(),
```
Dirbaio, line 101: it's better to do `Err(e) => match e{}` so the compiler checks for you that it's actually unreachable.

embedded-io-adapters/Cargo.toml
Dirbaio: i'd prefer to make it optional to not bloat up people's dependency trees and also because the 0.10 vs 0.11 thing

Dirbaio:
Interesting timing, we were discussing it yesterday in Matrix https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$e8f4Ojv7DjdxQcGhRarquhTGY88GPgchLVjG93ZPjxE?via=matrix.org&via=catircservices.org&via=mozilla.org

IMO it's a bit niche. For example, the `digest` crate itself doesn't provide a similar adapter for `std::io`. What is your use case for this? Do you have something that writes data to a `Write` and you want to just hash it and not keep the data? (if you keep the data you can just hash the `&[u8]` afterwards)

Also the rust-crypto ecosystem isn't known for its stability. Latest `digest` is 0.10, but there's already `0.11` release candidates in the pipeline. If we do this we should name the features `digest-010` and `digest-011`.

PR #679, opened 2025-08-12
Release embedded-io 0.7

chrysn:
Contributes-To: https://github.com/rust-embedded/embedded-hal/issues/566

This is a Draft PR until

* [x] I've satisfied all the interdependencies between other crates from the same workspace (as they have both version and path, they have to change their dependency version in the same PR),
* [x] I've updated the date to my best guess of the release, and
* [x] I think we won't wait more for any new changes (#566 has been open for a while, so that shouldn't be too long).

### DIFF
--- a/embedded-io-adapters/CHANGELOG.md
+++ b/embedded-io-adapters/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.7.0 - 2025-09-30
+
+- Update to embedded-io and embedded-io-async 0.7
+- Fix missing forwarding of provided method into std types
+
 ## 0.6.2 – 2025-08-06
 
 - Added `ToFmt` adapter for `core::fmt::Write`.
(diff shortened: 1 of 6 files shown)


ColinFinck:
@chrysn What's the status on this? Any blockers left for a 0.7 release?
I've already updated the README in #706 to reflect some of the changes to embedded-io.

chrysn:
I don't think there's anything other than someone from the team saying "Yes" missing. (I can rebase this any time, but there's probably no point to that until there's confirmation, given there should be a date to include).

jamesmunns:
Ping @rust-embedded/hal

mkroening:
The following PRs would be very useful to us:

- https://github.com/rust-embedded/embedded-hal/pull/696
- https://github.com/rust-embedded/embedded-hal/pull/697

Just to make sure they don't go unnoticed. :)

therealprof (dismissed):
LGTM

chrysn:
Thanks. Date is adjusted to today, hope this holds until the button is pressed :-)

Dirbaio:
We should do the release at the same time for -io and -io-async, otherwise async users are left without being able to use it.

This means merging #696.

chrysn:
With #696 merged, I've updated to release all 3 embedded-io, all with current changelogs.

While not breaking-updte critical, if #697 is ready to merge (I think it is), that could still be taken in; if not, let's go?

chrysn:
Done. Conveniently, this also updates the MSRV of async to 1.81, which would otherwise not be achievable due to the embedded-io dependency anyway.

Dirbaio (approved):

PR #683, opened 2025-08-18
Add `mutex-traits`-based implementations of `SpiDevice` and `I2c`

### LINKED ISSUE #119: Generic Mutex Trait

JalonWong:
Fix #119

### DIFF
--- a/embedded-hal-bus/Cargo.toml
+++ b/embedded-hal-bus/Cargo.toml
@@ -16,7 +16,7 @@ version = "0.3.0"
 
 [features]
 # Enable shared bus implementations using `std::sync::Mutex`
-std = ["alloc"]
+std = ["alloc", "mutex/std"]
 # Use `portable-atomic` to enable `atomic-device` on devices without native atomic CAS
 #
 # `portable-atomic` emulates atomic CAS functionality, allowing `embedded-hal-bus` to use `atomic-device` on hardware
@@ -37,6 +37,7 @@ embedded-hal-async = { version = "1.0.0", path = "../embedded-hal-async", option
 critical-section = { version = "1.0" }
 defmt-03 = { package = "defmt", version = "0.3", optional = true }
 portable-atomic = {version = "1.3", default-features = false, optional = true, features = ["require-cas"]}
+mutex = "1.0"
 
 [package.metadata.docs.rs]
 features = ["std", "async"]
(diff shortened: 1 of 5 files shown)


FeldrinH:
What exactly are the mutex/mutex-trait crates? I took a quick peek at them and they don't really seem to be used by anything in the ecosystem and are maintained under the tosc-rs GitHub organization (which I have never heard of before), rather than rust-embedded.

embedded-hal-bus/Cargo.toml
```diff
@@ -37,6 +37,7 @@ embedded-hal-async = { version = "1.0.0", path = "../embedded-hal-async", option
 critical-section = { version = "1.0" }
 defmt-03 = { package = "defmt", version = "0.3", optional = true }
 portable-atomic = {version = "1.3", default-features = false, optional = true, features = ["require-cas"]}
+mutex = "1.0"
```
jannic, line 40: I'm a bit confused:

The `mutex` crate itself contains the following explanation of the relationship between `mutex` and `mutex-traits`:
> The [mutex-traits](https://crates.io/crates/mutex-traits) crate should be used by library crates that want to be generic over different ways of exclusive access.
> 
> The [mutex](https://crates.io/crates/mutex) crate should be used by applications that need to select which implementation is appropriate for their use case.

As `embedded-hal-bus` is a library, shouldn't it depend on `mutex-traits`?

But then, I don't get how code like `MutexTraitsDevice` implementations should be written without using the `mutex` crate. So this is more likely a documentation issue in `mutex` and `mutex-traits` (or just me being particularly slow today) than an actual issue with this pull request.

Issue #686, opened 2025-08-26
Missing methods on the `Seek` trait


Conaclos:
I am in the process of adapting some code from `std::io` to `embedded-io` and I noticed some missing methods on `mbedded_io::Seek` compared to `std::io::Seek`:

- [`stream_position`](https://doc.rust-lang.org/std/io/trait.Seek.html#method.stream_position)
- [`seek_relative`](https://doc.rust-lang.org/std/io/trait.Seek.html#method.seek_relative)

Do you plan to add them?
I am specifically interested in `seek_relative` that claims to perform better in some cases.

PR #687, opened 2025-08-26
io: add `Seek::seek_relative`

Conaclos:
Partially address https://github.com/rust-embedded/embedded-hal/issues/686

This PR adds the `Seek::seek_relative` method that is stable since rust 1.80.0.
I took the opportunity of adding more docs to the other methods.
This is mostly a copy-paste of the `std::io::Seek` docs removing references to `std::io` types such as `BufReader`.

### DIFF
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -469,22 +469,58 @@ pub trait Write: ErrorType {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Conaclos:
Should I submit a PR for adding `Seek::seek_relative` in `async-io`?

PR #696, opened 2025-09-04
feat(async-io): make async `Write::flush` a required method

mkroening:
This PR makes `embedded-io-async`'s `Write::flush` method a required method, meaning all implementors have to implement it.

This aligns this method with [`std`'s `Write::flush`](https://doc.rust-lang.org/nightly/std/io/trait.Write.html#tymethod.flush) and [`embedded-io`'s `Write::flush`](https://docs.rs/embedded-io/latest/embedded_io/trait.Write.html#tymethod.flush).

This PR is more a question than a proposal.
I could not find any documented reasons for the async `flush` being a provided method.
If this turns out to be intentional, I am happy to leave it as is and document the reason instead.

CC: @Dirbaio, who added this in the initial design ([`d6f6419`](https://github.com/rust-embedded/embedded-hal/commit/d6f641995fc1597920b7d86bf58d0332b7ebbaeb#diff-12ff19effb01592514fbd9b611c8e98309918524b9e794c1e3aea9c08b49e4a1R52-R54)).

### DIFF
--- a/embedded-io-async/src/impls/slice_mut.rs
+++ b/embedded-io-async/src/impls/slice_mut.rs
@@ -26,6 +26,11 @@ impl Write for &mut [u8] {
     }
 
     #[inline]
+    async fn flush(&mut self) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    #[inline]
     async fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
         if self.len() < buf.len() {
             return Err(SliceWriteError::Full);
--- a/embedded-io-async/src/impls/vec.rs
+++ b/embedded-io-async/src/impls/vec.rs
@@ -11,6 +11,11 @@ impl Write for Vec<u8> {
     }
 
     #[inline]
+    async fn flush(&mut self) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    #[inline]
     async fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
         self.write(buf).await?;
         Ok(())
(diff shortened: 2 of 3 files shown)


mkroening:
I think aligning with `std` as close as possible would be best. Any deviation has the potential to be confusing.

According to the crate description, the crate provides traits equivalent to their `std` counterpart. The two major deviations (`Error` and `WouldBlock`) are explained in the [README](https://github.com/rust-embedded/embedded-hal/blob/c5a45aec40cd9c06c69e095cf230f70cd459f07b/embedded-io/README.md).

Also, there are recent precedents to being more restrictive than strictly necessary to align with `std` (https://github.com/rust-embedded/embedded-hal/pull/690, https://github.com/rust-embedded/embedded-hal/pull/695).

> This is not so much a technical but a educational / social question: Do we expect that people who buffer know that they need to implement a flush (because in their write impl, they explicitly chose not to really write), or do we want to rub it in visibly, even at the cost of annoying users who don't do buffering?

This question has been answered by the Rust team as well (https://github.com/rust-lang/rfcs/pull/576#issuecomment-72515231):

> > [rust-lang/rust#21835 (comment)](https://github.com/rust-lang/rust/pull/21835#discussion_r23898163)
> > Should `flush` return `Ok` by default? We may want to make an explicit decision about the default here (there hasn't been much discussion on the flushing topic)
> 
> I think @mahkoh is raising a good point here, and `flush` should be a required method (no default impl) to help ensure it has a sensible implementation.

After having researched this a bit more, I think making this a required method would be the right call. Deviating from it would require a good explanation, in my opinion.

PR #698, opened 2025-09-08
Expand embedded_io::ErrorKind to better match std::io::ErrorKind

### LINKED ISSUE #648: Expand embedded_io::ErrorKind to better match std::io::ErrorKind.
embedded_io::ErrorKind only has 18 variants while std::io::ErrorKind has 41. I am porting a zip library to no_std and would like to use more of the variants. Some of these variants would also be useful for filesystems on sdcards.
These are some of the errors I need to use.
NotADirectory,
IsADirectory,
DirectoryNotEmpty,
ReadOnlyFilesystem,
StorageFull,
NotSeekable,
QuotaExceeded,
FileTooLarge,
CrossesDevices,
InvalidFilename,
UnexpectedEof,

I don't see why not implementing all of them though?

Oakchris1955:
Resolves #648

I am creating this as a draft PR because of [this comment](https://github.com/rust-embedded/embedded-hal/pull/679#issuecomment-3262213239) stating that this change isn't an urgent one. Furthermore, while I added some `ErrorKind` variants to better match `std::io::ErrorKind`, there might be more that should be added.

### DIFF
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -114,6 +114,26 @@ pub enum ErrorKind {
     OutOfMemory,
     /// An attempted write could not write any data.
     WriteZero,
+    /// The filesystem or storage medium is read-only,
+    /// but a write operation was attempted.
+    ReadOnlyFilesystem,
+    /// The underlying storage (typically, a filesystem) is full.
+    ///
+    /// This does not include out of quota errors.
+    StorageFull,
+    /// Seek on unseekable file.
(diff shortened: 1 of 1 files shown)


Oakchris1955:
It seems that these additional ErrorKind variants were stabilized with Rust `1.83.0`, so adding them would mean bumping up the msrv. However, since `1.83.0` was released on November 2024, and taking into account that `1.81` is the current msrv, I think that this change is justified

PR #706, opened 2025-09-23
Update embedded-io README for defmt feature and MSRV

ColinFinck:
As a preparation for the (hopefully) upcoming release. Looking forward to #679.

CC @chrysn


eldruin (approved):
Thanks!

PR #708, opened 2025-09-30
io: remove supertraits for ReadReady, WriteReady. Release 0.7.1

Dirbaio:
- **io: remove `Read`/`Write` supertrait for `ReadReady`/`WriteReady`.**

turns out we screwed up this in 0.7.0. I've yanked those for now.  `ReadReady`/`WriteReady` are shared between blocking and async (the embedded-io-async crate just reexports them) so they shouldn't have the blocking `Read`/`Write` as supertraits.

- **Release embedded-io 0.7.1**


MabezDev (approved):
Makes sense, thanks!

PR #711, opened 2025-10-02
io: use clippy to ensure that methods are forwarded in adapters

Conaclos:
This adds `deny(clippy::missing_trait_methods)` on `impl` that must forward calls to the underlying type. This will avoid issues I fixed #705, #703, #701, and #700.

This allowed me to catch a miss.


Dirbaio (dismissed):
nice!

Dirbaio:
seems clippy is angry about std impls now

Conaclos:
> seems clippy is angry about std impls now

I always forgot to pass `--all-features` to Clippy!
It is now fixed. I had to revert my change for the std adapters.

Also, this discovered missing method forwarding for `Box<T>` types.

PR #712, opened 2025-10-07
Propagate lower-level errors as sources for embedded_io error types.

Anders429:
This expands the `core::error::Error` implementations for both `embedded_io::ReadExactError` and `embedded_io::WriteFmtError` to implement the `source()` function to return the underlying error contained in the `Other` variant.

Note that this changes the bounds for the `core::error::Error` implementations to both types to require `E: core::error::Error + 'static`, but in practice this shouldn't cause many issues since using both `Read::read_exact()` and `Write::write_fmt()` relies on the `ErrorType` trait, which requires its `Error` to implement `core::error::Error`. Technically this is a breaking change, and would possibly break places where someone is constructing these error types outside of the context of the io traits.

### DIFF
--- a/embedded-io/src/lib.rs
+++ b/embedded-io/src/lib.rs
@@ -261,7 +261,14 @@ impl<E: fmt::Debug> fmt::Display for ReadExactError<E> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Dirbaio (approved):
oh, very nice! :)

PR #716, opened 2025-11-27
Fully migrate to defmt 1.x

brainstorm:
Related to: https://github.com/embassy-rs/embassy/pull/4957

(and tangentially to https://github.com/mkj/sunset/pull/38)

### DIFF
--- a/embedded-hal/src/lib.rs
+++ b/embedded-hal/src/lib.rs
@@ -17,5 +17,5 @@ mod private {
 }
 
 // needed to prevent defmt macros from breaking, since they emit code that does `defmt::blahblah`.
-#[cfg(feature = "defmt-03")]
-use defmt_03 as defmt;
+#[cfg(feature = "defmt")]
+use defmt;
--- a/embedded-hal/src/lib.rs
+++ b/embedded-hal/src/lib.rs
@@ -15,7 +15,3 @@ mod private {
     impl Sealed for SevenBitAddress {}
     impl Sealed for TenBitAddress {}
 }
-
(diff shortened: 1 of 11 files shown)


embedded-hal/src/lib.rs
```diff
@@ -17,5 +17,5 @@ mod private {
 }
 
 // needed to prevent defmt macros from breaking, since they emit code that does `defmt::blahblah`.
-#[cfg(feature = "defmt-03")]
-use defmt_03 as defmt;
+#[cfg(feature = "defmt")]
+use defmt;
```
Dirbaio: this is not needed now that we're not renaming the dep anymore. (here and in other files either as `use defmt` or `use crate::defmt`)

brainstorm:
@Dirbaio @MabezDev Is there anything else I should address in this PR?

Dirbaio:
as I said in my previous comment:

> Renaming the feature defmt-03 -> defmt is a breaking change. Can you instead add a new feature defmt and keep defmt-03 as a deprecated alias?

brainstorm:
> as I said in my previous comment:
> 
> > Renaming the feature defmt-03 -> defmt is a breaking change. Can you instead add a new feature defmt and keep defmt-03 as a deprecated alias?

My apologies, I misunderstood your last message between parenthesis:


> (actually upgrading 0.3 -> 1.0 is not a breaking change since defmt 0.3.99 reexports 1.0 so they're compatible)

Anyway, thanks for your patience and for bumping the dependencies in my other PR! :)

PR #717, opened 2025-11-27
Add Cursor type

### LINKED ISSUE #631: [Feature Request] Provide `Cursor` Implementation in `embedded-io`
## Description
I've encountered a need for [`Cursor`](https://doc.rust-lang.org/stable/std/io/struct.Cursor.html) while using `embedded-io`. The current codebase doesn't seem to provide this feature, so I implemented it myself. However, I'm unsure if adding a `Cursor` to this repository aligns with the project's goals.

## Motivation
A `Cursor` can be useful for navigating through data streams or buffers, and it might benefit other users. I wanted to check if this feature would be a good fit for `embedded-io`.

## Proposal
If this is something you'd consider adding, I can submit a pull request with the implementation. Looking forward to your thoughts!

zeenix:
This is a no_std variant of `std::io::Cursor`.

Fixes #631.

The implementation was generated by an LLM by copying the std impl and adapting it. I (Zeeshan) reviewed the code and adjusted small bits. The tests added are a further assurance that the code is doing what it's supposed to do. Of course, it needs to be still reviewed by at least another person before we can be 100% certain about it.


embedded-io-adapters/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+/// A `Cursor` wraps an in-memory buffer and provides it with a [`Seek`] implementation.
+///
+/// `Cursor`s are used with in-memory buffers, anything implementing [`AsRef<[u8]>`],
+/// to allow them to implement [`Read`] and/or [`Write`], allowing these buffers to be used
+/// anywhere you might use a reader or writer that does actual I/O.
+///
+/// The standard library implements some I/O traits on various types which
+/// are commonly used as a buffer, like `Cursor<Vec<u8>>` and `Cursor<&[u8]>`.
+///
+/// This is the `embedded-io` equivalent of [`std::io::Cursor`].
```
jannic, line 20: Maybe this should mention that this also took big parts of the implementation from std?
I'm not sure which amount of attribution is legally required by the licenses, but I think it's at least good style to disclose such copying.

zeenix: > it's at least good style to disclose such copying.

* If it's complicated code and/or it's a specific author(s) to attribute, then I could agree but it's not really the case here.
* In any case, I don't think this should be in user-facing documentation as it doesn't affect users in any way. I'm OK with adding a non-doc comment that says so.
* Is this not true for any other parts of this crate? If it is the case, then we should mention it there too, no? :thinking:

jannic: It's probably more a personal preference than a rule, and I have no idea if other parts of this crate have been copied from elsewhere. (And my wording "it's at least good style to ..." is likely too strong, I should have written something like "I'd prefer to").
And I agree that it shouldn't be in the user-facing documentation. A simple comment "Parts of this file were copied from std::io::Cursor" would do.
If no one else comments on it, just take it as my personal opinion and feel free to ignore my comment.

1-rafael-1: Just an opinion: A reason to comment copied code is if it turns out to have a bug the person troubleshooting can check if the original code has already been fixed.

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+/// A `Cursor` wraps an in-memory buffer and provides it with a [`Seek`] implementation.
+///
+/// `Cursor`s are used with in-memory buffers, anything implementing [`AsRef<[u8]>`],
+/// to allow them to implement [`Read`] and/or [`Write`], allowing these buffers to be used
+/// anywhere you might use a reader or writer that does actual I/O.
+///
+/// The standard library implements some I/O traits on various types which
+/// are commonly used as a buffer, like `Cursor<Vec<u8>>` and `Cursor<&[u8]>`.
+///
+/// This is the `embedded-io` equivalent of [`std::io::Cursor`].
```
1-rafael-1: Just an opinion: A reason to comment copied code is if it turns out to have a bug the person troubleshooting can check if the original code has already been fixed.

zeenix:
@spookyvision may I ask what makes you 👎 this PR? If it's the use of LLM, perhaps you would have felt fine about this if I hadn't followed the guidelines of the Rust embedded community and hadn't disclosed the use of LLM?

embedded-io-adapters/src/cursor.rs
zeenix: Sure, I'll add the comment. Thank you for your detailed feedback. 👍 

Any comments on the changes themselves? I'd like this to be merged soon.

zeenix: Sure, I'll ask the comment. Thank you for your detailed feedback. 👍 

Any comments on the changes themselves? I'd like this to be merged soon.

embedded-io-adapters/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+/// cursor.write_all(b"some more bytes").unwrap_err();
+///
+/// assert_eq!(cursor.into_inner(), b"some bytes");
+/// ```
+///
+/// [`std::io::Cursor`]: https://doc.rust-lang.org/std/io/struct.Cursor.html
+#[derive(Debug, Default, Eq, PartialEq)]
+pub struct Cursor<T> {
+    inner: T,
+    pos: u64,
```
0e4ef622, line 44: Why is `pos` a u64 and not a usize?

zeenix: Because the std-API this is modelled after and heavily derived from, [use a u64](https://doc.rust-lang.org/stable/src/std/io/cursor.rs.html#77). I have also wondered sometimes about that but I always assumed that if it's in std, there probably is a good reason for it.

0e4ef622: Hmm, if there's a good reason, I can't find it. The best I can come up with is, `Seek` uses `u64`/`i64`, so `cursor.seek(SeekFrom::Start(u64::MAX))` shouldn't error. Which isn't very convincing. I also found https://github.com/rust-lang/rfcs/issues/894 regarding this. But I'm not a maintainer here, just my 2 cents.

zeenix: @0e4ef622 Thanks. I'd prefer keeping it as close to std as possible since the point is to provide the exact same API. I appreciate you looking into this though. :pray:

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+/// We may want to write bytes to a [`Write`], but not consume the buffer:
+///
+/// ```rust
+/// use embedded_io::{Write, Cursor};
+///
+/// let mut buf = [0u8; 10];
+/// let mut cursor = Cursor::new(&mut buf[..]);
+///
+/// cursor.write_all(b"some bytes").unwrap();
+/// // Internal buffer is not at capacity so writing more bytes fails.
```
9names: I don't think this comment makes sense. Was it meant to be "Internal buffer is now at capacity"?

You also don't describe *how* it fails.
Does it return a Err? Which one? The `std` version will keep calling write until it's all written, so this difference in behaviour should be documented.

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+impl<T> ReadReady for Cursor<T>
+where
+    T: AsRef<[u8]>,
+{
+    fn read_ready(&mut self) -> Result<bool, Self::Error> {
+        Ok(true)
+    }
+}
+
+// Write implementation for Cursor<&mut [u8]>.
```
9names: Why does this comment exist? It's just restating what the code says in about the same number of words.
Also none of the other impl blocks above got a comment so it's pretty weird that you decided to document this one.

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+
+#[cfg(feature = "alloc")]
+mod alloc_impl {
+    use alloc::vec::Vec;
+
+    use crate::{Write, WriteReady};
+
+    use super::{cmp, Cursor};
+
+    // Write implementation for Cursor<Vec<u8>>.
```
9names: This comment provides negative value (it's just the function signature).

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    #[cfg_attr(docsrs, doc(cfg(any(feature = "std", feature = "alloc"))))]
+    impl WriteReady for Cursor<Vec<u8>> {
+        fn write_ready(&mut self) -> Result<bool, Self::Error> {
+            Ok(true)
+        }
+    }
+
+    // Write implementation for Cursor<&mut Vec<u8>>.
```
9names: This comment provides negative value (it's just the function signature).

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+
+#[cfg(feature = "alloc")]
+mod box_impl {
+    use alloc::boxed::Box;
+
+    use crate::{Write, WriteReady};
+
+    use super::{cmp, Cursor, CursorError};
+
+    // Write implementation for Cursor<Box<[u8]>>.
```
9names: This comment provides negative value (it's just the function signature).

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+    }
+}
+
+impl<T> Cursor<T>
+where
+    T: AsRef<[u8]>,
+{
+    /// Returns the remaining slice.
+    #[inline]
+    pub fn remaining_slice(&self) -> &[u8] {
```
9names: In std [they removed `remaining_slice` and `is_empty`](https://github.com/rust-lang/rust/pull/109174) and replaced them with [split](https://doc.rust-lang.org/src/std/io/cursor.rs.html#233-237) and [`split_mut`](https://doc.rust-lang.org/std/io/struct.Cursor.html#method.split_mut).
I think switching to those functions would be an improvement (since `remaining_slice` is equal to `Cursor.split().1`) and `is_empty` is equal to `Cursor.split().1.is_empty()`, but unlike these functions it would not be unambiguous what it means for both readers and writers.

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+        let len = self.inner.as_ref().len();
+        let pos = cmp::min(self.pos, len as u64) as usize;
+        &self.inner.as_ref()[pos..]
+    }
+
+    /// Returns `true` if the remaining slice is empty.
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.remaining_slice().is_empty()
+    }
```
9names: As above, this is `Cursor.split().1.is_empty()` on std.

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+{
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
+        let slice = self.remaining_slice();
+        let amt = cmp::min(buf.len(), slice.len());
+
+        if amt == 1 {
+            buf[0] = slice[0];
+        } else {
+            buf[..amt].copy_from_slice(&slice[..amt]);
+        }
```
9names: Is it worth having an optimisation for a read of length 1 here?
Have you checked that this produces better code in the general case?

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+// Write implementation for Cursor<&mut [u8]>.
+impl Write for Cursor<&mut [u8]> {
+    fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
+        let pos = cmp::min(self.pos, self.inner.len() as u64) as usize;
+        let remaining = &mut self.inner[pos..];
+        let len = cmp::min(buf.len(), remaining.len());
+
+        if !buf.is_empty() && len == 0 {
+            return Err(CursorError::Full);
+        }
```
9names: On `std` this would return Ok(0), does it make sense to deviate from that behaviour here?

embedded-io/src/cursor.rs
```diff
@@ -0,0 +1,643 @@
...
+    #[cfg_attr(docsrs, doc(cfg(any(feature = "std", feature = "alloc"))))]
+    impl Write for Cursor<Box<[u8]>> {
+        fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
+            let pos = cmp::min(self.pos, self.inner.len() as u64) as usize;
+            let remaining = &mut self.inner[pos..];
+            let amt = cmp::min(buf.len(), remaining.len());
+
+            if !buf.is_empty() && amt == 0 {
+                return Err(CursorError::Full);
+            }
```
9names: On std this would return `Ok(0)`, does it make sense to deviate from that behaviour here?

embedded-io/src/cursor.rs
zeenix: That seems very strange to me and a foot gun (I actually have a vague memory of getting bitten by it). OTOH I also want this API to be as close to that of std as possible so I'm a bit torn here. So I'd welcome a 3rd perspective here.

zeenix: Oh and while looking into this, it turns out that `embedded_io::Write::write_all` has one major difference with `std::io::Write::write_all`: The former panics on `write` returning `Ok(0)`. So I think LLM did the right thing here. 😉

embedded-io/src/cursor.rs
zeenix: See [here](https://github.com/rust-embedded/embedded-hal/pull/717/changes/BASE..35adbdfc825c48260c427c3d5cc811bcd9113a5a#r2653420170) for justification of returning `Error` instead here.

zeenix:
I believe I've now addressed all the comments (except for the very unhelpful :-1: reactions from @spookyvision and @jamwaffles, which I can't do anything about).

@Dirbaio Any chance you'd be able to review soon?

reitermarkus (approved):

MabezDev (dismissed):
In my opinion this is a solid change, thanks for taking the time to refine the impl.

LGTM, I'll leave the final decision to @Dirbaio as eio is mostly his work.

zeenix:
@reitermarkus @MabezDev Thanks so much! I rebased the branch on latest master now.

zeenix:
> LGTM, I'll leave the final decision to @Dirbaio as eio is mostly his work.

@MabezDev While I think that's fair, I think we also shouldn't block things on @Dirbaio. They have single-handedly created most of this amazing Rust embedded world after all and they are one person in the end. Moreover, I have pinged them many times over the months about this PR, including after your comment on the Matrix channel and here, so it's safe to say that they don't have time for this right now.

So I'd humbly request that this be merged, unless @Dirbaio raises an objection/cocern. It has received multiple reviews and IMO a pretty straight forward PR.

Dirbaio:
Should this be on embedded-io or on embedded-io-adapters? If it's in embedded-io we can never ever make breaking changes to it once released. (I believe we've already discussed this in Matrix?)

Dirbaio:
There's some divergence compared to std cursor:

- `impl<const N: usize> Write for Cursor<[u8; N]>` is in std but not in e-io.
- `get_mut`, `set_position` are const in std, not in e-io.
- `remaining_slice` doesn't exist in std. Do we want to diverge from std here? IMO `split()` (which *is* in std) already covers all the use cases of `remaining_slice`.

zeenix:
> remaining_slice doesn't exist in std. Do we want to diverge from std here? IMO split() (which is in std) already covers all the use cases of remaining_slice.

`remaining_slice` did exist in the first iteration but was removed: https://github.com/rust-embedded/embedded-hal/pull/717#discussion_r2637561642


---

P.S. I'm AFK for a week so I'll address your concerns when I'm back.

Dirbaio:
> remaining_slice did exist in the first iteration but was removed

it wasn't removed, it's still there

zeenix:
> Should this be on embedded-io or on embedded-io-adapters? If it's in embedded-io we can never ever make breaking changes to it once released. (I believe we've already discussed this in Matrix?)

IMO it belongs here (embedded-io) but I don't foresee any breaking changes in this API. Perhaps that's why it's probably best we discuss this API to death to ensure all is good. :) However, if you or other maintainers think that it's best in the adapters crate, I'm happy to move it.

> > remaining_slice did exist in the first iteration but was removed
> 
> it wasn't removed, it's still there

Ah right. Sorry, I was already on holidays brain. :laughing: Since you're the second person to say this here, I guess we can remove it, yeah.

Dirbaio:
> IMO it belongs here (embedded-io) but I don't foresee any breaking changes in this API. Perhaps that's why it's probably best we discuss this API to death to ensure all is good. :) However, if you or other maintainers think that it's best in the adapters crate, I'm happy to move it.

generally we've tried to keep the trait crates as traits only. The adapters could also be in embedded-io but they're in a separate crate due to this.

We can discuss on Tue in the meeting.

zeenix:
> * `impl<const N: usize> Write for Cursor<[u8; N]>` is in std but not in e-io.
> * `get_mut`, `set_position` are const in std, not in e-io.
> * `remaining_slice` doesn't exist in std. Do we want to diverge from std here? IMO `split()` (which _is_ in std) already covers all the use cases of `remaining_slice`.

All addressed, I believe.

zeenix:
> * `get_mut`, `set_position` are const in std, not in e-io.

Oh, turns out there was a reason for this: [our MSRV](https://github.com/rust-embedded/embedded-hal/actions/runs/24395388055/job/71251777105?pr=717#step:4:112).

zeenix:
> > * `get_mut`, `set_position` are const in std, not in e-io.
> 
> Oh, turns out there was a reason for this: [our MSRV](https://github.com/rust-embedded/embedded-hal/actions/runs/24395388055/job/71251777105?pr=717#step:4:112).

I made them non-const again for now. Let me know if you'd rather bump the MSRV and have these const.

I also took the liberty of refactoring a bit to avoid code duplication in all the writing impls.

Dirbaio:
i'd just bump the MSRV yes

zeenix:
> i'd just bump the MSRV yes

Done. What was the consensus on where to put this in the meeting? It was hard to follow the last meeting. :laughing:

zeenix:
@Dirbaio also, you'll need to update the name of the required MSRV check in the settings here.

zeenix:
@Dirbaio @adamgreig Hi, as agreed, I moved this to -adapters. PTALA.

Also remove the version from the MSRV job so kindly update the settings again. :pray:

zeenix:
> @Dirbaio @adamgreig Hi, as agreed, I moved this to -adapters. PTALA.

Can you folks please do that? This PR has seen so much scrutiny and it's been up for almost 6 months now.

jannic:
There seems to be a plan to move Cursor to core::io: https://github.com/rust-lang/rust/pull/156428

Currently only on nightly, but I wonder if it still makes sense to duplicate that in embedded-hal, or if it would be better to wait for (or even help with) `core::io::Cursor` being stabilized?

zeenix:
> There seems to be a plan to move Cursor to core::io: https://github.com/rust-lang/rust/pull/156428
> 
> Currently only on nightly, but I wonder if it still makes sense to duplicate that in embedded-hal, or if it would be better to wait for (or even help with) `core::io::Cursor` being stabilized?

That's a valid concern but keeping in mind all the work that's been put in here, why not do both? Once `core::io::Cursor` stabalizes, we can deprecate this and eventually drop it on the next semver bump.

Dirbaio (approved):

PR #720, opened 2025-12-23
Add `defmt` feature to embedded-can Cargo.toml

dymk:
In order for the `defmt` dep to be used, it needs to be enabled by a feature (in this case, of the same name). As far as I can tell, simply having the dependency be optional does not create an implicit feature to enable it.

### DIFF
--- a/embedded-can/Cargo.toml
+++ b/embedded-can/Cargo.toml
@@ -15,3 +15,6 @@ repository = "https://github.com/rust-embedded/embedded-hal"
 [dependencies]
 nb = "1"
 defmt = { version = "1", optional = true }
+
+[features]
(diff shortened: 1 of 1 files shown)


Dirbaio:
simply having the dependency be optional does create an implicit feature to enable it :)

dymk:
Oops, that's embarrassing - I was on v0.4.1 of embedded-can which doesn't have the `defmt` feature, I mistook that as "optional dep does not define the feature". My bad!

I see this is something others have ran into in the issue tracker - I'll toss another vote out there for a release of embedded-can, looks like it's been a while since there's been one. For now, I'll use git as the package source in my cargo.toml.

PR #722, opened 2026-01-17
Add `defmt::Format` bound to error types

JalonWong:


eldruin (commented):
Thank you for your proposal.
However, this would force every error to depend on `defmt` as soon as the feature is used.
I prefer the current situation where errors _can_ implement `defmt` traits but are not forced to so we can ensure maximum flexibility for implementers.
This is my opinion at least for `embedded-hal`.
Please feel free to argue your case otherwise (and consider that this is a breaking change).

JalonWong:
Hi @eldruin ,
Without the constraint, I can't use the interfaces with `defmt`. For example:
```rust
fn foo<T: embedded_hal::i2c::I2c>(dev: &mut T) {
    let rst = dev.write(ADDR, &[REG_ADDR, 0]);
    defmt::info!("{}", rst);
    // or
    defmt::unwrap!(dev.write(ADDR, &[REG_ADDR, 0]));
}
```
This code can not be compiled, even though I enable the `defmt` feature and implement the `defmt::Format` trait for my error type. Which means the `defmt` feature of these crates is not really useful.

ExplodingWaffle:
You should be able to use a where bound like this on the error type:
```rust
fn foo<T: embedded_hal::i2c::I2c>(dev: &mut T)
where
    T::Error: defmt::Format
{
    let rst = dev.write(ADDR, &[REG_ADDR, 0]);
    defmt::info!("{}", rst);
    // or
    defmt::unwrap!(dev.write(ADDR, &[REG_ADDR, 0]));
}
```

JalonWong:
@ExplodingWaffle Thank you! I'll close this PR.

PR #723, opened 2026-01-23
Add traits for CAN FD

mickeprag:
- **Implement CAN FD frame**
- **Implement a CAN FD interface**

### DIFF
--- a/embedded-can/src/blocking.rs
+++ b/embedded-can/src/blocking.rs
@@ -15,3 +15,19 @@ pub trait Can {
     /// Blocks until a frame was received or an error occurred.
     fn receive(&mut self) -> Result<Self::Frame, Self::Error>;
 }
+
+/// A blocking CAN FD interface that is able to transmit and receive fd frames.
+pub trait CanFd {
+    /// Associated frame type.
+    type FdFrame: crate::FdFrame;
+
+    /// Associated error type.
+    type Error: crate::Error;
+
+    /// Puts a fd frame in the transmit buffer. Blocks until space is available in
+    /// the transmit buffer.
+    fn transmit_fd(&mut self, frame: &Self::FdFrame) -> Result<(), Self::Error>;
+
+    /// Blocks until a fd frame was received or an error occurred.
+    fn receive_fd(&mut self) -> Result<Self::FdFrame, Self::Error>;
+}
--- a/embedded-can/src/lib.rs
+++ b/embedded-can/src/lib.rs
@@ -51,6 +51,59 @@ pub trait Frame: Sized {
     fn data(&self) -> &[u8];
 }
 
+/// A CAN FD Frame
+pub trait FdFrame: Sized {
+    /// Creates a new fd frame.
+    ///
+    /// This will return `None` if the data slice is not 0-8, 12, 16, 20, 24, 32, 48, or 64 bytes long.
+    fn new(id: impl Into<Id>, brs: bool, data: &[u8]) -> Option<Self>;
+
+    /// Returns true if this frame uses bit rate switching (BRS).
+    fn is_brs(&self) -> bool;
+
+    /// Returns true if this frame is an extended frame.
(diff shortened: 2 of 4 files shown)


embedded-can/src/blocking.rs
```diff
@@ -15,3 +15,19 @@ pub trait Can {
     /// Blocks until a frame was received or an error occurred.
     fn receive(&mut self) -> Result<Self::Frame, Self::Error>;
 }
+
+/// A blocking CAN FD interface that is able to transmit and receive fd frames.
+pub trait CanFD {
```
liamkinne: My preferred capitalization here would be `CanFd` and likewise with `FdFrame` just because that IMO is the neatest way to handle concatenated acronyms.

embedded-can/src/blocking.rs
```diff
@@ -15,3 +15,19 @@ pub trait Can {
...
+
+    /// Associated error type.
+    type Error: crate::Error;
+
+    /// Puts a fd frame in the transmit buffer. Blocks until space is available in
+    /// the transmit buffer.
+    fn transmit_fd(&mut self, frame: &Self::FDFrame) -> Result<(), Self::Error>;
+
+    /// Blocks until a fd frame was received or an error occurred.
+    fn receive_fd(&mut self) -> Result<Self::FDFrame, Self::Error>;
```
liamkinne: How does this work for receivers that want to receive both FD and non-FD frames? By your description this would only return FD frames.

embedded-can/src/lib.rs
```diff
@@ -51,6 +51,56 @@ pub trait Frame: Sized {
...
+    /// - 16 bytes: DLC = 10
+    /// - 20 bytes: DLC = 11
+    /// - 24 bytes: DLC = 12
+    /// - 32 bytes: DLC = 13
+    /// - 48 bytes: DLC = 14
+    /// - 64 bytes: DLC = 15
+    fn dlc(&self) -> usize;
+
+    /// Returns the length of the frame data in bytes.
+    fn length(&self) -> usize {
```
liamkinne, line 89: Is this any different to `.data().len()`? I'm assuming it would be implemented that way anyway.

embedded-can/src/lib.rs
```diff
@@ -51,6 +51,56 @@ pub trait Frame: Sized {
     fn data(&self) -> &[u8];
 }
 
+/// A CAN FD Frame
+pub trait FDFrame: Sized {
```
liamkinne: This should include a BRS field as part of `new()` and something like `fn is_brs() -> bool`.

liamkinne (commented):
Just some thoughts on this implementation.

embedded-can/src/blocking.rs
mickeprag: I have no strong opinions about this myself, I will change as you suggest!

embedded-can/src/lib.rs
mickeprag: The length will be derived from the dlc. So the `.data()` function must read the dlc anyway to be able to return the correct length for the slice. Doing `.data().len()` needs to call `dlc()` internally anyway. I see no reason for the extra roundtrip creating a slice. I am providing a default implementation anyway so consumers of the trait does not need to implement it.

embedded-can/src/lib.rs
mickeprag: Agree! Thanks

embedded-can/src/blocking.rs
mickeprag: Do you have a suggestion for a signature?
Should we perhaps return an enum that can handle both frame types?

embedded-can/src/lib.rs
mickeprag: To be honest, I rather think the implementation will be the other way around. Implementations of `data()` will be calling the `length()` method, instead of `length()` calling `data()`...

PR #724, opened 2026-01-28
Created a first draft of the I3C trait.

### LINKED ISSUE #586: Create an I3C Trait
We should think about how a I3C Trait may be implemented. This may be done like the CAN Trait in an extra crate allowing breakage and only when it really stabilized to be merged into the core hal.

Currently there aren't a lot of IC's using the protocol, but it seems to get adoption. Major Suppliers have started incorporating it into their new designs, including NXP, STM, Renesas, Nordic Semi, Espressif, Texas Instruments, Bosch and Microchip. [See here for a list](https://binho.io/blogs/i3c-reference/i3c-devices).

This is supposed to be a place where we can discuss how this might be implemented.[ The specification can be found here.](https://www.mipi.org/specifications/i3c-sensor-specification) However it can only be accessed after registration and then only a reduced version compared to the member version.

The protocol itself is combining the advantages of I2C and SPI. It is downwards compatibel to I2C and starts adressing like it, however when it is clear that only the an I3C device is adressed it switches to a push pull architecture to enable higher transfer speeds like SPI. It does hoewever only support 7 Bit Adressing for the I2C compatibility mode.

Tremoneck:
Idea to close #586

There are still things missing in order of priority, though we could probably close at least the features in additional PRs
Common Command Codes
Dynamic Address Assignment
Interrupts
A lot of documentation
The other two high speed modes
Second Master
Hot Join

### DIFF
--- a/embedded-hal-i3c/src/lib.rs
+++ b/embedded-hal-i3c/src/lib.rs
@@ -1,5 +1,4 @@
 #![no_std]
-pub use embedded_hal::i2c::Operation;
 
 /// I3C error.
 pub trait Error: core::fmt::Debug {
@@ -59,6 +58,18 @@ pub trait I3c<S: SpeedMode = Sdr>: ErrorType {
         address: u8,
         operations: &mut [Operation<'_>],
     ) -> Result<(), Self::Error>;
+}
+
+/// I3C opeation.
+///
+/// Several operations can be combined as part of a transaction.
+pub enum Operation<'a> {
+    /// Read data into the provided buffer.
+    Read(&'a mut [u8]),
+    /// Write data from the provided buffer.
+    Write(&'a [u8]),
+    /// Enable and disable highspeed mode.
+    HighSpeed(bool),
 }
 
 trait Sealed {}
--- a/embedded-hal-i3c/src/lib.rs
+++ b/embedded-hal-i3c/src/lib.rs
@@ -6,28 +6,31 @@
 //!
 //! Drivers should take all speed variant which they support
 //! ```
-//! use embedded_hal_i3c::I3c;
+//! use embedded_hal_i3c::{I3c, Sdr, Ddr, SpeedMode};
+//!
+//! const ADDR: u8 = 0x158;
+//! # const TEMP_REGISTER: u8 = 0x1;
 //!
 //! pub struct TemperaturSensorDriver<I3C> {
 //!     i3c: I3C,
 //! }
 //!
-//! impl <I3C: I2c<Sdr>> TemperaturSensorDriver<I3C> {
+//! impl <I3C: I3c<Sdr>> TemperaturSensorDriver<I3C> {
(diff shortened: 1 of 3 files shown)


Tremoneck:
Ping @felipebalbi for discussion

embedded-hal-i3c/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn kind(&self) -> ErrorKind {
+        match *self {}
+    }
+}
+
+#[non_exhaustive]
+pub enum ErrorKind {
+    Bus,
+    ArbitrationLoss,
+    Ovverun,
```
felipebalbi: Typo here :)

embedded-hal-i3c/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+/// This just defines the error type, to be used by the other traits.
+pub trait ErrorType {
+    type Error: Error;
+}
+
+impl<T: ErrorType + ?Sized> ErrorType for &mut T {
+    type Error = T::Error;
+}
+
+pub trait I3c<S: SpeedMode = Sdr>: ErrorType {
```
felipebalbi, line 85: Not sure I agree with the Speed mode typestate here. We can switch between DDR and SDR in runtime, right? Will you require folks to consume and recreate the I3C (presumably with helpers `into_sdr()` and `into_ddr()`)?

Tremoneck: I thought about the controller implementing all supported speeds for the same object. The client can then use the different supported speeds depending on which trait is used to call transaction. In this design the whole transaction has the same speed. The switch from i2c to i3c would happen after the first address byte has been sent and arbitration is over.

Tremoneck: Apparently my idea on how I would implement this didn't work out or I'm missing on how to constrain the types properly. As can be seen in the test. Ideas on another Idea. Different traits for the different speeds could work, but they'd be unwieldy.

felipebalbi: Yeah, it'll be hard to come up with something sensible. But I think a single trait that lets you do a `transaction` and the `transaction` knows if it's supposed to be `I2c`, `I3cSdr` or `I3cDdr` would be simpler. Then, anything implementing `I3c` gets an `I2c` implementation for free by calling `I3c` methods with all transactions set to `I2c`.

Tremoneck: My idea behin the typestates was to ensure that a user is never able to combine a slave and controller, where the speed is not suported by both.

MabezDev:
Please follow https://github.com/rust-embedded/embedded-hal/blob/master/docs/how-to-add-a-new-trait.md before undrafting this for review.

Tremoneck:
I don't have time to drive this forward in any meaningfull way. I'd still be happy to discuss it.

Issue #727, opened 2026-02-22
New release for embedded-hal digital async traits


jerrysxie:
Hi, I implemented embedded-hal digital async traits for an IO expander driver here: https://github.com/OpenDevicePartnership/pcal6416a/pull/21, and I would like to publish a new version of the IO expander driver crate.

However, I learned that the embedded-hal digital async traits for `OutputPin`, `StatefulOutputPin` and `InputPin` are unreleased. Is there any upcoming plan to do a new release anytime soon? Thanks.

jerrysxie:
@eldruin Is there any chance for a new release that add these async digital traits?

PR #728, opened 2026-02-24
Update embedded-hal-async changelog

eldruin:
PR #727 brought to my attention that the `embedded-hal-async` changelog was missing some entries.


MabezDev (approved):
Thanks!

Issue #37, opened 2018-02-09
Explore DMA-based APIs built on top of `&'static mut` references


japaric:
What the title says. This [blog post] describes the approach to building such APIs. The last part of
the post covers platform agnostic traits that could be suitable for inclusion in `embedded-hal`.

[blog post]: http://blog.japaric.io/safe-dma

This issue is for collecting feedback on the proposed approach and deciding on what should land in
`embedded-hal`.

therealprof:
Great stuff. I like how you're tackling DMA. However I feel there's one very important piece missing in the embedded-hal world: The ability to tie interrupts to HAL periperals.

You nicely described how DMA gets rid of the need to wait for the transfer to complete on the spot but you still need to spend (potentially) a lot of cycles to check whether your transfer has completed at some point. Whether that is better or not depends on the application, after all you still might do other processing in interrupt handlers while busy waiting for a single transmission to complete.

therealprof:
I should probably add that I view interrupt support as a much more important topic for async IO and DMA mostly as a method to reduce interrupt frequency and contention. DMA without interrupts is not actually that useful.

protomors:
@jamesmunns As far as I know STM32 does not support peripheral to peripheral DMA transfers. Do other microcontrollers have such feature?

jamesmunns:
@protomors based on reading [this thread](https://community.st.com/thread/29041), I thought it might be possible, though that might be more by accident than by design. [AN4031](http://www.st.com/content/ccc/resource/technical/document/application_note/27/46/7c/ea/2d/91/40/a9/DM00046011.pdf/files/DM00046011.pdf/jcr:content/translations/en.DM00046011.pdf) section 1.1.4 though seems to exclude p-to-p DMA.

I can find some mentions of peripheral to peripheral DMA being used on other architectures, such as the PIC32, as well as the PSoC3 (8051), but can't really find any compelling examples for Cortex-M based devices.

I don't have a use case (or experience with this), but it looks like this is "not a problem" at the moment.

jamesmunns:
Actually, I just found the PSoC5LP (Cortex-M3 based) does support this, judging by this (app note)[http://www.cypress.com/file/137856/download] from Cypress, but this still looks like a pretty rare feature (might not be worth defining interfaces/traits for the edge cases).

therealprof:
[...]

PR #734, opened 2026-04-11
embedded-io-async: clarify cancel safety semantics for Read and Write

### LINKED ISSUE #719: Cancellation safety of `Read::read` and `Write::write` in embedded-io-async
The current documentation of the `read` and `write` methods in e-hal async leaves the cancellation safety explicitly down to implementors:
- https://github.com/rust-embedded/embedded-hal/blob/cb1f73e4bec46daa47d1a405ceb413d2fe6acb69/embedded-io-async/src/lib.rs#L49
- https://github.com/rust-embedded/embedded-hal/blob/cb1f73e4bec46daa47d1a405ceb413d2fe6acb69/embedded-io-async/src/lib.rs#L119

I wonder whether "encouraged" and "leave it down to implementers to document" is good enough though?
Tokio - in contrast - is pretty explicit about that:
- https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#cancel-safety
- https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#cancel-safety

I wonder whether we actually need two pairs of traits instead? I.e. for `Read`:
- `ReadNonCancelSafe` (let's leave the name bike-shedding as well as whether the non-cancel safe or the cancel-safe variant shall be named `Read` for later, shall we?) - with the current `Read` semantics and the "encouragement" paragraph replaced with an explicit "read is NOT cancel safe"
- `Read` - with an enforced cancellation-safe semantics, just like `AsyncReadExt` from tokio?

This way, ~~implementations~~ users that do care about cancellation safety would demand `Read`.
Those that don't might require the weaker `ReadNonCancelSafe`.

(This all in the context of stabilizing embedded-io-async.)

aki1770-del:
Fixes #719

## Problem

The documentation for `Read::read` and `Write::write` in `embedded-io-async` used permissive language — "implementations are *encouraged*" and "implementations *should* document" — that left both callers and implementors without a clear contract. Issue #719 identifies the concrete failure mode this ambiguity enables:

When a `read()` future is dropped mid-transfer, bytes already received from the hardware FIFO or DMA buffer are silently discarded. No error is returned. The caller has no way to detect or recover from the data loss. In safety-critical embedded contexts (CAN/UART frames, ISO 26262), this is a **silent data loss** that can propagate through state machines without triggering any diagnostic path.

## Fix

Rewrite the cancel-safety sections using normative `SHOULD`/`MUST` language and a dedicated `# Cancel Safety` heading (consistent with Tokio's convention):

- States explicitly: **this method is NOT guaranteed to be cancel-safe by default**
- Names the concrete failure mode: bytes in hardware FIFO/DMA may be silently dropped
- Requires implementations that cannot guarantee cancel safety to **MUST document** this
- Gives callers actionable guidance: run the future to completion rather than relying on cancel safety
- References the future `CancelSafeRead` marker trait path for opt-in

The PR changes documentation only. No trait methods are added, no signatures change, no semver bump required. The architectural path to a `CancelSafeRead` marker trait (issue #719's long-term ask) is left for a follow-up RFC.

## Existing Behavior Audit (OPS-RULE-016)

| File | Line | Current state |
|------|------|--------------|
| `embedded-io-async/src/lib.rs` | 49–56 | `Read::read` cancel-safety: "encouraged", "should document" — no normative contract |
| `embedded-io-async/src/lib.rs` | 119–126 | `Write::write` cancel-safety: identical permissive framing |
| PR #469 ("io: expand docs", 2023-07-12) | — | Prior documentation pass; introduced "encouraged" language that this PR strengthens |

## Scope Classification (OPS-RULE-017)

**Category (b): Minor enhancement (documentation).** No trait method signatures changed. No API-breaking change. A normative documentation tightening for existing behavior that was already the intended semantic, per the issue discussion. A full API change (new `CancelSafeRead` trait) would be category (c) requiring RFC — that is out of scope here.

## Test

`cargo check -p embedded-io-async` passes. Documentation-only changes require no additional test code; the trait method signatures are unchanged.

## Prior Art (OPS-RULE-018)

1. **PR #469** — *"io: expand docs"* (merged 2023-07-12) — the prior documentation pass that added "encouraged" language; this PR closes the gap that #469 opened by making the non-guarantee explicit
2. **Issue #37** — *"Explore DMA-based APIs built on top of `&'static mut` references"* (open, 2018) — root structural reason cancel safety is non-trivial for hardware FIFO implementations; the DMA buffer ownership problem this fix warns about
3. **PR #627** — *"Add BufReader impl"* (open) — concrete mechanism by which cancel safety can be achieved at the adapter layer without trait changes; this PR's documentation explicitly allows for that approach

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/embedded-io-async/src/lib.rs
+++ b/embedded-io-async/src/lib.rs
@@ -46,14 +46,24 @@ pub trait Read: ErrorType {
     /// If `buf.len() == 0`, `read` returns without waiting, with either `Ok(0)` or an error.
     /// The `Ok(0)` doesn't indicate EOF, unlike when called with a non-empty buffer.
     ///
-    /// Implementations are encouraged to make this function side-effect-free on cancel (AKA "cancel-safe"), i.e.
-    /// guarantee that if you cancel (drop) a `read()` future that hasn't completed yet, the stream's
-    /// state hasn't changed (no bytes have been read).
+    /// # Cancel Safety
     ///
-    /// This is not a requirement to allow implementations that read into the user's buffer straight from
-    /// the hardware with e.g. DMA.
+    /// **This method is NOT guaranteed to be cancel-safe.**
     ///
-    /// Implementations should document whether they're actually side-effect-free on cancel or not.
+    /// If a `read()` future is dropped before it completes, bytes that have already been transferred
+    /// from the hardware FIFO or DMA buffer may be silently discarded — no error is returned, and
+    /// no indication is given to the caller that data was lost. The next call to `read()` will return
+    /// bytes received *after* the cancellation point, not the discarded bytes.
+    ///
+    /// Implementations that are able to guarantee cancel safety (i.e. that dropping a pending future
+    /// leaves the stream state unchanged — no bytes consumed) SHOULD document this explicitly and
+    /// MAY implement the `CancelSafeRead` marker trait when it becomes available.
+    ///
+    /// Implementations that cannot guarantee cancel safety (e.g. those that use DMA to write
+    /// directly into the caller's buffer) MUST document this limitation.
+    ///
+    /// Callers that require cancel-safe reads SHOULD use a pattern that ensures the future runs to
+    /// completion (e.g. a dedicated read task) rather than relying on cancel safety.
     async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error>;
 
     /// Read the exact number of bytes required to fill `buf`.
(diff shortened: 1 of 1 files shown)


Dirbaio:
As far as I can tell the old and new wording is equivalent (implementations aren't required to be cancel-safe, implementations should document whether they are cancel-safe or not). So this PR is just a reword, it doesn't "fix" anything.

I personally find RFC-style MUST, SHOULD, MAY somewhat obnoxious to read, and it feels out of place here since we don't use it in the rest of rustdocs.

PR #743, opened 2026-07-02
Add ADC trait definitions

i509VCB:
Based off discussion from https://github.com/rust-embedded/embedded-hal/issues/377

I figured a pull request would be the best way to get feedback. I plan to write a few implementations of these traits, some on MCUs with a form of internal multiplexing for an MCU and another implementation for an external ADC (something like the ADS1015).

### DIFF
--- a/embedded-hal-async/src/adc.rs
+++ b/embedded-hal-async/src/adc.rs
@@ -1,0 +1,38 @@
+//! Asynchronous Analog to Digital conversion.
+//!
+//! See the [blocking](embedded_hal::adc) types documentation for more information.
+
+use core::future::Future;
+
+use embedded_hal::adc::{Channel, ErrorType};
+
+/// Asynchronous ADC channel capable of a one shot measurement.
+pub trait OneShotChannel: Channel + ErrorType {
+    /// The maximum value that may be produced by the ADC.
+    ///
+    /// This returns [`Ready`](core::task::Poll::Ready) when the max value is obtained. Some implementations such
+    /// as a channel connected over I2C may return [`Pending`](core::task::Poll::Pending) while the max value is
+    /// being obtained.
+    ///
+    /// This is guaranteed to not change once a channel is constructed.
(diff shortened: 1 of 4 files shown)


embedded-hal-async/src/adc.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// Asynchronous ADC channel capable of a one shot measurement.
+pub trait OneShotChannel: Channel + ErrorType {
+    /// The maximum value that may be produced by the ADC.
+    ///
+    /// This returns [`Ready`](core::task::Poll::Ready) when the max value is obtained. Some implementations such
+    /// as a channel connected over I2C may return [`Pending`](core::task::Poll::Pending) while the max value is
+    /// being obtained.
+    ///
+    /// This is guaranteed to not change once a channel is constructed.
```
felipebalbi, line 17: in that case, would it make sense to have as a constant? Or are you considering a case for ADCs with selectable range (presumably trading performance vs resolution)?

bugadani:
Would a nursery crate be a better target for introducing this?

embedded-hal-async/src/adc.rs
i509VCB: Yes this is to consider ADCs which a selectable resolution when configured. I can see making a query of max value async as a potential issue. I could see some less than ideal code literally asking over SPI/I2C/remote MCU interface every time for the resolution which would not be so free.

I could make this a non-async function of course which would make access effectively free and allow moving it to the common `Channel` trait.

i509VCB:
> Would a nursery crate be a better target for introducing this?

I am willing to do this (especially so that I don't commit e-h to a major version bump if we mess up). I would prefer the nursery crate lives in this repository if we wanted to do that.

i509VCB:
These traits are probably too idealistic and need a complete redo to represent reality of most ADCs especially with async where you may preprogram a sequence.

